Issue #36, opened 2026-01-07
Ability to mount different filesystem at specific paths


izziaraffaele:
Add support for mounting multiple filesystem backends at specific paths within a single virtual filesystem. This would allow combining read-only and read-write filesystems (e.g. knowledge bases, shared assets, and agent workspaces) into a unified namespace.

Ideal usage
```ts
// default in memory fs
const fs = new InMemoryFS();

// mount shared readonly data
fs.mount('/mnt/knowledge', new OverlayFs({ root: "/path/to/read-only-knowledge" }))
fs.mount('/mnt/skills', new OverlayFs({ root: "/path/to/read-only-skills" }))

// mount agent workspace
fs.mount('/home/agent', new ReadWriteFs({ root: "/path/to/workspace" }))

const env = new Bash({ fs, cwd: overlay.getMountPoint() });
```

cramforce:
Please explain the real world use-case. Looks cool

PR #37, opened 2026-01-07
Add MountableFs for mounting multiple filesystems at different paths

izziaraffaele:
Implements a new filesystem wrapper that allows mounting multiple IFileSystem backends at specific paths, enabling unified namespace with mixed read-only and read-write filesystems. Key features:

- Mount/unmount API with path validation (no root, no nested mounts)
- Path routing to correct filesystem with longest prefix match
- Cross-mount cp/mv operations
- Directory listings show mount points as directories
- Support for constructor-time and dynamic mounts

Related to #36


izziaraffaele:
@cramforce done!

PR #45, opened 2026-01-13
Change mount dir to /, which always exists

quuu:
previously the FS would mount /home/user/project, and that would lead the agent to literally `/home/user/project`


cramforce:
I don't think `/` is a good default. It's so "weird"

PR #67, opened 2026-01-29
Add `LazyFs` for lazy in-memory files

defrex:
I've ben loving just-bash! It's solved a ton of problems for me by enabling filesystem agents without a full sandbox.

However, I'm loading so many files in there that initializing the tool has gotten quite slow, and most of the time the agent doesn't even need the files.

The new `LazyFs` class has allowed me to quickly initialize the virtual filesystem and only load files that directories on an as-needed bases, which has greatly improved performance.

In addition to the unit tests in the PR, I've integrated this into my agents and run through my evals, etc confirming that this is working great for me.

## README Snippet

**LazyFs** - Lazy-loading filesystem that loads content on-demand via callbacks. Ideal for remote content, API-backed storage, or large datasets where you don't want to load everything upfront:

```typescript
import { Bash, LazyFs, MountableFs, InMemoryFs } from "just-bash";

const lazyFs = new LazyFs({
  // Called when reading a directory - return entries or null
  listDir: async (dirPath) => {
    // Replace with your API call, e.g.: await fetchDirectoryFromAPI(dirPath)
    return [
      { name: "file.txt", type: "file" as const },
      { name: "subdir", type: "directory" as const },
    ];
  },
  // Called when reading a file - return content or null
  loadFile: async (filePath) => {
    // Replace with your API call, e.g.: await fetchFileFromAPI(filePath)
    return { content: "file content here" };
  },
  allowWrites: true, // default, writes go to in-memory cache
});

// Mount it at a path
const fs = new MountableFs({ base: new InMemoryFs() });
fs.mount("/mnt/remote", lazyFs);

const bash = new Bash({ fs });
await bash.exec("ls /mnt/remote"); // triggers listDir("/")
await bash.exec("cat /mnt/remote/data.txt"); // triggers loadFile("/data.txt")
```

Content is cached after first access. Writes and deletes stay in memory and shadow the lazy-loaded content.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ package-lock.json
 .DS_Store
 .vscode
 .idea
+.claude/
 .env
 .env.*
 dist
--- a/README.md
+++ b/README.md
@@ -177,15 +177,18 @@ const fs = new MountableFs({
 import { Bash, LazyFs, MountableFs, InMemoryFs } from "just-bash";
 
 const lazyFs = new LazyFs({
-  // Called when reading a directory
+  // Called when reading a directory - return entries or null
   listDir: async (dirPath) => {
-    const entries = await fetchDirectoryFromAPI(dirPath);
-    return entries; // [{ name: "file.txt", type: "file" }, ...] or null
+    // Replace with your API call, e.g.: await fetchDirectoryFromAPI(dirPath)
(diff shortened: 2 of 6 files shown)


ghostyfreak:
Bumping this up

PR #68, opened 2026-01-30
Add dynamic isAllowed function for network policy

ziegfried:
Allow configuring network access with a custom function that dynamically
checks whether a URL/method combination is allowed. The function receives
{ method, url } and can return boolean or Promise<boolean>.

This provides more flexibility than static allowedUrlPrefixes for cases
like hostname pattern matching or async authorization checks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -356,6 +356,16 @@ const env = new Bash({
 const env = new Bash({
   network: { dangerouslyAllowFullInternetAccess: true },
 });
+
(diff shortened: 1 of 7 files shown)


ziegfried:
@cramforce is this an API you'd consider? It would be useful for my use of just-bash, but I might be able to work around it.

As an alternative, some sort of middleware function to handle network requests may work just as well.

cramforce (approved):

PR #84, opened 2026-02-03
Nativejq - many simple jq commands failed

robertyates:
seeing jq fail with fairly simple jq commands, have replaced with jq-was and a worker model so we can timeout, needed to update a few tests to match expected behaviour for jq, everything was checked against the command line


cramforce:
I'm extremely conscious of increasing the attack surface of just-bash and introducing a wasm dependency must be deeply analyzed in terms of how it interacts with the virtual filesystem and other resources.

It would probably be better to import a good test suite and fix the implementation

src/commands/jq/jq.ts
```diff
@@ -1,19 +1,85 @@
...
 import type { Command, CommandContext, ExecResult } from "../../types.js";
 import { readFiles } from "../../utils/file-reader.js";
 import { hasHelpFlag, showHelp, unknownOption } from "../help.js";
-import {
-  type EvaluateOptions,
-  evaluate,
-  parse,
-  type QueryValue,
-} from "../query-engine/index.js";
+import { Worker } from "node:worker_threads";
```
privatenumber: Could we preserve a browser-compatible execution path for jq?

jq is still registered in the browser command registry, but this implementation unconditionally imports `node:worker_threads` and uses Node's `Worker`. Existing browser consumers therefore lose jq even though jq-web itself supports browsers. This needs browser-worker support or a fallback to the existing evaluator.

src/commands/jq/jq.ts
```diff
@@ -268,81 +297,125 @@ export const jqCommand: Command = {
...
+      if (sortKeys) flags.push("-S");
+      if (useTab) flags.push("--tab");
+      if (joinOutput) flags.push("-j");
+
+      const outputParts: string[] = [];
 
       if (nullInput) {
-        values = evaluate(null, ast, evalOptions);
+        // Null input mode: run filter with null input
+        const { output } = await executeJqWithTimeout("null", filter, flags);
```
privatenumber: Could we pass the virtual shell's environment into the jq runtime?

The previous evaluator received `ctx.env`, but the worker now receives only input, filter, and flags. jq-web 0.6.2 supplies its own Emscripten environment, so a command such as `export MARKER=hello; jq -n '$ENV.MARKER'` loses the shell's value. The bridge should preserve the virtual environment rather than inherit the host process environment.

src/commands/jq/jq.ts
```diff
@@ -268,81 +297,125 @@ export const jqCommand: Command = {
...
+        // jq-web doesn't add the final newline, so we need to add it
+        output = outputParts.map(part => part.endsWith("\n") ? part : `${part}\n`).join("");
+      }
+      
+      // Calculate exit code for -e flag
+      // We need to check if output represents null/false/empty
+      let exitCode = 0;
+      if (exitStatus) {
+        const trimmed = output.trim();
+        if (!trimmed || trimmed === "null" || trimmed === "false") {
```
privatenumber: Could we preserve jq's exit status directly instead of inferring it from stdout?

With `jq -ner '"false"'`, the output is a string, so the expected exit code is 0; this comparison returns 1. Conversely, `jq -ne 'false, false'` produces two lines and returns 0 here, although its last value is false and the expected status is 1. Passing `-e` through and preserving the resulting status would avoid confusing formatted output with JSON values.

privatenumber (changes requested):
Requesting changes on the jq-web integration. Three gaps look concrete and blocking:

1. jq is still registered in the browser command registry, but the new implementation imports `node:worker_threads` and uses Node's `Worker`.
2. The virtual shell environment no longer reaches jq (the previous evaluator passed `ctx.env`).
3. `-e` exit status is reconstructed from formatted stdout, misclassifying `jq -ner '"false"'` (expects 0) and `jq -ne 'false, false'` (expects 1).

Each is called out inline. Separately, the resource model is worth a design discussion: the configurable `maxJqIterations` limit is replaced by a per-input 1000ms timeout, and every input value spawns a new worker that reloads jq-web.

Not part of this review: the 8KB encoding scan in `redirections.ts` was already merged via #69. I did not verify the WASM runtime's own isolation depth.

PR #170, opened 2026-03-26
Add Windows support to real-fs-utils

mutewinter:
Fixes path containment check on Windows where the separator is backslash.

Currently this breaks find, grep, cat, etc.

Two sites fixed in `real-fs-utils.ts`:

- `isPathWithinRoot`: use `nodePath.sep` instead of hardcoded `"/"` so the boundary check works on Windows
- `sanitizeSymlinkTarget`: normalize the sliced relative path with `.replaceAll(nodePath.sep, "/")` so virtual paths always use forward slashes regardless of OS

Both fixes are covered by tests using `vi.doMock` to inject `sep="\\"` at the module level, making the tests falsifiable on POSIX rather than no-ops.

Fix tested and working with my fork on `Windows 10 Home`

SCR-20260415-iste

### DIFF
--- a/src/fs/real-fs-utils.test.ts
+++ b/src/fs/real-fs-utils.test.ts
@@ -7,8 +7,9 @@
 
 import * as fs from "node:fs";
 import * as os from "node:os";
+import type * as nodePath from "node:path";
 import * as path from "node:path";
-import { afterEach, beforeEach, describe, expect, it } from "vitest";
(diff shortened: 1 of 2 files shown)


mutewinter:
Closing in favor of #187, which was merged and fixes the same two spots (`isPathWithinRoot` separator check and `sanitizeSymlinkTarget` relativePath normalization) with equivalent behavior. Verified the merged fix passes my Windows-mocked tests locally.

Thanks!

PR #186, opened 2026-04-10
Make MountableFs available to browser environments

boyeln:
We need MountableFs in browser environments but it's currently only exported from the main entry point, which Vite/bundlers resolve to the browser bundle where it's missing. Unlike OverlayFs and ReadWriteFs, MountableFs doesn't have any Node.js dependencies so it should be safe to include in the browser export.

We currently have a patch to enable this, but would be nice to remove that.

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -41,6 +41,11 @@ export type {
   RmOptions,
   SymlinkEntry,
 } from "./fs/interface.js";
+export {
+  MountableFs,
+  type MountableFsOptions,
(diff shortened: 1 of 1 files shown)


cramforce (approved):

cramforce:
Thanks. Could you fix the lint error?

boyeln:
Ah, sorry 🤦 Amended the fix to the existing commit (not sure if you're on team squash or team merge).

PR #190, opened 2026-04-13
Ship sqlite3 worker so npm consumers can use the command

subsetpark:
## Summary

The `sqlite3` command is broken when consumed via the published npm tarball. `findWorkerPath()` falls through to `<chunks>/worker.js` — which `build:worker` populated with the **python3** worker — so every `sqlite3` invocation fails with `Expected ArrayBuffer for the first argument` on protocol mismatch.

The bug is invisible in this repo because `tsc` emits `dist/commands/sqlite3/worker.js` locally, satisfying the resolver before the buggy fallback is reached. But the `package.json` `files` allowlist excludes that path from the tarball (only `dist/commands/**/*.d.ts` ships from that tree), so consumers never get it.

## Changes

- **`build:worker`**: bundle `src/commands/sqlite3/worker.ts` (externalizing `sql.js`) and place it at `dist/commands/sqlite3/worker.js`, `dist/bin/chunks/sqlite3-worker.js`, and `dist/bundle/chunks/sqlite3-worker.js` — mirrors the existing `js-exec` pattern with a uniquely-named chunk so it never collides with python3's `worker.js`.
- **`files`**: ship `dist/commands/sqlite3/worker.js` (chunk copies are already covered by the `dist/bin/` and `dist/bundle/` entries).
- **`findWorkerPath()`**: prefer `<currentDir>/sqlite3-worker.js` first, and gate the bare `<currentDir>/worker.js` lookup on the directory actually being `commands/sqlite3` so the silent fallback to a wrong-command worker can't return. The function now accepts an injectable `currentDir` and is exposed via `_internals` for testing.
- **`scripts/check-worker-sync.js`**: enforce that the bundled `sqlite3/worker.js` stays in sync with `worker.ts`.
- **lint configs**: skip the generated `src/commands/sqlite3/worker.js` artifact in `biome.json` and `scripts/check-banned-patterns.js` (mirrors existing python3 / js-exec treatment).
- **regression test** (`src/commands/sqlite3/sqlite3.worker-resolution.test.ts`): pin the resolution order with 5 unit tests, including one that proves the resolver does **not** fall back to a stray `worker.js`. Would have caught the original bug.

## Test plan

- [x] `pnpm build` — produces all three sqlite3 worker outputs
- [x] `pnpm typecheck`, `pnpm lint:fix`, `pnpm knip`, `pnpm check:worker-sync` — clean
- [x] `pnpm vitest run --config vitest.wasm.config.ts src/commands/sqlite3/` — 138/138 pass (including 5 new resolver tests)
- [x] `pnpm test:run --exclude src/spec-tests` — 9367/9367 pass (the one spec failure is a pre-existing, unrelated TZ test marked `## SKIP` that now passes; not touched here)
- [x] **Tarball smoke test**: `npm pack` → install in clean consumer → `Bash().exec("echo 'CREATE TABLE t(x); INSERT INTO t VALUES(42); SELECT * FROM t;' | sqlite3 :memory:")` returns `42` with exit 0 — no `Expected ArrayBuffer` error, no post-install workaround needed


cramforce:
Please sign your commits and re-push

subsetpark:
Hi @cramforce - I think I did already? My commit is listed as "verified"; let me know if it needs further work

cramforce (approved):

PR #206, opened 2026-04-27
fix(awk): honor POSIX statement continuation across newlines

subsetpark:
## Summary

The awk lexer emitted a `NEWLINE` token unconditionally whenever it saw `\n`, even when the previous token was one of the continuation-allowing tokens POSIX awk specifies (`,`, `{`, `&&`, `||`, `?`, `:`, `do`, `else`, `if`, `while`). Common multi-line idioms like

```awk
printf "%s=%d\n",
  $1, $2
```

(comma at end-of-line followed by indented args on the next line) parsed as two separate statements with a stray `NEWLINE` in the middle, surfacing as `Unexpected token: NEWLINE` — even though gawk, mawk, and the BSD one-true-awk all accept this form.

## Why this matters

This is the difference between just-bash awk being usable for routine scripting and not. Every TSV-to-SQL pipeline I've seen reaches for the comma-newline form, because keeping the printf format string and its args on separate lines is the only way to keep things readable past two columns. Hitting `Unexpected token: NEWLINE` on the most natural form of the most common idiom is a tough first impression.

## What changed

The lexer already tracks `lastTokenType` as an instance property, so the fix is a small inject in `nextToken`: when the next character is `\n` and `lastTokenType` is in a new `CONTINUES_ACROSS_NEWLINE` set, swallow the newline and recurse to the next real token instead of emitting one. The set covers every token POSIX awk lists as continuation-allowing.

Eight regression tests in `awk.parsing.test.ts` cover:
- comma in `printf` args (the motivating case)
- comma in function-call args (`substr`)
- `&&`, `||`
- `{` (block opening)
- `else`
- ternary `?`
- the TSV → SQL `INSERT` printf idiom end-to-end

## Test plan

- [x] `pnpm vitest run packages/just-bash/src/commands/awk/awk.parsing.test.ts` — 56 pass (8 new + 48 existing)
- [x] `pnpm vitest run packages/just-bash/src/commands/awk/` — full awk suite, 659 pass
- [x] Manually verified the TSV → SQL idiom that motivated this against the previous failure mode

## Notes

- POSIX 1003.1 ("Shell and Utilities", awk grammar): "A `<newline>` shall act as a statement separator… terminator may be replaced by certain tokens" — and lists exactly the set this PR uses.
- No new dependencies; pure lexer-internal change.
- Changeset added (`patch` bump) per the repo's release flow.

### DIFF
--- a/.changeset/awk-comma-continuation.md
+++ b/.changeset/awk-comma-continuation.md
@@ -1,0 +1,11 @@
+---
+"just-bash": patch
+---
+
+Fix awk lexer to honor POSIX statement continuation across newlines after `,`,
+`{`, `&&`, `||`, `?`, `:`, `do`, `else`, `if`, and `while`. Previously, a
+multi-line idiom like `printf "%s=%d\n", \n  $1, $2` (comma at end-of-line
+followed by indented args on the next line) failed with `Unexpected token:
+NEWLINE` because the lexer emitted a NEWLINE token unconditionally. The
+lexer now suppresses the NEWLINE when it immediately follows one of the
+continuation-allowing tokens, matching POSIX awk.
--- a/packages/just-bash/src/commands/awk/awk.parsing.test.ts
+++ b/packages/just-bash/src/commands/awk/awk.parsing.test.ts
@@ -40,6 +40,105 @@ describe("awk parsing", () => {
     });
   });
 
+  // POSIX awk specifies that a newline immediately following a comma (or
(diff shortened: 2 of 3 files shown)


cramforce:
Looks good. Could you sign your commits and repush?

subsetpark:
Thanks @cramforce , signed and force-pushed!

cramforce (approved):

PR #214, opened 2026-05-03
jq: accept literal control chars inside JSON strings

subsetpark:
## Summary
Real `jq` accepts JSON containing literal control characters inside strings (newlines, tabs, carriage returns, etc.) even though RFC 8259 forbids them. Real-world inputs — webhooks, log lines piped through `jq`, terminal output copy-pasted into a script — frequently carry literal tabs/newlines inside string values, and rejecting those at the `JSON.parse` boundary surprises users who expect parity with `jq`.

This patch pre-processes each JSON slice that `parseJsonStream` is about to hand to `JSON.parse`: it walks the slice tracking quote/escape state and, when a byte ≤ `0x1f` appears **inside** a string, replaces it with the appropriate JSON escape (`\n`, `\t`, `\r`, `\b`, `\f`, or `\uXXXX`). Outside strings the input is left untouched, so structural whitespace keeps its meaning.

Found while running agent-generated pipelines against just-bash where webhook payloads with embedded newlines triggered `Unexpected token` from `jq`, despite the same payload working under real `jq`.

## Test plan
- [x] `pnpm --filter just-bash exec vitest run src/commands/jq/` — 12 files / 276 tests pass (6 new in `jq.permissive-control-chars.test.ts`)
- [x] `pnpm lint:fix` — clean

New tests cover:
- literal newline / tab / CR inside a string value
- preserved escape sequence (`\\n`) sitting next to a literal control char
- control chars *outside* strings (structural whitespace) leaving compact output unchanged
- concatenated JSON values where one contains a literal newline

### DIFF
--- a/.changeset/sunny-donuts-matter.md
+++ b/.changeset/sunny-donuts-matter.md
@@ -1,0 +1,5 @@
+---
+"just-bash": patch
+---
+
+jq: accept control characters inside JSON strings
--- a/packages/just-bash/src/commands/jq/jq.permissive-control-chars.test.ts
+++ b/packages/just-bash/src/commands/jq/jq.permissive-control-chars.test.ts
@@ -1,0 +1,83 @@
+import { describe, expect, it } from "vitest";
+import { Bash } from "../../Bash.js";
+
+// Real jq accepts JSON containing literal control characters inside strings,
+// even though RFC 8259 forbids them. Real-world inputs (webhooks, copy-paste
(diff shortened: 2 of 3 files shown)


cramforce:
This looks great. Thanks for your contributions. Would you mind running changeset for this and your other PR?

subsetpark:
@cramforce done, thank you!

cramforce (approved):

PR #215, opened 2026-05-03
sqlite3: add dot-commands and -init/-batch invocation flags

subsetpark:
## Summary
Real `sqlite3` accepts a number of invocation shapes that just-bash's `sqlite3` didn't model. Agents and scripts shelling out to `sqlite3` regularly emit them, so getting an opaque misparse for them surfaces as a confusing failure rather than expected output.

Adds:
- **Dot-commands**: `.tables`, `.schema`, `.mode`, `.separator`, `.quit` / `.exit`. `.tables` and `.schema` are translated into `SELECT … FROM sqlite_master` queries; `.mode list|column|csv|…` toggles formatter state; `.separator` updates the column separator independently of `.mode` (matching real sqlite3); `.quit` / `.exit` short-circuit further input.
- **`-init <file>`** to run a SQL file before the inline command.
- **`-batch`** to run non-interactively.
- **Empty `ArrayBuffer` dbBuffer is coerced to `null`** so an empty file isn't treated as a corrupt database.
- **No-SQL guard** distinguishes absent vs. empty SQL so an empty heredoc doesn't print usage.

Coverage lives in four new test files plus extensions to `sqlite3.test.ts`:
- `sqlite3.dot-commands.test.ts` — dot-command parsing + execution
- `sqlite3.invocation-shapes.test.ts` — argv permutations agents pin against
- `sqlite3.flags.test.ts` — `-init` / `-batch` combinations
- `sqlite3.sql-features.test.ts` — SQL feature parity touched by the writeback worker changes

`docs/sqlite3-invocation-shapes.md` documents the supported shapes for downstream consumers.

## Test plan
- [x] `pnpm --filter just-bash exec vitest run src/commands/sqlite3/` — 15 files / 225 tests pass + 2 todo
- [x] `pnpm lint:fix` — clean

### DIFF
--- a/packages/just-bash/docs/sqlite3-invocation-shapes.md
+++ b/packages/just-bash/docs/sqlite3-invocation-shapes.md
@@ -1,0 +1,150 @@
+# sqlite3 invocation shapes — agent parity catalogue
+
+This index tracks how reasoning agents actually invoke `sqlite3` in production
+(via the `bash` tool) and our parity status. Source: a Braintrust review of
+~924 spans across the four `flowglad-pay-agent*` projects in 2026-04.
+
+Each shape maps to one or more pinning tests under
+`src/commands/sqlite3/sqlite3.{invocation-shapes,sql-features,dot-commands,flags}.test.ts`.
+If a test in this directory regresses, an agent in production has hit (or will
+hit) the same failure.
+
+This file is the **triage list** — the GitHub issue tracker is disabled on
+this fork, so deferred work lives here. Update the **Status** column when
+patches land.
+
+## Shell invocation shapes
+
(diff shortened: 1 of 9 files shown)


subsetpark:
Closing — the strict-layer dot-command preprocessor in this PR enforces a tighter contract than agent harnesses on top of just-bash were depending on (errors exit 1; line-only tokenization rejects `;`-separated dot-commands; unsupported/unknown commands fail rather than emitting in-band SELECTs; no SQL-string-literal awareness in the scanner). We're keeping the lenient char-scanner that already lives in worker.ts on the fork side. If we revisit this upstream we'll do it as a fresh PR that fixes those issues in the new layer rather than reverting the surface.

PR #218, opened 2026-05-06
perf(grep): 5–123× speedup via RE2 matcher reuse and literal pre-filter

Hazzng:
## Summary

Three layered optimisations in the regex engine and grep search loop that together produce a **5–123× speedup** on common grep patterns over a ~25k-line corpus, with all 2,051 directly-relevant tests passing.

## The investigation

Started with a benchmark report: `grep -ri 'async'` over 113 TypeScript files (25,956 lines) on a deployed sandbox API was taking **680 ms server-side**. Other "expensive" grep variants — `-rw`, `-rE` with alternation — were similarly slow at 476 ms and 823 ms. Simple greps (`-r 'interface'`, `-rn 'export'`) were fine (~20–40 ms). The pattern: anything that exercised RE2JS's case-insensitive, word-boundary, or alternation paths was 20–40× slower than expected.

### Issue 1 — RE2JS allocates a fresh Matcher per `test()`/`exec()` call

```ts
// before — src/regex/user-regex.ts
test(input: string): boolean {
    const matcher = this._re2.matcher(input);   // new RE2 Matcher every call
    return matcher.find();
}
```

For grep over 25 k lines, that's ~25,000 Matcher allocations per invocation — plus a second one per matched line for the column position (see Issue 3). RE2JS's per-Matcher setup cost dominates the actual matching work for short inputs.

### Issue 2 — `Matcher.resetMatcherInput(string)` is broken in re2js 1.2.1

The natural fix — cache the Matcher and rebind via \`resetMatcherInput(newInput)\` — silently corrupts grep:

```js
// re2js Matcher constructor wraps strings before assigning:
constructor(pattern, input) {
    if (input instanceof MatcherInputBase) this.resetMatcherInput(input);
    else if (Array.isArray(input))         this.resetMatcherInput(MatcherInput.utf8(input));
    else                                   this.resetMatcherInput(MatcherInput.utf16(input));  // ← string wrap
}

// but resetMatcherInput itself does NOT wrap:
resetMatcherInput(input) {
    this.matcherInput = input;        // raw string assigned directly
    this.reset();                     //   → reads this.matcherInput.length() — METHOD CALL
}                                     //   → string has .length (property), not .length() → TypeError
                                      //   → swallowed by grep's try/catch
                                      //   → 0 matches reported, no error surfaces
```

I hit this in the first version of the cache fix — the benchmark sped up but every grep returned 0 matches. The working fix: mutate the cached `Utf16MatcherInput.charSequence` field in place, then call `reset()`.

### Issue 3 — `searchContent` calls `test()` then `exec()` on the same line

```ts
// before — src/commands/search-engine/matcher.ts (default mode)
const matches = regex.test(line);          // RE2 find() #1
if (matches !== invertMatch) {
    regex.lastIndex = 0;
    const firstMatch = regex.exec(line);   // RE2 find() #2 — same line!
    const column = firstMatch.index + 1;
    // ...
}
```

`exec()` already returns null/non-null *and* the match position. Two `find()` calls per matched line is one too many.

### Issue 4 (the big one) — no Boyer-Moore-style literal pre-filter

GNU grep extracts literal substrings from the pattern at compile time and uses `memmem` / Boyer-Moore as a pre-filter. Lines that don't contain the literal are skipped without ever invoking the regex engine. RE2JS doesn't do this, and just-bash's grep didn't either — so for `-ri 'async'`, RE2 was running case-insensitive DFA scans on **all 25 k lines**, even though only ~1.7 k of them contained "async".

`String.indexOf` runs at memory-bandwidth speed in V8. RE2JS, being a JS port, doesn't come close. Skipping the regex entirely on no-needle lines turned out to be the single biggest win — and explains why the cumulative wins compound the way they do (matcher cache alone cut the slow patterns from ~700 ms to ~120 ms; the pre-filter then took them down further to single-digit ms).

### `src/regex/user-regex.ts` — Matcher cache

- Added \`acquireMatcher(input)\` that allocates the Matcher once, then mutates \`matcherInput.charSequence\` in place when the input changes. Calls \`reset()\` afterwards to clear match state.
- \`test()\` and \`exec()\` route through it. Other methods (\`replace\`, \`match\`, \`matchAll\`, \`search\`) still allocate per-call — they're not on the grep hot path and \`replace()\` invokes user-supplied callbacks, which makes shared mutable Matcher state risky.

### `src/commands/search-engine/matcher.ts` — combined `test()` + `exec()`

- Replaced the two-call pattern with a single \`regex.exec(line)\`. The result is null / non-null for the boolean *and* provides \`firstMatch.index\` for column info.
- For \`-o\` (only-matching) and vimgrep, the upfront \`exec()\` doubles as the first iteration of the all-matches loop.

### `src/commands/search-engine/regex.ts` — substring pre-filter

- New \`extractPreFilter(jsPattern, ignoreCase)\` returns \`{ needles, ignoreCase }\` when at least one literal substring can be safely required from the pattern, else \`null\`.
- Recognised shapes: bare literal (\`interface\`), \`\b…\b\` and \`\b(?:…)\b\` wrapping (i.e. \`-w\`), top-level alternation of literals (\`a|b|c\`), \`-F\`-mode escapes decoded back to literal text, and \`\n\`/\`\t\`/\`\r\` decoded to actual whitespace.
- Hard-rejects anything with quantifiers, anchors, character classes, character-class escapes (\`\d\` / \`\w\` / \`\s\`), groups, dots, or alternation where any branch contains meta. **False positives are fine; false negatives would silently break grep.**
- Returned alongside \`regex\` from \`buildRegex\`.

### `src/commands/search-engine/matcher.ts` — pre-filter integration

- New \`preFilterMatches(preFilter, line)\` helper.
- Wired into the four per-line loops (default / count / passthru / context). When the pre-filter says "definitely no match", regex is skipped and the boolean result is set to false. The \`-v\` interaction is preserved (a line that the pre-filter rejects still gets emitted by \`-v\` because it's a confirmed non-match).

### `src/commands/search-engine/regex.test.ts` — new test file (29 cases)

- **8 happy-path** — bare literal, \`-w\` wraps (single + alt), top-level alternation, \`-F\` decode, \`-i\` lowercase, \`\n\` decode
- **16 safety negatives** — every quantifier (\`+\`, \`*\`, \`?\`, \`{n,m}\`), char class, anchors (\`^\`, \`$\`), bare \`.\`, char-class escapes (\`\d\`/\`\w\`/\`\s\`), bare \`\b\`, capture group, non-capturing group without \`\b\b\` frame, mixed alternation, nested alternation in group
- **5 structural** — \`|\` correctly *not* split inside \`(?:…)\` or \`[…]\`, alternation order preserved, partial \`\b\` not over-stripped, empty pattern

The negatives are the safety-critical ones — false-positive needles in the extractor would silently cause grep to miss valid matches.

## Benchmark

Setup:  Recursive grep over `/home/user/src` in my private bench— 113 files, 25,956 lines of TypeScript. Server-reported \`duration_ms\`, p50 of 3 measured runs after one warmup run.

| Pattern | matches | **before** | **after** | **speedup** |
|---|---:|---:|---:|---:|
| \`grep -ri 'async'\` | 1655 | 680 ms | **14 ms** | **48×** |
| \`grep -rw 'type'\` | 366 | 476 ms | **5 ms** | **90×** |
| \`grep -rE 'interface\|type'\` | 634 | 823 ms | **7 ms** | **123×** |
| \`grep -r 'interface'\` | 50 | 44 ms | **4 ms** | 11× |
| \`grep -rF 'Promise<'\` | 316 | 26 ms | **3 ms** | 9× |
| \`grep -rn 'export'\` | 141 | 21 ms | **4 ms** | 5× |
| \`grep -rl 'SqlFs'\` | — | 20 ms | **3 ms** | 7× |

Cumulative breakdown for the three slow patterns (each row stacks on the previous):

| Stage | -ri 'async' | -rw 'type' | -rE 'interface\|type' |
|---|---:|---:|---:|
| baseline | 680 ms | 476 ms | 823 ms |
| + Matcher cache | 135 ms | 96 ms | 150 ms |
| + combined test/exec | 120 ms | 85 ms | 138 ms |
| + indexOf pre-filter | **14 ms** | **5 ms** | **7 ms** |

Match counts identical at every stage — verified against the deployed baseline and against system grep via \`src/comparison-tests/grep.comparison.test.ts\`.

## Test plan

- [x] \`pnpm vitest run src/regex\` — 67/67 pass
- [x] \`pnpm vitest run src/commands/grep\` — 212/212 pass (basic, advanced, perl, exclude, binary)
- [x] \`pnpm vitest run src/commands/search-engine\` — 29/29 pass (new test file)
- [x] \`pnpm vitest run src/commands/sed src/commands/awk src/commands/rg\` — 1379/1379 pass (sister commands sharing UserRegex)
- [x] \`pnpm vitest run src/spec-tests/grep\` — 332/332 pass
- [x] \`pnpm test:comparison src/comparison-tests/grep.comparison.test.ts\` — 32/32 pass (validates output against system grep)
- [x] Match counts in the live deployed-vs-local benchmark match exactly (1655, 366, 634, 141, 50, 316, 23)

Total **2,051 / 2,051** directly-relevant tests pass. Eight failures elsewhere in the full unit suite (\`src/cli/just-bash.bundle.test.ts\`, \`src/fs/cross-fs-security.test.ts\`, \`src/security/{attacks,sandbox}/python-*\`) reproduce on clean main without this patch — Python/CPython-Emscripten runtime tests and a filesystem symlink-to-real-dir test, none of which touch files in this diff.

### DIFF
--- a/.changeset/tasty-papers-love.md
+++ b/.changeset/tasty-papers-love.md
@@ -1,0 +1,5 @@
+---
+"just-bash": patch
+---
+
+grep: 5-123x faster pattern matching via RE2 matcher reuse and literal pre-filter
--- a/packages/just-bash/src/commands/grep/grep.ts
+++ b/packages/just-bash/src/commands/grep/grep.ts
@@ -237,6 +237,7 @@ export const grepCommand: Command = {
         afterContext,
         maxCount,
         kResetGroup,
+        preFilter,
       });
       if (quietMode) {
         return { stdout: "", stderr: "", exitCode: result.matched ? 0 : 1 };
--- a/packages/just-bash/src/commands/grep/grep.ts
+++ b/packages/just-bash/src/commands/grep/grep.ts
@@ -206,6 +206,7 @@ export const grepCommand: Command = {
 
     let regex: UserRegex;
     let kResetGroup: number | undefined;
+    let preFilter: import("../search-engine/regex.js").PreFilter | undefined;
     try {
       const regexResult = buildRegex(pattern, {
         mode: regexMode,
@@ -215,6 +216,7 @@ export const grepCommand: Command = {
       });
       regex = regexResult.regex;
       kResetGroup = regexResult.kResetGroup;
+      preFilter = regexResult.preFilter;
     } catch {
       return {
         stdout: "",
@@ -361,6 +363,7 @@ export const grepCommand: Command = {
               afterContext,
               maxCount,
               kResetGroup,
+              preFilter,
             });
 
             return { file, result };
--- a/packages/just-bash/src/commands/search-engine/matcher.ts
+++ b/packages/just-bash/src/commands/search-engine/matcher.ts
@@ -225,7 +225,11 @@ export function searchContent(
         if (onlyMatching) {
           // firstMatch is the first iteration of the all-matches loop.
           // exec()'s side effect already advanced lastIndex past it.
-          for (let match = firstMatch; match !== null; match = regex.exec(line)) {
+          for (
+            let match = firstMatch;
+            match !== null;
+            match = regex.exec(line)
+          ) {
             // If \K was used, extract from the capture group instead of full match
             const rawMatch =
               kResetGroup !== undefined ? (match[kResetGroup] ?? "") : match[0];
@@ -240,7 +244,11 @@ export function searchContent(
           }
         } else if (vimgrep) {
           // Vimgrep mode: output each match separately with full line
-          for (let match = firstMatch; match !== null; match = regex.exec(line)) {
+          for (
+            let match = firstMatch;
+            match !== null;
+            match = regex.exec(line)
+          ) {
             let prefix = filename ? `${filename}:` : "";
             if (showByteOffset) prefix += `${byteOffset + match.index}:`;
             if (showLineNumbers) prefix += `${i + 1}:`;
--- a/packages/just-bash/src/commands/search-engine/matcher.ts
+++ b/packages/just-bash/src/commands/search-engine/matcher.ts
@@ -3,6 +3,23 @@
  */
 
 import type { UserRegex } from "../../regex/index.js";
+import type { PreFilter } from "./regex.js";
+
+/**
+ * Substring fast-path check: returns true if at least one needle is present
+ * in the line, meaning the regex *might* match. False means provably no match.
+ *
+ * For case-insensitive filters, the line is lowercased once per call. Both the
+ * needles and the line are compared in lowercase.
+ */
+function preFilterMatches(preFilter: PreFilter, line: string): boolean {
+  const haystack = preFilter.ignoreCase ? line.toLowerCase() : line;
+  const needles = preFilter.needles;
+  for (let i = 0; i < needles.length; i++) {
+    if (haystack.indexOf(needles[i]) !== -1) return true;
+  }
+  return false;
+}
 
 /**
  * Apply a replacement pattern using capture groups from a regex match
@@ -61,6 +78,11 @@ export interface SearchOptions {
   multiline?: boolean;
   /** If \K was used, this is the capture group index containing the "real" match */
   kResetGroup?: number;
+  /**
+   * Optional substring fast-path: skip RE2 entirely for lines where no needle
+   * is present. Pre-computed by buildRegex from the source pattern.
+   */
+  preFilter?: PreFilter | null;
 }
 
 export interface SearchResult {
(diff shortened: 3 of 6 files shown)


Hazzng:
@cramforce Ive made some optimization for grep so its lightning fast now. Can you please have a look at this PR as i really need this grep performance. Thanks !

cramforce:
Nice! Could you re-push with signed commits?

Hazzng:
@cramforce just signed it. Can you have a look ? thanks alot

cramforce:
Looks good. There is a simple lint error. Please fix

cramforce:
And while you are at it, run pnpm changeset

Hazzng:
@cramforce All have been resolved. Thanks

cramforce (approved):

PR #219, opened 2026-05-07
fix(jq): preserve UTF-8 in path-assignment output

code-yeongyu:
## Summary
`jq` corrupts UTF-8 multibyte characters (Korean / CJK / any non-ASCII) inside string values when the filter performs a path assignment such as `.foo[0].bar = N` or `(.foo[] | select(.x==Y)).bar = N`. Identity filters and simple top-level edits are unaffected.

## Reproduction
A new test in `packages/just-bash/src/commands/jq/jq.utf8.test.ts` reproduces the bug with a Korean payload. The corruption signature is double-encoding: UTF-8 bytes are interpreted as Latin-1 and then re-encoded as UTF-8 (so `한` 0xED 0x95 0x9C becomes `í`).

## Root cause
`jq` reads file/stdin content through the shared binary path (`readFiles(..., "binary")`), but the jq command parsed those latin1 byte-strings as Unicode text directly. During serialization, those mis-decoded characters were emitted as mojibake.

## Fix
Decode jq inputs from binary byte-strings to UTF-8 before JSON parsing (`decodeBinaryToUtf8IfNeeded` in `packages/just-bash/src/commands/jq/jq.ts`). This is a minimal scoped fix in jq’s input path, with no parser/evaluator refactor.

## Verification
- New regression tests pass (`jq.utf8.test.ts` covers nested assignment and selected-path assignment)
- Full existing `pnpm test:run` suite remains green

### DIFF
--- a/.changeset/few-mice-explain.md
+++ b/.changeset/few-mice-explain.md
@@ -1,0 +1,6 @@
+---
+"just-bash": patch
+---
+
+jq: preserve UTF-8 multibyte text when applying nested path assignment filters.
+This prevents mojibake caused by treating JSON file bytes as Latin-1 before serialization.
--- a/packages/just-bash/src/commands/jq/jq.ts
+++ b/packages/just-bash/src/commands/jq/jq.ts
@@ -22,6 +22,32 @@ import {
 } from "../query-engine/index.js";
 import { sanitizeParsedData } from "../query-engine/safe-object.js";
 
+const strictUtf8Decoder = new TextDecoder("utf-8", { fatal: true });
+
+function decodeBinaryToUtf8IfNeeded(input: string): string {
+  if (!input) return input;
+
+  let hasHighByte = false;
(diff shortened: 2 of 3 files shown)


cramforce:
Thanks! Please re-push with signed commits!

code-yeongyu:
@cramforce Re-pushed with a signed commit ([`05b2c32`](https://github.com/vercel-labs/just-bash/commit/05b2c3230d9fb97b359461426dd68b274ba6bd97)) — same diff, GitHub web-flow signature, `Verified` ✅. Branch was briefly auto-closed during the rewrite; reopened. Ready for re-review.

code-yeongyu:
Closing this inactive PR as part of the author's requested stale-PR cleanup. The head has had no new commits since 2026-05-08, and there has been no substantive follow-up for over 30 days. This closes the proposal only: it does not mark the change as merged or fixed. The head branch is retained, so the work can be reopened or resumed separately.

PR #222, opened 2026-05-08
fix(yq): preserve UTF-8 bytes during YAML parsing

code-yeongyu:
## Summary

Same root cause as #219 (jq), applied to yq.

`yq` reads `ctx.stdin` and file content as latin1 binary strings (each char's charCode = one byte). When the YAML parser/serializer re-encodes the data, multi-byte UTF-8 sequences get double-encoded (e.g. `한` → `í•œ`).

## Fix

Apply `decodeBinaryToUtf8IfNeeded` at every `ctx.stdin` and file-content site, using the exact helper from the jq fix.

## Test

New regression test at `packages/just-bash/src/commands/yq/yq.utf8.test.ts` covering Korean / accented / CJK text via stdin and files.

### DIFF
--- a/.changeset/yq-utf8-fix.md
+++ b/.changeset/yq-utf8-fix.md
@@ -1,0 +1,5 @@
+---
+"just-bash": patch
+---
+
+fix yq UTF-8 mojibake by decoding binary stdin/file input before YAML parsing.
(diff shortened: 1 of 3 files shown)


cramforce:

PR #223, opened 2026-05-08
fix(html-to-markdown): preserve UTF-8 bytes during conversion

code-yeongyu:
Same root cause as #219 (jq). decodeBinaryToUtf8IfNeeded applied at ctx.stdin / file content sites in html-to-markdown.ts. Regression test at html-to-markdown.utf8.test.ts.


cramforce:

PR #224, opened 2026-05-08
fix(sed): preserve UTF-8 bytes during regex substitution

code-yeongyu:
Same root cause as #219 (jq). sed read `ctx.stdin` / file content as latin1 binary strings — multi-byte UTF-8 sequences in input got double-encoded after substitution. Apply `decodeBinaryToUtf8IfNeeded` at every input site. Regression test at `sed.utf8.test.ts`.


cramforce:

PR #225, opened 2026-05-08
fix(grep): preserve UTF-8 bytes during pattern matching

code-yeongyu:
Same root cause as #219 (jq). Apply `decodeBinaryToUtf8IfNeeded` at `ctx.stdin` / file content sites in grep.ts. Regression test at `grep.utf8.test.ts` covers Korean / CJK content matching, line counting, and anchored patterns.


cramforce:

PR #226, opened 2026-05-08
fix(sqlite3): preserve UTF-8 bytes in SQL input

code-yeongyu:
Same root cause as #219 (jq). SQL piped via stdin could arrive as a latin1-style binary string, so UTF-8 literals were corrupted before reaching sql.js. This normalizes only SQL input from ctx.stdin and adds regression coverage for UTF-8 round-trips via sqlite3 stdin.


cramforce:

PR #227, opened 2026-05-08
fix(awk): preserve UTF-8 bytes during text processing

code-yeongyu:
Same root cause as #219 (jq).


cramforce:

PR #228, opened 2026-05-08
fix(rg): preserve UTF-8 bytes during ripgrep search

code-yeongyu:
Same root cause as #219 (jq). Apply `decodeBinaryToUtf8IfNeeded` at `ctx.stdin` / file content sites in rg-search.ts. Regression test at `rg.utf8.test.ts`.


cramforce:

PR #229, opened 2026-05-08
fix(xan): preserve UTF-8 bytes in CSV input

code-yeongyu:
Same root cause as #219 (jq). decodeBinaryToUtf8IfNeeded at all three ctx.stdin sites in xan/csv.ts and xan/xan-data.ts. Regression test at xan/xan.utf8.test.ts.

### DIFF
--- a/.changeset/xan-csv-utf8-fix.md
+++ b/.changeset/xan-csv-utf8-fix.md
@@ -1,0 +1,5 @@
+---
+"just-bash": patch
+---
+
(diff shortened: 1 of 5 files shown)


code-yeongyu:
Addressed Vercel agent's additional suggestion. `xan cat` (xan-simple.ts cmdCat) now also passes file content through `decodeBinaryToUtf8IfNeeded` before `parseCsv`, with a new regression test in `xan.utf8.test.ts`. Force-pushed signed commit [`373e757`](https://github.com/code-yeongyu/just-bash/commit/373e757576b8bd95632f847b352ca3e90aea7dff).

cramforce:

PR #230, opened 2026-05-08
fix(js-exec): preserve UTF-8 bytes when forwarding stdin / inline code

code-yeongyu:
Same root cause as #219 (jq). Apply `decodeBinaryToUtf8IfNeeded` at `ctx.stdin` sites in js-exec.ts before forwarding to the QuickJS worker. Regression test at `js-exec.utf8.test.ts`.


cramforce:
I'd prefer one larger utf-8 related PR. This is too much too review

cramforce:

PR #231, opened 2026-05-08
fix(python3): preserve UTF-8 bytes when forwarding stdin / inline code

code-yeongyu:
Same root cause as #219 (jq). Apply `decodeBinaryToUtf8IfNeeded` at `ctx.stdin` sites in python3.ts before forwarding to the CPython worker. Regression test at `python3.utf8.test.ts` covers code-via-stdin, runtime stdin, and `-c` arg paths.


cramforce:

PR #245, opened 2026-05-15
test: cover filesystem interface contract

### LINKED ISSUE #196: Feature request: expose IFileSystem conformance test suite for custom implementations
## Problem

We're building a custom `IFileSystem` implementation (backed by cloud object storage) and want to ensure it handles edge cases correctly — especially security-sensitive ones like path traversal and null byte injection.

`cross-fs-security.test.ts` is exactly what we need, but it's not exported from the package. We ended up:
1. Reading through the test suite to understand the expected behaviors
2. Copying `path-utils.ts` functions locally (since they're also not exported)
3. Writing our own subset of security tests based on them.

This works, but it means our tests can drift from the canonical implementation, and other `IFileSystem` implementers will repeat this process.

## Proposal

Export a conformance test helper that implementers can run against their `IFileSystem`:

```typescript
import { runIFileSystemConformanceTests } from 'just-bash/testing';

describe('MyCustomFs', () => {
  runIFileSystemConformanceTests(() => new MyCustomFs());
});
```

Happy to help with a PR if y'all are open to this.

EfeDurmaz16:
## Summary
- add internal regression coverage for the existing `IFileSystem` behavior
- cover null-byte, traversal, and virtual symlink edge cases
- keep the coverage inside the existing test suite without adding a public testing entry point or new user-facing concept

Fixes #196.

## Verification
- `pnpm --filter just-bash exec vitest run src/fs/interface.contract.test.ts`
- `git diff --check`

The remaining red checks are Vercel preview deployment authorization checks.

### DIFF
--- a/packages/just-bash/src/fs/interface.contract.test.ts
+++ b/packages/just-bash/src/fs/interface.contract.test.ts
@@ -1,0 +1,72 @@
+import { describe, expect, it } from "vitest";
+import { InMemoryFs } from "./in-memory-fs/in-memory-fs.js";
+
+describe("IFileSystem contract", () => {
+  it("reads, writes, appends, stats, lists, and removes files", async () => {
+    const fs = new InMemoryFs();
+
+    await fs.mkdir("/docs", { recursive: true });
+    await fs.writeFile("/docs/readme.md", "hello");
+    await fs.appendFile("/docs/readme.md", " world");
+
+    expect(await fs.readFile("/docs/readme.md")).toBe("hello world");
+    expect(await fs.exists("/docs/readme.md")).toBe(true);
+    expect((await fs.stat("/docs/readme.md")).isFile).toBe(true);
+    expect(await fs.readdir("/docs")).toContain("readme.md");
+
+    await fs.rm("/docs/readme.md");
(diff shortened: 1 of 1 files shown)


cramforce:
I'd rather not add yet another concept

EfeDurmaz16:
Agreed. I removed the public `just-bash/testing` entry point, the README docs, and the changeset. This is now just internal filesystem contract coverage under the existing test suite, without adding a new user-facing concept.

EfeDurmaz16:
I also cleaned up the branch history after narrowing the scope: this is now a single signed commit on current `origin/main` with only `packages/just-bash/src/fs/interface.contract.test.ts` changed.

Re-ran the focused checks after the rewrite:

- `pnpm --filter just-bash exec vitest run src/fs/interface.contract.test.ts`
- `git diff --check`

The remaining red checks are still Vercel preview deployment authorization checks.

cramforce:
Please repush with a signed commit

EfeDurmaz16:
Re-pushed with a GitHub-verified SSH-signed commit: 987a15c. No content changes beyond rewriting the commit signature. The remaining red checks are Vercel deployment authorization checks.

cramforce (approved):

PR #248, opened 2026-05-18
perf(grep): up to 14.5× speedup via preFilter extensions and matcher reuse

Hazzng:
## Summary

- **Anchored alternation preFilter** (`regex.ts`): `extractPreFilter` now strips leading `^` / trailing `$` from each alternative before extracting the literal needle. Patterns like `^def \|^async def` (previously unoptimized) now get the `String.indexOf` fast-path instead of running the full RE2 NFA against every line.
- **Matcher reuse extended** (`user-regex.ts`): `match()`, `replace()` (both string and callback paths), `search()`, and `matchAll()` now route through `acquireMatcher()`. Previously only `test()` and `exec()` used the cached matcher, leaving awk/sed hot-paths allocating a new `RE2JS.Matcher` on every call.
- **Multiline file-level preFilter** (`matcher.ts`): `searchContentMultiline()` now receives `preFilter` and performs a whole-file `content.includes(needle)` check before splitting lines or invoking RE2 — files with no matching needle are rejected in O(n) string scan instead of O(n·m) NFA.
- **grep loop preFilter short-circuit** (`grep.ts`): after `readFile`, a file-level needle check now skips `searchContent` (and `content.split("\n")`) entirely for files with no match. Count-only mode (`-c`) emits `0\n` correctly without entering the line loop.

## Benchmark results

100 files × 100 lines, 5 runs, median reported. Baseline: `just-bash@3.0.1` (npm latest).

| Pattern | Before (npm) | After (branch) | Speedup |
|---|---|---|---|
| `^def \|^async def` (anchored BRE — **the key case**) | 139.5 ms | 9.6 ms | **14.5×** |
| `def` (simple literal — baseline) | 17.6 ms | 7.7 ms | **2.3×** |
| `def \|async def` (unanchored alternation) | 29.9 ms | 9.3 ms | **3.2×** |

The baseline itself dropped 2.3× because `acquireMatcher` now covers `match()`, `search()`, and `replace()` — reducing GC pressure for all regex operations, not just the grep hot path.

## Root cause (the `^def \|^async def` case)

`literalFromAlternative` in `regex.ts` rejected any alternative containing `^` or `$`, treating them as regex metacharacters. Since `^L` implies the line contains `L`, and `L$` implies the line contains `L`, stripping outer anchors before needle extraction is provably sound (false positives are safe; RE2 re-checks the match). The fix is a 17-line anchor-strip in `literalFromAlternative` before the existing character loop.

## Tests added

| File | New tests |
|---|---|
| `src/commands/search-engine/regex.test.ts` | anchored single / alternation / mixed-anchor / anchor-only / escaped-anchor cases |
| `src/regex/user-regex.test.ts` | `acquireMatcher` reuse for `match`, `replace` (string + callback), `search`, `matchAll`, 1000-call state leak check |
| `src/commands/search-engine/matcher.test.ts` | multiline preFilter: no-needle early exit, needle-present normal run, invertMatch bypass |

### DIFF
--- a/packages/just-bash/src/regex/user-regex.ts
+++ b/packages/just-bash/src/regex/user-regex.ts
@@ -277,9 +277,15 @@ export class UserRegex implements RegexLike {
       return matcher.replaceFirst(replacement, true);
     }
 
-    // Callback replacement - we need to do this manually
+    // Callback replacement - we need to do this manually.
+    // Use a fresh Matcher rather than the shared cached one: the user-provided
+    // callback may re-enter this same UserRegex instance (e.g. call test/exec/
+    // replace), which would route through acquireMatcher and repoint the shared
+    // matcher's charSequence to a different input. The next matcher.find(pos)
+    // would then advance through the wrong string. A fresh matcher keeps the
+    // iteration state private to this replace() call.
     const result: string[] = [];
-    const matcher = this.acquireMatcher(input);
+    const matcher = this._re2.matcher(input);
     let lastEnd = 0;
     let pos = 0;
     const groupCount = this._re2.groupCount();
@@ -312,9 +318,9 @@ export class UserRegex implements RegexLike {
         args.push(groups);
       }
 
-      // Capture positions before invoking callback — the callback may re-enter this
-      // UserRegex instance, which would cause acquireMatcher to mutate the shared
-      // matcher's charSequence in-place, corrupting start/end reads after the call.
+      // Capture positions before invoking callback. The matcher is private to
+      // this call, but capturing now avoids relying on matcher state being
+      // unchanged across the callback boundary.
       const matchStart = matcher.start(0);
       const matchEnd = matcher.end(0);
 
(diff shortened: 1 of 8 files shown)


Hazzng:
@cramforce i've made some code optimization for grep so its getting quite abit faster now compared to the last PR. Can you have a look please ? thanks

packages/just-bash/src/regex/user-regex.ts
```diff
@@ -280,7 +279,7 @@ export class UserRegex implements RegexLike {
 
     // Callback replacement - we need to do this manually
     const result: string[] = [];
-    const matcher = this._re2.matcher(input);
+    const matcher = this.acquireMatcher(input);
```
cramforce: replace() callback mode now uses the shared cached matcher. The callback is user-controlled and can re-enter the same UserRegex instance. The code captures matchStart/matchEnd, but after the callback returns the next matcher.find(pos) still uses the possibly repointed matcher. Example: a callback that calls re.test("x") on the same regex can make re.replace("foo", cb) replace only the first o. Callback replacement should keep using a fresh matcher, or save/restore the cached matcher state around the callback.

Hazzng: nice catch! I just committed the fix to rollback to a fresh matcher.

cramforce (approved):

PR #250, opened 2026-05-19
fix(sqlite3): handle Bun structured-clone surfacing null as zero-length ArrayBuffer

subsetpark:
## Summary

Bun's `worker_threads` structured-clone has regressed across versions — most visibly in the build shipped in Trigger.dev's runtime container — and surfaces a host-side `null` `dbBuffer` as a zero-length `ArrayBuffer` inside the worker thread.

The current call site treats any truthy `data.dbBuffer` as a database image:

```ts
if (data.dbBuffer) {
  db = new SQL.Database(data.dbBuffer);
} else {
  db = new SQL.Database();
}
```

An empty `ArrayBuffer` is truthy in JS, so the buggy clone slips past the guard and reaches sql.js — which rejects bare `ArrayBuffer` (it wants `Uint8Array`) with:

```
Expected ArrayBuffer for the first argument
```

Net effect: every `:memory:` `sqlite3` invocation throws under affected Bun builds.

## Change

Introduce `coerceDbBuffer(raw: unknown): Uint8Array | null` in the worker and route the buffer through it:

```ts
const buf = coerceDbBuffer(data.dbBuffer);
db = buf ? new SQL.Database(buf) : new SQL.Database();
```

Behavior:

- `Uint8Array` (and `Buffer`, which extends `Uint8Array`) — returned as-is
- non-empty `ArrayBuffer` — wrapped in a `Uint8Array`
- `null`, `undefined`, zero-length buffers — collapsed to `null`, so `executeQuery` falls back to `new SQL.Database()` (a fresh in-memory db, which is what the host meant for `:memory:`)

No Node-side behavior change. Fixes `:memory:` invocations under affected Bun builds.

## Test plan

- [x] `pnpm --filter just-bash exec vitest run src/commands/sqlite3/` — **157/157 pass** (6 new unit tests pin each branch of `coerceDbBuffer`, including the zero-length `ArrayBuffer` regression that previously broke)
- [x] `pnpm --filter just-bash check:worker-sync` — clean (`worker.js` rebuilds deterministically; the `.js` is gitignored so it isn't shipped in this diff)

## Background

We've been carrying this in `flowglad/just-bash` since Trigger.dev's container shipped a Bun version that exhibits the regression. The fix is orthogonal to the dot-command preprocessor work in vercel-labs#249; opening as a separate PR for cleaner review.

### DIFF
--- a/packages/just-bash/src/commands/sqlite3/worker.ts
+++ b/packages/just-bash/src/commands/sqlite3/worker.ts
@@ -28,6 +28,32 @@ import {
 // Cached SQL.js module (initialized once)
 let cachedSQL: SqlJsStatic | null = null;
 
+/**
+ * Coerce a host-supplied dbBuffer into the form sql.js expects.
+ *
+ * Why: Bun's worker_threads structured-clone has regressed across versions
+ * (notably the build shipped in Trigger.dev's container) and surfaces a
+ * host-side `null` dbBuffer as a zero-length ArrayBuffer here. A truthy
+ * empty ArrayBuffer would slip past `if (data.dbBuffer)` and reach
+ * `new SQL.Database(arrayBuffer)`, which throws "Expected ArrayBuffer for
+ * the first argument" (sql.js wants Uint8Array, not bare ArrayBuffer).
+ * Treat empty/non-Uint8Array values as "no buffer" → fresh in-memory db,
+ * matching the host's intent for :memory: databases.
+ */
+export function coerceDbBuffer(raw: unknown): Uint8Array | null {
+  if (raw instanceof Uint8Array) {
+    return raw;
+  }
+  if (
+    raw &&
+    typeof (raw as { byteLength?: unknown }).byteLength === "number" &&
+    (raw as ArrayBuffer).byteLength > 0
+  ) {
+    return new Uint8Array(raw as ArrayBufferLike);
+  }
+  return null;
+}
+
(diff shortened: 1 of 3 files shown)


packages/just-bash/src/commands/sqlite3/worker.ts
```diff
@@ -28,6 +28,32 @@ import {
...
+ * (notably the build shipped in Trigger.dev's container) and surfaces a
+ * host-side `null` dbBuffer as a zero-length ArrayBuffer here. A truthy
+ * empty ArrayBuffer would slip past `if (data.dbBuffer)` and reach
+ * `new SQL.Database(arrayBuffer)`, which throws "Expected ArrayBuffer for
+ * the first argument" (sql.js wants Uint8Array, not bare ArrayBuffer).
+ * Treat empty/non-Uint8Array values as "no buffer" → fresh in-memory db,
+ * matching the host's intent for :memory: databases.
+ */
+export function coerceDbBuffer(raw: unknown): Uint8Array | null {
+  if (raw instanceof Uint8Array) {
```
cramforce, line 44: Explain how this change is not a security risk

subsetpark: What's the vector you've got in mind?

cramforce: It's a sensitive area of code going from simple to complicated, so it is worth asking

PR #251, opened 2026-05-20
feat(date): use local timezone by default, delegate to strftime

trieloff:
~~The comments in the code indicated that using UTC in the sandbox is desirable, but I'm not sure if that is actually what you want, @cramforce or just agent cope.~~

As the UTC masking is desired, I've updated the PR to default on that. Set $TZ to something to get local time.

I'm running `just-bash` in the browser and get annoyed by my agent being unable to tell the hour (it always gets the minute right), so this PR aligns it more closely with `date` on Linux.

Replace the hand-rolled UTC-only format loop with a call to the existing `formatStrftime()` from `printf/strftime.ts`, which already handles local timezone, TZ env var, and the full set of Linux format specifiers.

- `date` now shows time from `$TZ` as local time by default; `-u` still forces UTC, if no `$TZ` env var is set, stay with UTC.
- TZ env var is respected for per-command timezone override
- Adds @timestamp parsing (`date -d @1705276800`, GNU extension)
- Gains `%A %B %C %D %G %g %j %k %l %N %r %U %V %W %x %X` at no cost
- Fixes `strftime.ts`: hour/minute/second = 0 would fall back to local time via `||` operator instead of using the `Intl`-resolved value

### DIFF
--- a/THREAT_MODEL.md
+++ b/THREAT_MODEL.md
@@ -174,6 +174,7 @@ The following components are **trusted** and outside the scope of just-bash's ru
 | process.channel | IPC channel access | Blocked in **worker contexts only** (WorkerDefenseInDepth); main thread skipped for same reason | `src/security/blocked-globals.ts` |
 | Host PID/UID | Expose process identity | Virtualized (processInfo option, defaults: pid=1, uid=1000) | `src/Bash.ts` |
 | hostname/whoami/uname | System enumeration | Return generic/virtual values | `src/commands/hostname/` |
+| Host timezone | `date` leaks host TZ via `%Z`/`%z` or time values | Defaults to UTC; only honored when the host explicitly sets `$TZ` to an IANA zone | `src/commands/date/date.ts` |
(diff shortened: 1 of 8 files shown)


cramforce:
Using UTC is desired to avoid leaking information from the environment into the sandbox. Happy to accept a way to set the timezone from the outside

trieloff:
I'll check if I can read (and set) `$TZ` env var. Then you'll get default behavior out of the box, but you can set local time if you want to.

trieloff:
@cramforce default is now UTC, with $TZ overrides possible.

cramforce:
This looks good, and I'll merge. Would you mind adding a doc reference for this?

trieloff:
Will do.

trieloff:
https://github.com/vercel-labs/just-bash/pull/251/changes/bbf4956f270e2038db46e2474a7e47f288d6a211 has the documentation updates. I aimed for something terse and surgical so your docs don't bloat.

cramforce:
Perfect. Unfortunately, that last commit wasn't signed. Could you repush?

trieloff:
Done

cramforce:
I think you need to force-push harder :)

There is still a commit in the PR that is not signed

trieloff:
All verified.

trieloff:
Sorry for the back and forth, I hope this additional check would prevent this in the future https://github.com/vercel-labs/just-bash/pull/263

cramforce (approved):

PR #260, opened 2026-06-01
feat(jq): add raw input mode (-R/--raw-input)

trieloff:
## Summary

Adds `-R` / `--raw-input` support to `jq` (including the common `-Rs` combination), so input can be read as raw strings instead of JSON.

This **builds on #189** by @nicoster, which has gone stale and is now `CONFLICTING`. The reasons #189 can't merge as-is:

1. **Monorepo refactor (#198)** moved everything from `src/...` to `packages/just-bash/src/...`, so #189 edits paths that no longer exist on `main`.
2. **UTF-8 input refactor (#233)** reworked the stdin/file read path (`decodeBytesToUtf8`), changing the insertion context #189 relied on.
3. **A correctness bug:** #189 processed each input file independently. Real jq concatenates all inputs into a single stream and splits lines across file boundaries.

## What changed vs #189

- Rebased onto the current `packages/just-bash` layout.
- **Fixed the file-boundary bug:** inputs are concatenated *before* line splitting. With `a.txt`=`first` (no trailing newline) and `b.txt`=`second\n`, real jq emits `"firstsecond"` (one line spanning the boundary), now matched exactly.
- `-Rs` slurps the entire concatenated input into a single JSON string.
- Trailing newline does not yield a final empty string; interior blank lines are preserved; `\r` is kept (jq splits only on `\n`).

## Testing

- `pnpm typecheck`, `pnpm lint:fix`, `pnpm lint:banned`, `pnpm knip` — all clean
- `pnpm test:run src/commands/jq/jq.raw-input.test.ts src/commands/jq/jq.test.ts` — 47 passing
- `pnpm test:comparison src/comparison-tests/jq.raw-input.comparison.test.ts` — 8 passing, fixtures recorded against `jq-1.7.1`

New coverage: stdin line-by-line, blank lines, `-Rs` slurp, multi-file reads, the cross-file-boundary case, the `-` stdin marker mixed with files, CRLF preservation, empty input, and `-Rr`.

Closes the gap that made e.g. `jq -Rs '{action:"load-code", code: .}' < file.js` fail with `jq: invalid option -- 'R'`.

Credit to @nicoster for the original implementation and tests (#189).

### DIFF
--- a/packages/just-bash/src/commands/jq/jq.raw-input.test.ts
+++ b/packages/just-bash/src/commands/jq/jq.raw-input.test.ts
@@ -1,0 +1,123 @@
+import { describe, expect, it } from "vitest";
+import { Bash } from "../../Bash.js";
+
+describe("jq raw input (-R)", () => {
+  it("should read stdin lines as strings with -R", async () => {
+    const env = new Bash();
+    const result = await env.exec("printf 'a\\nb\\n' | jq -R '.'");
+    expect(result.stdout).toBe('"a"\n"b"\n');
+    expect(result.stderr).toBe("");
+    expect(result.exitCode).toBe(0);
+  });
+
+  it("should preserve blank lines with -R", async () => {
+    const env = new Bash();
+    const result = await env.exec("printf 'a\\n\\nb' | jq -R '.'");
+    expect(result.stdout).toBe('"a"\n""\n"b"\n');
+    expect(result.stderr).toBe("");
+    expect(result.exitCode).toBe(0);
+  });
+
+  it("should slurp raw stdin into a single string with -Rs", async () => {
+    const env = new Bash();
+    const result = await env.exec("printf 'a\\nb\\n' | jq -Rs '.'");
+    expect(result.stdout).toBe('"a\\nb\\n"\n');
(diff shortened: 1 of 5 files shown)


trieloff:
Addressed both review comments in 4e5328e:

1. **Help text mismatch** — updated to track `jq --help`: usage is now `jq [OPTIONS] FILTER [FILE...]`, the `-c` long flag reads `--compact-output` (which is what the parser already accepts, and matches real jq), and `-R` is described as "read each line as string instead of JSON".
2. **Memory usage** — raw input is now scanned incrementally, carrying only the unterminated trailing fragment across inputs, instead of allocating the full concatenated string plus a complete array of lines. File-boundary behavior (a line spanning a missing trailing newline) is preserved and still covered by the comparison fixtures.

cramforce (approved):

cramforce:
Please re-push with signed commits!

trieloff:
Done — both commits are now GPG-signed and show as Verified (the earlier ones only had a `Signed-off-by` DCO trailer, not a cryptographic signature). Force-pushed: `567958b`, `51e316a`.

Issue #110, opened 2026-02-11
UTF-8 text becomes mojibake when reading files via cat/jq (likely "binary" read path)


Haoxincode:
## Package / Version
- `just-bash`: `2.9.6`
- Observed via integration in `agentfs-sdk` + custom wrapper

## Summary
Non-ASCII UTF-8 text (e.g. Chinese) is corrupted when commands read from files (`cat`, `jq`), while direct stdout from `printf` is correct.

## Minimal Reproduction

```bash
node --input-type=module -e "
import { Bash } from 'just-bash';
const b = new Bash();

const direct = await b.exec('printf \"中文测试\\n\"');
console.log('direct:', JSON.stringify(direct.stdout)); // expected: 中文测试

await b.exec('printf \"中文测试\\n\" > t.txt');
const fromFile = await b.exec('cat t.txt');
console.log('cat:', JSON.stringify(fromFile.stdout)); // actual: mojibake

await b.exec('cat > t.jsonl <<\\'EOF\\'\n{\"summary\":\"集成AI生图功能\"}\nEOF');
const jq = await b.exec('jq -r \".summary\" t.jsonl');
console.log('jq:', JSON.stringify(jq.stdout)); // actual: mojibake
"
```

## Expected Behavior
`cat` / `jq` should output valid UTF-8 text from UTF-8 files.

## Actual Behavior
Outputs mojibake like `éæAI...`.

## Likely Root Cause
In bundled helper for file-reading commands, files are read with `"binary"` and then treated as JS string, causing UTF-8 bytes to be interpreted as Latin-1.

Reference from distributed bundle:
- `node_modules/just-bash/dist/bin/chunks/chunk-5WFYIUU2.js:2`
  - `t.fs.readFile(h, "binary")`

## Impact
- Any non-ASCII text read from file is corrupted.
- Text filtering with localized strings (e.g. Chinese names in `jq`) fails.

## Suggested Fix
- For text-oriented command paths, avoid `"binary"` string mode and read as UTF-8.
- If binary safety is required, keep bytes through the pipeline and decode explicitly at the final text boundary.
- The fix should be at the exec() output boundary (where stdout/stderr are returned to the caller), not at the readFile call site. This preserves the byte-transparent internal pipeline while correctly decoding the final output as UTF-8.

PR #265, opened 2026-06-05
fix(interpreter): decode each statement's stdout before concatenating to fix UTF-8 mojibake

### LINKED ISSUE #247: [BUG] Mixed-string boundaries bypass UTF-8 decoding, leading to Mojibake
When a string contains a mixture of both **proper Unicode characters** (e.g. `> 0xFF` from a script literal like `"你好: "`) and **Latin-1 binary-encoded characters** (e.g. `0x80 - 0xFF` representing UTF-8 bytes from a command substitution like `$(cat file)`), the boundary decoders incorrectly bypass decoding, leaving the output garbled.

## Reproducer

Run the following two simple commands:

```bash
echo "世界" > /tmp/world.txt
echo "你好: $(cat /tmp/world.txt)"
```

* **Expected Output:**
  ```text
  你好: 世界
  ```

* **Actual Output:**
  ```text
  你好: ä¸–ç•Œ
  ```

Image

gerkensm:
## Problem

When a single `exec()` call runs multiple statements that produce different output shapes, non-ASCII text in the byte-shaped output comes out as mojibake.

Two shapes exist internally:

- **Text-shaped** (sed, awk, echo): a decoded JS-Unicode string where `ö` is one char `U+00F6`
- **Byte-shaped** (grep | head, cat, tee): a latin1 byte-view string where `ö` is two chars `0xC3 0xB6`

`executeScript` concatenated each statement's `stdout` raw. The single UTF-8 decode at the `exec()` output boundary then ran over the combined string. Because the text-shaped half contributed a bare `0xF6` byte (not valid UTF-8 on its own), `TextDecoder` with `{ fatal: true }` threw, and the decoder fell back to the raw string. The byte-shaped half never got decoded.

### Reproduction

```ts
const b = new Bash({ files: { "/doc.txt": "Köpenicker\n" } });
const r = await b.exec('sed -n 1p /doc.txt\ngrep Köpenicker /doc.txt | head -1');

// actual:   "Köpenicker\nKÃ¶penicker\n"
// expected: "Köpenicker\nKöpenicker\n"
```

Same failure with `;` or `&&` as the separator.

## Fix

Decode each statement's `stdout` to UTF-8 text in isolation before concatenating. `decodeBytesToUtf8` short-circuits on already-decoded text (any char > `0xFF` means it's already Unicode), so the existing boundary decode in `Bash.ts` stays idempotent.

The same `executeScript` path is used for command substitution bodies (`$(cat file)`), so this also fixes non-ASCII mojibake in command substitutions like `echo "你好: $(cat /world.txt)"`. Closes #247.

**Why not use `stdoutAsBytes` / `stdoutKind` metadata here:** byte-passthrough commands (`head`, `tail`, `cat`, `tee`) don't set `stdoutKind` on their result — it defaults to `"text"`. A metadata-based normalization approach double-encodes their already-byte output (verified: it regresses `grep | head`).

## Tests

9 new cases in `encoding-pipeline.test.ts`:

- Primary bug: newline- and semicolon-separated `sed` + `grep|head` (the exact repro above)
- Regression guards for standalone text and byte producers
- CJK (3-byte) and emoji (4-byte) codepoints across the interleave
- Command substitution with `cat` and with `grep`
- Binary file round-trip to confirm non-UTF-8 bytes are unaffected

Relates to #110 (established the `exec()` boundary as the correct decode point — this PR handles a case that single-boundary decoding can't cover).

### DIFF
--- a/.changeset/fix-statement-utf8-interleave.md
+++ b/.changeset/fix-statement-utf8-interleave.md
@@ -1,0 +1,13 @@
+---
+"just-bash": patch
+---
+
+interpreter: fix mojibake when a script interleaves text-output and byte-output statements
+
+When a single `exec()` call ran multiple statements that produced different output "shapes" — e.g. `sed` (text-shaped: `ö` as U+00F6) followed by `grep | head` (byte-shaped: `ö` as bytes 0xC3 0xB6) — the non-ASCII text in the byte-shaped portion came out as mojibake (`KÃ¶penicker` instead of `Köpenicker`).
+
+The root cause: `executeScript` concatenated each statement's raw stdout string before the single UTF-8 decode at the `exec()` output boundary. The text-shaped half contributed a bare `0xF6` byte (not valid UTF-8 on its own), making the combined byte stream invalid, so the boundary decoder fell back to returning the raw string. The byte-shaped half then never got decoded.
+
+The fix decodes each statement's stdout to UTF-8 text in isolation, before concatenating. The decoder is a no-op on already-decoded text (chars > 0xFF short-circuit it), so the existing boundary decode in `Bash.ts` stays idempotent and nothing else changes.
+
+The same `executeScript` path is called for command substitution bodies (`$(cat /file)`), so this also fixes the analogous case where a byte producer inside `$(...)` produced mojibake when its output was spliced into an outer string (e.g. `echo "你好: $(cat /file)"`).
--- a/packages/just-bash/src/encoding-pipeline.test.ts
+++ b/packages/just-bash/src/encoding-pipeline.test.ts
@@ -206,6 +206,68 @@ echo 한 | f`,
     });
   });
 
+  describe("interleaved text + byte statements in one exec() call", () => {
+    // sed is text-shaped (ö as U+00F6); grep|head is byte-shaped (ö as 0xC3 0xB6).
+    // When concatenated raw, the lone 0xF6 makes the combined stream invalid UTF-8
+    // and the boundary decoder in Bash.ts bails, leaving the byte half as mojibake.
+    it("newline-separated text + byte statements round-trip non-ASCII", async () => {
+      const env = new Bash({ files: { "/doc.txt": "Köpenicker\n" } });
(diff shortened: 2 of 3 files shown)


Jueast:
I ran into a similar mojibake issue downstream. Thanks for working on this! I tried applying this PR downstream and it fixed the command-substitution case for us.
I noticed one related compound-statement case where the mixed text/byte output seems to happen before the result gets back to `executeScript`.

Reproduction:
```bash
printf 'Hello 世界\n' > /世界.txt
for f in /世界.txt; do
  echo "--- $(basename "$f")"
  grep Hello "$f" | head -n 1
done
 ```

 Expected:

 ```text
--- 世界.txt
Hello 世界
 ```

 Observed:

 ```text
--- 世界.txt
Hello ä¸–ç•Œ
 ```

 Inspired by this PR, the local change that fixed it for us was:

 ```diff
   diff --git a/packages/just-bash/src/interpreter/interpreter.ts b/packages/just-bash/src/interpreter/interpreter.ts
   @@ -443,7 +443,7 @@ export class Interpreter {
          if (operator === "||" && exitCode === 0) continue;

          const result = await this.executePipeline(pipeline);
   -      stdout += result.stdout;
   +      stdout += decodeBytesToUtf8(unsafeBytesFromLatin1(result.stdout));
          stderr += result.stderr;
          exitCode = result.exitCode;
          lastExecutedIndex = i;
 ```

 A possible regression test:

 ```ts
it("compound statements decode mixed text and byte pipeline output", async () => {
     const env = new Bash({ files: { "/世界.txt": "Hello 世界\n" } });
     const r = await env.exec(
       'for f in /世界.txt; do echo "--- $(basename "$f")"; grep Hello "$f" | head -n 1; done'
     );

     expect(r.stdout).toBe("--- 世界.txt\nHello 世界\n");
});
 ```

 Hope this extra case helps!

cramforce (approved):

cramforce:
Closing in favor of https://github.com/vercel-labs/just-bash/pull/272

PR #268, opened 2026-06-09
fix(jq): handle nested double-quoted strings in \(...) interpolation

trieloff:
`<human-comment>`
I ran into this issue with Opus 4.7 driving `gh` and `jq`, and stumbling over the string interpolation. I can see that there were test skips for this scenario, so this PR addresses my failure, and the existing skipped test. Paid extra attention to make sure every commit is verified, this time.
`</human-comment>`

`<bot-comment>`

> ## Summary

> Fixes a `jq` parse error when a string interpolation `"\(...)"` contains a **nested double-quoted string** — e.g. `"\(.m | sub("T.*"; ""))"` or `"\(.m | ltrimstr("ab"))"`. These previously failed with `jq: parse error: Unexpected token EOF` (or `Unexpected character ...`).

> Discovered while running an agent workflow that summarized merged PRs with:

```
jq -r '.[] | "\(.m | sub("T.*"; "")) #\(.n) @\(.u): \(.t)"'
```

> ## Root cause

> The jq tokenizer in `packages/just-bash/src/commands/query-engine/parser.ts` scanned a `"..."` string literal up to the next `"` **without tracking `\(...)` interpolation depth**. The `"` belonging to a nested string inside the interpolation (the quotes in `sub("T.*";"")`) was mistaken for the end of the outer string, so the interpolation was truncated and the remaining characters became orphaned tokens. A secondary issue: `parseStringInterpolation` balanced parens by counting `(`/`)` without skipping nested string literals, so a paren inside a nested string would also miscount.

> The pipe `|`, the `.n` field, and `@tsv` were **not** the cause — those all work fine; only nested quotes inside `\(...)` were affected.

> ## What changed

> Two surgical edits in `query-engine/parser.ts`:

> 1. The tokenizer's string branch now tracks `\(...)` interpolation depth and consumes nested `"..."` literals verbatim (with `\X` escape handling) instead of terminating the outer string early.
> 2. `parseStringInterpolation` skips over nested string literals when balancing parentheses, so the full interpolation source is captured and handed to the expression parser intact.

> ## Testing

> - New collocated test `packages/just-bash/src/commands/jq/jq.string-interp-nested-quotes.test.ts` (7 cases): `sub`, `+ "!"`, `ltrimstr`, the full transcript filter, a paren-in-nested-string regression guard, plus control cases (`"\(.n + 1)"`, `@tsv`). All expectations validated against system `jq 1.7.1`.
> - `pnpm test:run src/commands/jq/ src/commands/query-engine/` — 341 passing, no regressions.
> - `pnpm typecheck`, `pnpm lint:fix`, `pnpm knip` — clean.
> - Spot-checked 16 cases against real jq (escaped quotes inside nested strings, multiple/adjacent/recursive interpolations) — all match exactly.

> The original transcript command now produces byte-identical output to real jq.

> A `patch` changeset is included.

`</bot-comment>`

### DIFF
--- a/.changeset/jq-nested-interp.md
+++ b/.changeset/jq-nested-interp.md
@@ -1,0 +1,9 @@
+---
+"just-bash": patch
+---
+
+jq: allow nested double-quoted strings inside `"\(...)"` string interpolation
+
+jq string interpolation of the form `"\(...)"` that contained a nested double-quoted string — for example `"\(sub("T.*";""))"` or `"\(ltrimstr("ab"))"` — previously failed with a parse error. The tokenizer terminated the outer string at the first `"` it saw inside the interpolation expression, so the rest of the expression became orphaned tokens.
+
+The lexer now tracks `\(...)` depth while consuming a string literal and treats nested `"..."` pairs as opaque content while inside an interpolation, restoring them verbatim into the captured interpolation source. `parseStringInterpolation` similarly skips over nested strings when balancing parentheses, so the interpolation expression is captured as a whole and handed to the expression parser intact.
--- a/packages/just-bash/src/commands/jq/jq.string-interp-nested-quotes.test.ts
+++ b/packages/just-bash/src/commands/jq/jq.string-interp-nested-quotes.test.ts
@@ -1,0 +1,114 @@
+import { describe, expect, it } from "vitest";
+import { Bash } from "../../Bash.js";
+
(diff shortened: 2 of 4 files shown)


cramforce (approved):

cramforce:
Would you mind re-pushing with a signed commit

trieloff:
@cramforce done.

PR #269, opened 2026-06-09
fix(security): set env vars directly on target in blocking proxy set trap

mutewinter:
I wrote this:

We tried to use the `file` command and were getting `'process.env' only accepts a configurable, writable, and enumerable data descriptor`. Had an agent (Fable 5) track it down along with the other bug in #270.

SCR-20260609-oibg

---

Agent wrote this:

**Root cause:** `createBlockingObjectProxy`'s `set` trap forwards `receiver` (the proxy itself) to `Reflect.set`. Per ordinary `[[Set]]` semantics, when `receiver !== target` an assignment to an *existing* key degrades into `receiver.[[DefineOwnProperty]]` with a value-only descriptor. The `defineProperty` trap forwards that partial descriptor to the real `process.env`, whose interceptor rejects anything but a full data descriptor — so any plain `process.env.X = y` for a pre-existing key throws Node's `ERR_INVALID_OBJECT_DEFINE_PROPERTY` even when the box is *not* blocking.

**Trigger in practice:** `debug` (via `supports-color`, lazily loaded by `file`/`file-type`) runs `process.env.DEBUG = namespaces` in `save()` when `DEBUG` is set in the host env. With the box proxy installed (it installs on every `Bash#exec` since `defenseInDepth` defaults to true), the write throws and `file` fails.

**Change:** the `set` trap now sets directly on the target (`Reflect.set(target, prop, value)`), in both `defense-in-depth-box.ts` and `worker-defense-in-depth.ts`. The blocking path is unchanged — it throws before reaching this. New regression test covers writing a pre-existing env key while the box is activated but outside `run()` (fails on main, passes with the fix).

### DIFF
--- a/packages/just-bash/src/security/defense-in-depth-box.test.ts
+++ b/packages/just-bash/src/security/defense-in-depth-box.test.ts
@@ -899,6 +899,36 @@ describe("DefenseInDepthBox", () => {
 
         handle.deactivate();
       });
+
+      it("should only allow writing process.env.DEBUG outside sandbox context", () => {
+        // Regression: the proxy's set trap forwarded `receiver` (the proxy)
+        // to Reflect.set, so assignment to an existing key degraded into a
+        // value-only defineProperty on the real process.env, which Node
+        // rejects with "'process.env' only accepts a configurable, writable,
+        // and enumerable data descriptor". Seen via debug/supports-color
+        // (`process.env.DEBUG = ...`) loaded lazily by the `file` command.
+        const originalDebug = process.env.DEBUG;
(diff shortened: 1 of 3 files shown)


cramforce:
Can we limit this only to the DEBUG key. I'd rather not allow this at all

mutewinter:
Updated and force-pushed the branch. The proxy workaround is now limited specifically to writes to `process.env.DEBUG`; all other environment keys retain the previous behavior. I also updated the regression test to verify both cases.

cramforce (approved):

PR #272, opened 2026-06-14
fix(interpreter): decode each statement's stdout via stdoutKind to fix UTF-8 mojibake

### LINKED ISSUE #247: [BUG] Mixed-string boundaries bypass UTF-8 decoding, leading to Mojibake

trieloff:
## Problem

A single `exec()` can interleave **text-shaped** statements (`sed`, `awk`, `echo` — `ö` as `U+00F6`) with **byte-shaped** ones (`grep | head`, `cat` — `ö` as bytes `0xC3 0xB6`). `Interpreter.executeScript` / `executeStatement` concatenated each statement's raw `stdout`, so the lone high byte from the text half made the combined stream invalid UTF-8. The single decode at the `exec()` output boundary (`Bash.ts`) then bailed (`{ fatal: true }` throws → returns raw), leaving the byte half as Latin-1 mojibake.

```ts
const b = new Bash({ files: { "/doc.txt": "Köpenicker\n" } });
await b.exec("sed -n 1p /doc.txt\ngrep Köpenicker /doc.txt | head -1");
// actual:   "Köpenicker\nKÃ¶penicker\n"
// expected: "Köpenicker\nKöpenicker\n"
```

The same `executeScript` path backs command substitution, so `echo "你好: $(cat /file)"` and `VAR=$(cat utf8); printf %s "$VAR"` were affected too.

## Fix

The library already ships the right signal — `stdoutKind` / `stdoutAsBytes` (introduced in #233) — and the pipe glue (`pipeline-execution.ts`) already uses it. The gap was only that the **statement/script/command-substitution concat sites ignored it**.

1. **`decodedTextFromResult(result)`** (`encoding.ts`): normalize a result to text using its explicit `stdoutKind` (byte-shaped → UTF-8 decode once, falling back to raw latin1 for non-UTF-8; text-shaped → unchanged, never re-decoded). It consults the metadata, never the characters, so text whose latin1 code units merely *look* like UTF-8 (`Ã¶`) is preserved and shell control flow doesn't change.
2. **`executeScript` + `executeStatement`** decode each statement/pipeline result with that helper before concatenating. (2 lines.)

## Per-output shape, not per-command

The shape is a property of each **output**, not of the command, so it's declared at the point of production:

- The byte-passthrough producers already self-mark (`cat`/`head`/`tail`/`tee`/`gzip` via the legacy `stdoutEncoding` alias; `cut`/`base64`/`tar`/`uniq`/`join` via `stdoutKind`).
- This PR adds the two that were missing it: **`tac`** (only the stdin path forwards raw bytes — the file path reads decoded text and stays text) and **`curl`** (the response body). This per-output precision is exactly what a per-command label can't express: `tac`'s file path is text while its stdin path is bytes.

I verified completeness by temporarily making `ExecResult.stdoutKind` *required* and letting `tsc` enumerate every return site (your "make the issue visible in the type system" approach from #233), confirming `tac` + `curl` were the only byte-passthrough outputs not already self-marking, then relaxing the field back to optional so nothing breaks for external command authors.

## Tests

`encoding-statement-interleave.test.ts`: interleave (newline / `;`), 3-byte CJK + 4-byte emoji, `$(...)` byte producers, the `VAR=$(cat); printf` report, the for-loop body case, the `echo KÃ¶penicker` SAME non-regression guard, the lone-Latin-1 `Köpenicker` guard, and a raw-binary passthrough guard.

`encoding-issue-957.test.ts`: the downstream reproducers from ai-ecoverse/slicc#957 — `$(cat file)` em-dash capture (asserting `e2 80 94`, not the reported `c3 a2 c2 80 c2 94`), no triple-encode when the capture is re-wrapped, `head -c`/`tail -c` byte-window alignment, a real-world multi-codepoint payload round-tripped to disk, plus the issue's "NOT broken" guards (heredoc literal, `cat | sed`, `printf`/`echo`).

## Relation to #265

#265 found this same bug first and correctly identified `executeScript` as a decode point. This PR differs in two ways that matter:

- **Metadata instead of char-sniffing.** #265 decodes every statement with `decodeBytesToUtf8(unsafeBytesFromLatin1(stdout))`, inferring the shape from the code units. That misfires on text whose Latin-1 code units happen to form valid UTF-8: `$(echo KÃ¶penicker)` gets folded to `Köpenicker`, flipping `[ "$(echo KÃ¶penicker)" = KÃ¶penicker ]` from `SAME` to `DIFF` — a real control-flow change. Keying on the explicit `stdoutKind` leaves text untouched and only decodes outputs that declare themselves byte-shaped.
- **Both concat sites.** #265 patches `executeScript` only; the review comment on #265 (thanks @Jueast) shows a compound-statement case — `grep | head` inside a `for` body — that still mojibakes because it flows through the `executeStatement` `&&`/`||` loop. This PR fixes both sites and includes that exact `for`-loop repro as a regression test.

#265's note that the metadata path "double-encodes byte output" was accurate when written, but #233/#264 since made the byte-passthrough commands (`cat`/`head`/`tail`/`tee`) self-declare their shape via the legacy `stdoutEncoding` alias; this PR fills the last two gaps (`tac` stdin, `curl` body) so the metadata is now complete. Closes #247 (as #265 also does); intended to supersede #265.

Two cases in `encoding-statement-interleave.test.ts` are concrete #265 differentiators — they pass here and fail under #265's char-sniffing approach (verified by reconstructing it locally): **`does NOT decode text-shaped output whose chars look like mojibake`** (#265 folds `$(echo KÃ¶penicker)` to `Köpenicker`, flipping the `[ … = KÃ¶penicker ]` test to `DIFF`) and **`decodes byte output inside a for-loop body (Jueast case)`** (#265 patches `executeScript` only, so the `for`-body `grep | head` still mojibakes through `executeStatement`).

> Opened as draft for your review of the `decodedTextFromResult` boundary and the `tac`/`curl` per-output markers.

### DIFF
--- a/.changeset/byte-text-output-shape.md
+++ b/.changeset/byte-text-output-shape.md
@@ -1,0 +1,20 @@
+---
+"just-bash": patch
+---
+
+interpreter: fix UTF-8 mojibake when a script interleaves text-output and byte-output statements
+
+A single `exec()` can interleave text-shaped statements (sed, awk, echo — `ö`
+as `U+00F6`) with byte-shaped ones (grep | head, cat — `ö` as bytes
+`0xC3 0xB6`). `executeScript` / `executeStatement` concatenated each result's
+raw stdout, so the lone high byte from the text half made the combined stream
+invalid UTF-8, the output-boundary decoder bailed, and the byte half came back
+as Latin-1 mojibake (`KÃ¶penicker` for `Köpenicker`). The same path backs
+command substitution, so `echo "你好: $(cat /file)"` was affected too.
+
+The fix decodes each statement/pipeline result to text via its explicit
+`stdoutKind` (`decodedTextFromResult`) before concatenating — no guessing from
+string contents, so text whose code units merely look like UTF-8 (`Ã¶`) is
+preserved. `tac` (stdin path) and `curl` (response body) now declare
+`stdoutKind: "bytes"` on the results that forward raw bytes, so the decode is
+driven per output rather than by inspecting characters.
--- a/packages/just-bash/src/commands/curl/curl.ts
+++ b/packages/just-bash/src/commands/curl/curl.ts
@@ -322,7 +322,10 @@ export const curlCommand: Command = {
         }
       }
 
-      return { stdout: output, stderr: "", exitCode: 0 };
+      // The response body is a latin1-shaped byte buffer (see
+      // `fetchBodyToStdoutString`); any prepended headers / verbose markers are
+      // ASCII, so the whole stream is byte-shaped and must be marked "bytes".
+      return { stdout: output, stderr: "", exitCode: 0, stdoutKind: "bytes" };
     } catch (error) {
       const message = getErrorMessage(error);
 
--- a/packages/just-bash/src/commands/tac/tac.ts
+++ b/packages/just-bash/src/commands/tac/tac.ts
@@ -37,7 +37,9 @@ async function tacExecute(
     }
   }
 
-  // Read from stdin. tac is byte-clean — splits on \n then reverses.
+  // Read from stdin. tac is byte-clean — splits on \n then reverses, so the
+  // stdin path forwards the original latin1 bytes and must be marked "bytes"
+  // (the file path above reads decoded text and stays text-shaped).
   const lines = latin1FromBytes(ctx.stdin).split("\n");
   if (lines[lines.length - 1] === "") {
     lines.pop();
@@ -47,6 +49,7 @@ async function tacExecute(
     stdout: reversed.length > 0 ? `${reversed.join("\n")}\n` : "",
     stderr: "",
     exitCode: 0,
+    stdoutKind: "bytes",
   };
 }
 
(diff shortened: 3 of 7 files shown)


trieloff:
Why is `curl` classified as bytes? Shouldn't it be dynamic?

gerkensm:
Great work @trieloff. Shall I close #265 in favour of this? There's also a valuable comment, but I think it might already be covered by your PR.

trieloff:
I'd leave the call to @cramforce. My agent looked at @Jueast's comment and considered it, and looking at the thumbs up from them, I think we should be good.

cramforce (approved):

cramforce:
Thanks both! I'll merge this one but I had already approved the other one!

PR #278, opened 2026-06-26
chore(deps): bump node-liblzma optional dep ^2.0.3 → ^5.0.0

### LINKED ISSUE #274: node-liblzma: latest version may work better as the optional dep
Hi! 👋 I maintain [`node-liblzma`](https://github.com/oorabona/node-liblzma), which `just-bash` pulls in (optionally) for `.xz` support.

I noticed the optional dep is pinned to `^2.0.3`. The current **5.x** should be a smoother experience — recent releases fixed native builds on Windows (VS 2026) and pnpm/mise-managed setups, and ship prebuilt binaries for Linux/macOS/Windows, so the "may fail on some systems" case your lazy-load guards against should hit far less often. The API you use (`xzSync` / `unxzSync`) is unchanged.

One thing to note: 5.x raises the minimum to **Node ≥ 22** (2.x was ≥ 16). Since it's an optional dependency with your existing fallback, Node < 22 users would just keep degrading gracefully rather than breaking — but flagging it in case you still target older Node.

Happy to send a one-line PR if you'd like to try it. Thanks for building just-bash! 🙏

oorabona:
Bumps the optional `node-liblzma` dependency from `^2.0.3` (which resolves to the 4-major-old 2.2.0) to `^5.0.0`.

## Why

- 5.x ships prebuilt binaries for Linux/macOS/Windows and fixed native builds on Windows + Visual Studio 2026 and pnpm/mise-managed runtimes — so this optional dep (which `tar` lazy-loads, and whose comment notes "may fail on some systems") is far less likely to fail to load.
- It also fixes a WASM stream-decode crash on Node 26 / Deno.

## Compatibility

- The API `just-bash` uses (`xzSync` / `unxzSync`) is unchanged in 5.x.
- **Verified:** the full `src/commands/tar` suite — **170 tests**, including the xz source and bundled-binary paths — passes with `node-liblzma@5.0.3`.

## One trade-off

5.x raises the engine floor from Node ≥16 to **Node ≥22**. As an optional dependency with the existing lazy-load fallback, Node <22 users degrade gracefully (xz unavailable, with the current error) rather than break.

Closes #274. Happy to adjust the range (e.g. accept both 2.x and 5.x) if you'd prefer to keep Node <22 xz support.

### DIFF
--- a/packages/just-bash/package.json
+++ b/packages/just-bash/package.json
@@ -123,6 +123,6 @@
   },
   "optionalDependencies": {
     "@mongodb-js/zstd": "^7.0.0",
-    "node-liblzma": "^5.0.0"
+    "node-liblzma": "^5.1.1"
   }
 }
--- a/packages/just-bash/package.json
+++ b/packages/just-bash/package.json
@@ -123,6 +123,6 @@
   },
   "optionalDependencies": {
     "@mongodb-js/zstd": "^7.0.0",
-    "node-liblzma": "^2.0.3"
+    "node-liblzma": "^5.0.0"
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -246,8 +246,8 @@ importers:
         specifier: ^7.0.0
         version: 7.0.0
       node-liblzma:
-        specifier: ^2.0.3
-        version: 2.2.0
+        specifier: ^5.0.0
+        version: 5.0.3
 
   packages/just-bash-executor:
     dependencies:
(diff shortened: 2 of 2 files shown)


cramforce (approved):

cramforce:
It currently doesn't compile

oorabona:
I think I found the root cause, node-liblzma is in your onlyBuiltDependencies allowlist, so pnpm runs its install scripts.

The package shipped `gypfile: true` together with a postinstall build, which makes pnpm do a from-source node-gyp rebuild at install time. On runners that don't have the native toolchain (CMake, liblzma headers) that build fails, and because the dependency is optional pnpm just silently drops it, which is where the TS2307: Cannot find module 'node-liblzma' comes from.

I fixed it in node-liblzma 5.1.1 by switching to the canonical prebuildify setup (install: node-gyp-build, no gypfile), so pnpm uses the shipped prebuild directly. No source build, no toolchain needed. I've bumped this PR to require ^5.1.1.

I validated it locally on Node 24 with pnpm install --frozen-lockfile and the dependency allowlisted, the same way your CI runs it: before the fix, node-gyp rebuild runs and fails without a toolchain; after it, there's no source build at all, the prebuild is used, the module loads, and tsc passes with no TS2307.

The Actions on this PR are sitting in action_required since it comes from a fork, so once you approve the run, 22 and 24 should come back green.

One heads up on Node 20: node-liblzma 5.x needs Node >= 22, and that floor is intentional. The native build relies on node-gyp 13 for the Node 26 Windows toolchain, and node-gyp 13 itself drops Node < 22.22.2 ([node-gyp v13.0.0 release notes](https://github.com/nodejs/node-gyp/releases/tag/v13.0.0)), so supporting Node 20 and Node 26 on Windows in the same toolchain isn't possible (and Node 20 has been EOL since April anyway). So the Node 20 legs will stay red on that engine floor.

Totally your call how you want to handle it, and I'm happy to drop Node 20 from the CI matrix in this PR if that's useful, or leave it to you.

oorabona:
Hey @cramforce, just a friendly nudge in case this fell off your radar. Feel free to reach out if there's anything I can help with!

PR #281, opened 2026-06-29
fix(rg): search piped stdin when no paths are given

### LINKED ISSUE #280: `rg` ignores piped stdin and searches cwd instead
## Summary

When `rg` receives piped stdin with no explicit path arguments, it ignores stdin and defaults to recursively searching the current working directory (`.`). Real ripgrep searches stdin when it detects piped input and no path is given.

`grep` in just-bash already handles this correctly.

## Reproduction

```bash
# rg ignores stdin, searches cwd instead
echo "hello world" | rg "hello"
# Expected: hello world
# Actual: recursively searches . and returns all file matches containing "hello"

printf "line1 foo\nline2 bar\nline3 foo\n" | rg "foo"
# Expected: line1 foo\nline3 foo
# Actual: recursively searches . for "foo"

# grep handles stdin correctly (for comparison)
echo "hello world" | grep "hello"
# Output: hello world  ✅
```

## Root Cause

In `src/commands/rg/rg-search.ts`, `executeSearch` defaults to the current directory when no paths are given:

```ts
// Default to current directory
const paths = inputPaths.length === 0 ? ["."] : inputPaths;
```

It never checks `ctx.stdin`. Compare with `grep.ts` which correctly handles this:

```ts
// If no files and stdin is provided (including empty string), read from
// stdin.
if (files.length === 0 && ctx.stdin !== undefined) {
  const result = searchContent(decodeBytesToUtf8(ctx.stdin), regex, { ... });
  ...
}
```

## Suggested Fix

Before defaulting to `.`, check if stdin is available (similar to grep):

```ts
// If no paths and stdin is provided, search stdin
if (inputPaths.length === 0 && ctx.stdin !== undefined) {
  // search ctx.stdin content with the regex
  // return results
}

// Otherwise default to current directory
const paths = inputPaths.length === 0 ? ["."] : inputPaths;
```

## Impact

This is particularly problematic when the working directory maps to a large filesystem (e.g. a remote context base mount) — `rg` recursively walks the entire tree, producing huge output or timing out, when the user only intended to search a small piped input.

mohan-garimella:
## Summary

When `rg` receives piped stdin with no explicit path arguments, it ignores stdin and defaults to recursively searching the current working directory (`.`). This is inconsistent with real ripgrep and with how `grep` already handles stdin in just-bash.

## Reproduction

```bash
# rg ignores stdin, searches cwd instead
echo "hello world" | rg "hello"
# Expected: hello world
# Actual: recursively searches . and returns all file matches containing "hello"

# grep handles stdin correctly
echo "hello world" | grep "hello"
# Output: hello world ✅
```

## Fix

After building the regex, check for non-empty stdin before defaulting to `.`. When stdin has content and no paths are given, `searchContent` is called directly on the stdin text with all applicable options (invert, count, context, quiet, `-l`, etc.).

Also fixes `rg.utf8-stdin.test.ts` which was previously passing by accident — it tested `cat /in.txt | rg '한글'` but rg was actually finding the match by searching the filesystem (it found `in.txt`), not by reading stdin.

## Tests

- 7 new stdin-specific tests covering: basic piped search, no-match exit code, case-insensitive, inverted match, count mode, explicit-path-overrides-stdin, and multibyte UTF-8
- All 515 existing rg tests pass (82 skipped)
- All 205 existing grep tests pass (8 skipped)

Fixes #280

### DIFF
--- a/packages/just-bash/src/commands/rg/rg-search.ts
+++ b/packages/just-bash/src/commands/rg/rg-search.ts
@@ -116,9 +116,6 @@ export async function executeSearch(
     };
   }
 
-  // Default to current directory
-  const paths = inputPaths.length === 0 ? ["."] : inputPaths;
-
   // Determine case sensitivity
   const effectiveIgnoreCase = determineIgnoreCase(options, patterns);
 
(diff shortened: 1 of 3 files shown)


mohan-garimella:
Addressed in 15f5d15 — added a `stdinConsumedByPatternFile` guard that skips the stdin search path when `-f -` was used. Also added a test for `echo "hello" | rg -f - /data.txt` to ensure patterns from stdin are used against the explicit file, not searched as content.

cramforce:
Please sign commits

mohan-garimella:
@cramforce Done, Please review.

cramforce (approved):

PR #284, opened 2026-06-30
feat(sandbox): forward capability options through Sandbox.create

### LINKED ISSUE #283: Sandbox.create drops BashOptions capability flags (python, javascript, commands, customCommands, fetch)
## Summary

`Sandbox.create(opts: SandboxOptions)` silently drops the capability flags that
`new Bash(opts: BashOptions)` accepts. A host that drives just-bash through the
`Sandbox` API (rather than constructing `Bash` directly) therefore cannot enable
`python`, `javascript` (js-exec), a restricted `commands` set, `customCommands`, or a
custom `fetch` — even though the bundled runtimes ship in the package.

## Where it happens

`SandboxOptions` (packages/just-bash/src/sandbox/Sandbox.ts) exposes only a subset of
`BashOptions`: `cwd, env, timeoutMs, fs, overlayRoot, maxCallDepth, maxCommandCount,
maxLoopIterations, network, defenseInDepth`. And `Sandbox.create` constructs the
internal `Bash` with exactly those fields:

```ts
// packages/just-bash/src/sandbox/Sandbox.ts
static async create(opts?: SandboxOptions): Promise<Sandbox> {
  let fs: IFileSystem | undefined = opts?.fs;
  if (opts?.overlayRoot) {
    if (opts?.fs) throw new Error("Cannot specify both 'fs' and 'overlayRoot' options");
    fs = new OverlayFs({ root: opts.overlayRoot });
  }
  const bashEnv = new Bash({
    env: opts?.env,
    cwd: opts?.cwd,
    fs,
    maxCallDepth: opts?.maxCallDepth,
    maxCommandCount: opts?.maxCommandCount,
    maxLoopIterations: opts?.maxLoopIterations,
    network: opts?.network,
    defenseInDepth: opts?.defenseInDepth,
  });
  return new Sandbox(bashEnv, opts?.timeoutMs);
}
```

There is no `python` / `javascript` / `commands` / `customCommands` / `fetch` in either
`SandboxOptions` or the `new Bash({...})` literal, so the `Bash` constructor's
capability registration (which keys off `options.python`, `options.javascript`,
`options.commands`, `options.customCommands`, `options.fetch`) never fires for a
sandbox created via `Sandbox.create`. The TypeScript type also rejects passing the
flags as excess properties, so there is no escape hatch short of `new Bash(...)`.

This is identical in the published 3.0.1 and 3.0.2 tarballs (the compiled
`dist/sandbox/Sandbox.d.ts` and `dist/bundle` `Sandbox.create` are byte-equivalent on
this surface).

## Motivating use case

The eve framework (vercel/eve) ships a just-bash sandbox backend that creates the
interpreter via `Sandbox.create({ cwd, env, fs, network })`. Because of the gap above,
that backend cannot offer `python3` (or js-exec) to authored agents at all — the
bundled CPython is present but unreachable — and users fall back to Docker /
microsandbox just to run a stdlib-only Python helper. The natural fix on the host side
is a one-line passthrough, but it has nothing to forward to: the capability flags do
not exist on `SandboxOptions`. See vercel/eve#431 for the downstream report.

This generalizes beyond Python: any of the optional `BashOptions` capabilities
(`javascript`/js-exec, a restricted `commands` allow-list, host-defined
`customCommands`, a custom `fetch`) is equally unreachable through `Sandbox`.

## Proposed change (backward compatible)

Add the capability fields to `SandboxOptions` and forward them into the internal
`new Bash({...})` in `Sandbox.create`. Behavior is unchanged when a caller omits them
(each defaults to just-bash's existing default — Python/js-exec stay off, full command
set stays available). Concretely, forward `python`, `javascript`, `commands`,
`customCommands`, and `fetch` (the remaining `BashOptions` capability inputs that
`Sandbox.create` does not already pass).

I have a backward-compatible patch + tests (a `customCommands`/`commands` forwarding
test in the default suite, plus a `Sandbox.create({ python: true })` end-to-end test in
the python3/wasm suite) and will open a draft PR referencing this issue. Filing the
issue first to confirm the direction and the exact field set you'd want on
`SandboxOptions`.

## Versions

just-bash 3.0.1 / 3.0.2.

arimxyer:
## What

`Sandbox.create(opts: SandboxOptions)` builds its internal `Bash` from only a subset of
`BashOptions`, silently dropping the optional capability flags. This PR adds those flags
to `SandboxOptions` and forwards them into the `new Bash({...})` that `Sandbox.create`
constructs:

- `python?: boolean`
- `javascript?: boolean | JavaScriptConfig`
- `commands?: CommandName[]`
- `customCommands?: CustomCommand[]`
- `fetch?: SecureFetch`

(`env`, `cwd`, `fs`/`overlayRoot`, `maxCallDepth`, `maxCommandCount`, `maxLoopIterations`,
`network`, `defenseInDepth` were already forwarded.)

## Why

A host that drives just-bash through the `Sandbox` API rather than `new Bash(...)` could
not enable python3, js-exec, a restricted command set, custom commands, or a custom
fetch — even though the runtimes ship in the package. The flags exist only on
`BashOptions`, and `Sandbox.create` did not pass them through, so they were unreachable
(and the TypeScript type rejected them as excess properties).

Concretely, vercel/eve's just-bash sandbox backend constructs the interpreter via
`Sandbox.create(...)`, so it cannot offer `python3` to agents at all and users fall back
to Docker/microsandbox to run a stdlib-only Python helper. This is the root cause behind
that downstream report: **vercel/eve#431** (and eve's consumer-side passthrough that
extends to these flags once this lands).

Fixes #283.

## Backward compatibility

Behavior is unchanged when a caller omits the new fields: each is forwarded as
`undefined`, falling back to its existing `BashOptions` default (Python/js-exec stay off,
the full command set stays available, fetch derives from `network` as before). No
existing `SandboxOptions` field changes.

## Tests

- `src/sandbox/Sandbox.test.ts` (default unit suite): two new cases proving forwarding
  without WASM — a `customCommands` command runs through `Sandbox.create`, and a
  restricted `commands: ["echo"]` allow-list registers only the listed command.
- `src/commands/python3/python3.sandbox.test.ts` (WASM suite): `Sandbox.create({ python:
  true })` registers `python3` (and it stays unavailable by default) — the direct
  end-to-end proof.

## Validation

- `pnpm typecheck` — clean
- `pnpm exec vitest run --config vitest.unit.config.ts src/sandbox/Sandbox.test.ts` — pass
- `pnpm exec vitest run --config vitest.wasm.config.ts src/commands/python3/python3.sandbox.test.ts` — pass
- `pnpm lint`, `pnpm knip`, `pnpm build` — clean
- A changeset is included (`minor`).

The main open design question is the exact field set surfaced on `SandboxOptions` — I
forwarded the five `BashOptions`-only capability inputs (`python`, `javascript`,
`commands`, `customCommands`, `fetch`), but I'm happy to narrow or reshape that to match
your preference.

### DIFF
--- a/.changeset/sandbox-forward-capability-options.md
+++ b/.changeset/sandbox-forward-capability-options.md
@@ -1,0 +1,17 @@
+---
+"just-bash": minor
+---
+
+sandbox: forward capability flags from `SandboxOptions` into the underlying `Bash`
+
+`Sandbox.create(opts)` previously constructed its internal `Bash` with only a subset
+of `BashOptions`, silently dropping the optional capability flags (`python`,
+`javascript`, `commands`, `customCommands`, `fetch`). A host that drives just-bash
+through the `Sandbox` API (rather than `new Bash(...)`) therefore could not enable
+python3, js-exec, a restricted command set, custom commands, or a custom fetch — even
+though the runtimes ship in the package.
+
+`SandboxOptions` now exposes those fields and `Sandbox.create` forwards them into the
+`Bash` it builds. Behavior is unchanged when a caller omits them (each falls back to
+its existing `BashOptions` default — Python/js-exec stay off, the full command set
+stays available). Fixes the root cause behind vercel/eve#431.
--- a/packages/just-bash/src/commands/python3/python3.sandbox.test.ts
+++ b/packages/just-bash/src/commands/python3/python3.sandbox.test.ts
@@ -1,0 +1,25 @@
+import { describe, expect, it } from "vitest";
+import { Sandbox } from "../../sandbox/Sandbox.js";
+
+// End-to-end proof that the `python` capability flag forwarded through
+// `SandboxOptions` reaches the underlying `Bash` and actually registers the
+// python3 command. Without forwarding, python3 stays unavailable even though
+// the CPython runtime ships in the package (see vercel/eve#431).
+describe("Sandbox.create({ python: true })", () => {
+  it(
+    "registers python3 when the capability is enabled",
+    { timeout: 60000 },
+    async () => {
+      const sandbox = await Sandbox.create({ python: true });
+      const cmd = await sandbox.runCommand("python3 --version");
+      expect(await cmd.stdout()).toContain("Python 3.");
+    },
+  );
+
+  it("leaves python3 unavailable by default", async () => {
+    const sandbox = await Sandbox.create();
+    const cmd = await sandbox.runCommand("python3 --version");
+    // Either "not found" or "not available" depending on the bundle context.
+    expect(await cmd.stderr()).toMatch(/command not (found|available)/);
+  });
+});
--- a/packages/just-bash/src/sandbox/Sandbox.test.ts
+++ b/packages/just-bash/src/sandbox/Sandbox.test.ts
@@ -1,5 +1,6 @@
 import { PassThrough } from "node:stream";
 import { describe, expect, it } from "vitest";
+import { defineCommand } from "../custom-commands.js";
 import { Command, type OutputMessage, Sandbox } from "./Sandbox.js";
 
 describe("Sandbox API", () => {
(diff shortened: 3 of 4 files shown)


cramforce:
Why do you use the Sandbox API. It really makes little sense if you need specific options

arimxyer:
@cramforce, It's not a choice I can make as a consumer — `vercel/eve`'s just-bash backend constructs the interpreter via `Sandbox.create(...)` internally (it relies on the Sandbox session/filesystem lifecycle), so a host driving eve never reaches the `Bash` constructor. And these flags aren't just un-forwarded — they're not on `SandboxOptions` at all, so there's no way to pass them through that API today. That's the root of vercel/eve#431.

This PR is the minimal fix from just-bash's side. The alternative is eve switching its backend to `new Bash(...)`, but that's a change on eve's end. Happy to go that route instead if you'd rather keep `Sandbox` capability-free — just wanted to surface that the gap is real for anyone consuming just-bash *through* eve.

cramforce (approved):

PR #285, opened 2026-07-02
refactor(interpreter): model stdin as a single shared StdinStream

a0preetham:
Fixes stdin position tracking in loops and compound commands:

```bash
while IFS= read -r line; do
  echo "$line" | cat   # cat consumed all remaining stdin — loop stopped after first line
done <<< $'a\nb\nc'

# Before: a <-- stops here
# After:  a
#         b
#         c
```

`groupStdin` was a string snapshot. This fix makes stdin stream-like: one shared object that tracks the offset already read, so any consumer advances the position for everyone.

## What

Replaces the interpreter's `IOState.groupStdin?: string` snapshot with a single `StdinStream` object — the only representation of stdin anywhere in the interpreter. It works like a bash file descriptor: content bytes plus a read offset, shared by reference. **Reading is consuming**: a command that drains stdin advances the offset for every other holder of the stream, including across subshell and pipeline-stage boundaries.

> Note: this supersedes the earlier `StdinCursor` design on this branch. That version kept stdin in two parallel forms (a threaded `stdin: string` parameter plus a cursor in interpreter state) with precedence rules at every consumer, and required each of ~30 commands to remember to call `consumeStdin()` — a convention-enforced invariant. The redesign makes the invariant structural: `ctx.stdin.readAll()` is the only way to get the content, so "read stdin but forgot to advance it" is impossible by construction.

## Why

The old string-snapshot approach had a recurring bug class: commands inside a `while read` loop body (pipelines, `cat`, `grep`, `tr`, …) consumed the loop's stdin without advancing the shared position, so the next iteration re-read the same input. The fallback rules (`stdin || groupStdin`) were duplicated across `builtin-dispatch`, `read`, `mapfile`, and the pipeline glue, and `pipeline-execution.ts` cleared `groupStdin` without restoring it.

## Design

- **`src/stdin-stream.ts`** — `StdinStream` with `readAll()` (consume), `peek()` (forward without consuming; used by `timeout`/`time`/`env`/`bash`), `advance(n)` (partial consumers: `read`, `mapfile`).
- **`src/interpreter/stdin-redirect.ts`** — one resolver for `<`, `<<`, `<<-`, `<<<`, `<&`, shared by simple commands, **all** compound commands (`if`/`for`/`while`/`until`/`case`/subshell/group), and function-def redirects; plus `withStdin()` for scoped install/restore. This replaces four diverged copies of the redirect-processing loop and fixes the heredoc UTF-8 byte-encoding those copies were missing.
- **Pipeline stages** install a fresh stream per stage; the first stage inherits the enclosing stream by reference (bash fd-inheritance semantics). Empty pipe output can no longer fall back to the enclosing scope's stdin.
- **`CommandContext.stdin` is now a `StdinStream`** (breaking, TypeScript): commands call `ctx.stdin.readAll()`; `StdinStream` is exported from the package root. A changeset (`minor`) documents the migration.

Behavior fixes verified against real bash: subshells share the stdin offset with the parent; function definition redirects apply per call and win over piped stdin; a missing redirect file on a function errors without running the body; `diff - -` compares stdin against itself; `cat - -` reads stdin once.

## Test coverage

- **`src/interpreter/while-loop-stdin.test.ts` — 40 tests** covering stdin redirects (`<`, `<<`, `<<-`, `<<<`, file-not-found) on all six compound commands, pipelines inside loop bodies, and `read` position tracking. Carried over from the previous design and passing **unchanged** — the rewrite preserves all of its behavior.
- **`src/comparison-tests/stdin-consumption.comparison.test.ts` — 15 new comparison tests** with fixtures recorded from real bash: consumption by `cat`/`grep`/`tr` in loop and group bodies, subshell offset sharing, pipeline-stage isolation, piped-loop position keeping, function redirect-vs-pipe precedence, missing-file redirect errors, `cat - -`, `diff - -`, empty-pipe no-fallback, nested loop redirects, and `eval` consuming group stdin.
- **`src/stdin-stream.test.ts` — 12 unit tests** for the stream primitive itself (consume/peek/advance/exhaustion/shared-reference semantics).
- Full suite: `pnpm typecheck`, lint, knip, spec tests (3825 passed), comparison tests (553 passed), and unit tests (13k+) all green. The only failures in a full run are pre-existing python3/WASM tests caused by unmaterialized git-LFS blobs in the local checkout (identical on `main`).

### DIFF
--- a/.changeset/stdin-stream.md
+++ b/.changeset/stdin-stream.md
@@ -1,0 +1,39 @@
+---
+"just-bash": minor
+---
+
+interpreter: model stdin as a single shared stream (`StdinStream`), fixing loops and compound commands losing stdin position
+
+Stdin is now represented exactly one way everywhere: a `StdinStream` object
+that behaves like a bash file descriptor — it holds the content bytes and a
+read offset, and is shared by reference. Reading is consuming: a command
+that drains stdin (`cat`, `grep`, `sed`, ...) advances the offset for every
+other holder of the stream, including across subshell and pipeline-stage
+boundaries. This replaces the old `groupStdin?: string` snapshot plus
+per-call-site fallback rules, which made it easy for a command to read the
+loop's stdin without advancing it.
+
+Fixes, all verified against real bash via recorded comparison fixtures:
+
+- Commands inside `while read` loop bodies (pipelines, `cat`, `grep`, `tr`,
+  and ~30 other stdin consumers) now advance the loop's stdin instead of
+  re-reading the same input each iteration.
+- Stdin redirections (`<`, `<<`, `<<-`, `<<<`, `<&`) now work uniformly on
+  all compound commands (`if`, `for`, `while`, `until`, `case`, subshells,
+  groups) through one shared resolution path.
+- Subshells share the stdin offset with their parent (matching bash fd
+  semantics): `{ (read x); read y; } < f` gives `y` the second line.
+- A pipeline stage that produces empty output no longer lets the next
+  command fall back to the enclosing scope's stdin.
+- Heredoc/here-string content on compound commands and functions is now
+  UTF-8 encoded into the byte pipeline consistently (previously only simple
+  commands did this).
+- Function definition redirects (`f() { ...; } < file`) now apply on every
+  call and win over piped stdin; a missing redirect file is an error and
+  the body does not run (both match bash).
+
+BREAKING (TypeScript API): `CommandContext.stdin` is now a `StdinStream`
+instead of a `ByteString`. Custom commands call `ctx.stdin.readAll()` to
+consume input (or `ctx.stdin.peek()` to inspect without consuming) and
+convert with `latin1FromBytes` / `decodeBytesToUtf8` as before. The
+`StdinStream` class is exported from the package root.
--- a/packages/just-bash/README.md
+++ b/packages/just-bash/README.md
@@ -37,9 +37,10 @@ const hello = defineCommand("hello", async (args, ctx) => {
 });
 
 const upper = defineCommand("upper", async (args, ctx) => {
-  // ctx.stdin is a ByteString — decode to text before string ops.
+  // ctx.stdin is a stream: readAll() consumes it and returns a ByteString —
+  // decode to text before string ops.
   return {
-    stdout: decodeBytesToUtf8(ctx.stdin).toUpperCase(),
+    stdout: decodeBytesToUtf8(ctx.stdin.readAll()).toUpperCase(),
     stderr: "",
     exitCode: 0,
   };
@@ -51,7 +52,7 @@ await bash.exec("hello Alice"); // "Hello, Alice!\n"
 await bash.exec("echo 'test' | upper"); // "TEST\n"
 ```
 
-Custom commands receive a `CommandContext` with `fs`, `cwd`, `env`, `stdin`, and `exec` (for subcommands), and work with pipes, redirections, and all shell features.
+Custom commands receive a `CommandContext` with `fs`, `cwd`, `env`, `stdin` (a consumable stream: `readAll()` to drain it, `peek()` to inspect without consuming), and `exec` (for subcommands), and work with pipes, redirections, and all shell features.
 
 <details>
 <summary><h2>Supported Commands</h2></summary>
--- a/packages/just-bash/src/Bash.ts
+++ b/packages/just-bash/src/Bash.ts
@@ -62,6 +62,7 @@ import {
   SecurityViolationError,
 } from "./security/defense-in-depth-box.js";
 import type { DefenseInDepthConfig } from "./security/types.js";
+import { StdinStream } from "./stdin-stream.js";
 import { serialize } from "./transform/serialize.js";
 import type {
   BashTransformResult,
@@ -374,6 +375,7 @@ export class Bash {
     this.state = {
       env,
       cwd,
+      stdin: new StdinStream(),
       previousDir: "/home/user",
       functions: new Map<string, FunctionDefNode>(),
       localScopes: [],
@@ -648,7 +650,9 @@ export class Bash {
       // UTF-8 bytes. Callers that already prepared a byte buffer (e.g.
       // `Buffer.from(buf).toString("latin1")`) opt into raw passthrough
       // via `stdinKind: "bytes"`.
-      groupStdin: encodeStdinForPipeline(options?.stdin, options?.stdinKind),
+      stdin: new StdinStream(
+        encodeStdinForPipeline(options?.stdin, options?.stdinKind) ?? "",
+      ),
       // Cooperative cancellation signal (used by timeout command)
       signal: options?.signal,
       // Extra arguments injected directly into first command's arg list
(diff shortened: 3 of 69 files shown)


a0preetham:
@cramforce Claude seems to write loops and the pipes within a loop do not work currently.

cramforce:
Please run changeset, sign the commit, and revisit test coverage

a0preetham:
Done!

✅ Changeset — [.changeset/while-loop-stdin-cursor.md](https://github.com/a0preetham/just-bash/blob/refactor/stdin-cursor/.changeset/while-loop-stdin-cursor.md) with patch bump
✅ Signed commit — SSH-signed, GitHub shows verified: true
✅ Test coverage — 4 new tests + grep cursor-advancement fix

a0preetham:
Updated with the requested changes. Here's a summary of what's in the latest push:

**Changeset** — `.changeset/while-loop-stdin-cursor.md` added (patch bump).

**Additional fixes surfaced by test coverage:**
- `executeUntil` was missing all stdin/redirection handling — `until ... done < file` now works correctly
- `executeFor` had the same gap — `for i in ...; do read ...; done < file` now works
- `grep` (and by extension `rg`) read `ctx.stdin` directly without advancing the shared cursor — added `ctx.stdinCursor?.readAll()` to the stdin path

**Test coverage** (25 tests across 2 files, all passing):
- `src/stdin-cursor.test.ts` — 11 unit tests covering `StdinCursor` directly (advance, readAll, exhausted, edge cases)
- `src/interpreter/while-loop-stdin.test.ts` — 14 integration tests covering: simple pipelines, multi-stage pipelines, nested loops, group commands, heredocs, `stdin` API option, `cat`/`grep` reading from loop stdin, `for` loop with stdin redirect, `until` loop

**Coverage for changed files** (unit suite, WASM tests excluded):

| File | Stmts | Branch |
|---|---|---|
| `stdin-cursor.ts` | 100% | 100% |
| `utils/file-reader.ts` | 100% | 100% |
| `commands/grep/grep.ts` | 81.4% | 80.8% |
| `interpreter/control-flow.ts` | 83.8% | 70.0% |

The gaps in `grep.ts` and `control-flow.ts` are in pre-existing unrelated paths (regex error branches, case/if handling), not in the stdin-cursor code paths.

a0preetham:
Follow-up push: extended stdin-redirect support to all remaining compound commands and fixed a subtle `async` bug.

### What changed

**New: stdin redirect support for `if`, C-style `for`, word-list `for`, `until`, `case`**

All five now process `<`, `<<`, `<<<` redirections and install a `StdinCursor`, so `read` inside the body sees the redirected content:

```bash
# Now works correctly in all compound commands
if true; then
  IFS= read -r a
  IFS= read -r b
  echo "$a $b"
fi < /file          # "line1 line2"

for ((i=0; i<3; i++)); do
  IFS= read -r line
  echo "$i: $line"
done < /file        # "0: x\n1: y\n2: z"

case 1 in
  1) IFS= read -r a; IFS= read -r b; echo "$a $b";;
esac < /file        # "hello world"
```

**Bug fix: `return await` in `executeIf`**

`executeIf` had `return executeStatements(...)` (no `await`) inside a `try-finally`. In JS, `return promise` inside `try-finally` fires the `finally` synchronously — before the promise resolves — so the cursor was restored to `undefined` before `read` could run. Changed to `return await executeStatements(...)`.

### Coverage (8141 tests, 377 files)

| Metric | Value |
|--------|-------|
| Statements | 61.25% (26318/42963) |
| Branches | 56.77% (17426/30695) |
| Functions | 64.19% (2456/3826) |
| Lines | 61.85% (25257/40835) |

Key files:
- `stdin-cursor.ts`: **100%** statements/branches/lines (11 unit tests)  
- `control-flow.ts`: 76% statements, 63% branches

18 new tests added covering: `if` condition/body with stdin redirect, C-style `for` body, `case` body, word-list `for` body, `until` loop with pipeline.

a0preetham:
**Coverage before/after comparison** (unit suite, WASM/security tests excluded):

#### Key files

| File | Stmts before | Stmts after | Branch before | Branch after |
|------|-------------|-------------|---------------|--------------|
| `stdin-cursor.ts` | 100% | 100% | 100% | 100% |
| `utils/file-reader.ts` | 100% | 100% | 100% | 100% |
| `commands/grep/grep.ts` | 81.4% | 81.4% | 80.8% | 80.8% |
| `interpreter/control-flow.ts` | 83.8% | 76.1% | 70.0% | 63.1% |

> `control-flow.ts` coverage % dropped because the follow-up push added ~200 lines of stdin-redirect handling for `if`, `for`, `until`, `case`, and C-style `for`. The 18 new tests cover the `<` file-redirect path in each; `<<` and `<<<` paths in those constructs are untested (pre-existing branches in `while` aren't fully covered either).

#### Overall suite (tests only, not line counts changed)

| | Before | After |
|--|--------|-------|
| Test files | 359 | 377 |
| Tests passing | 7940 | 8141 |
| New tests | — | +201 |

a0preetham:
**Updated coverage** — added `<<` (heredoc) and `<<<` (herestring) tests for every compound command (`if`, word-list `for`, C-style `for`, `until`, `case`).

#### Key files — before → after

| File | Stmts before | Stmts after | Branch before | Branch after |
|------|-------------|-------------|---------------|--------------|
| `stdin-cursor.ts` | 100% | 100% | 100% | 100% |
| `utils/file-reader.ts` | 100% | 100% | 100% | 100% |
| `commands/grep/grep.ts` | 81.4% | 81.4% | 80.8% | 80.8% |
| `interpreter/control-flow.ts` | 83.8% | **82.75%** | 70.0% | **74.76%** |

Branch coverage on `control-flow.ts` is now higher than before the PR (70% → 74.76%). Statement % is slightly down (83.8% → 82.75%) because the new redirect-handling code added more lines than the new tests cover — remaining gaps are `<<-` (strip-tabs heredoc) and error paths for missing files in the new constructs.

#### Overall

| | Initial PR | +fix commit | +heredoc/herestring tests |
|--|-----------|------------|--------------------------|
| Tests passing | 7940 | 8141 | **8151** |
| Test files | 359 | 377 | 377 |

a0preetham:
**Final coverage update** — added `<<-` (strip-tabs) and file-not-found error path tests for all six compound commands.

#### `control-flow.ts` progression

| Stage | Stmts | Branches | Functions |
|-------|-------|----------|-----------|
| Before this PR | 83.8% | 70.0% | — |
| After StdinCursor refactor | 76.1% | 63.1% | — |
| + if/for/until/case fix + `<`/`<<`/`<<<` tests | 82.75% | 74.76% | — |
| + `<<-` strip-tabs + file-not-found tests | **90.18%** | **78.5%** | **100%** |

#### Overall (8163 tests, 377 files)

| Metric | Value |
|--------|-------|
| Statements | 61.38% (26372/42963) |
| Branches | 56.88% (17460/30695) |
| Functions | 64.34% (2462/3826) |
| Lines | 61.98% (25311/40835) |

Remaining uncovered branches in `control-flow.ts` are pre-existing edge cases unrelated to this PR: `failglob` mode in `for` loops, `break`/`continue` inside a `while` condition expression, and the `loopError="error"` path in nested loop error handling.

a0preetham:
@vercel run a review

a0preetham:
@cramforce not sure what the process here is. I cannot seem to run CI or see suggestions. I would love some feedback on this. Thanks!

cramforce:
This is a very large PR. I approved tests to run given that @vercel did not find immediate issues. However, I'm not sure I can ever find the time to land this

a0preetham:
No worries! I will close this then. Think treating stdin and stdout as stream like things instead of strings is cleaner in the longer term. But I understand the change+maintenance burden. Thanks for the review!

PR #292, opened 2026-07-08
fix(interpreter): seed cd dash from OLDPWD exec option

skovranek:
## Summary

- seed the per-exec interpreter `previousDir` from `OLDPWD` when callers pass an env snapshot into `exec()`
- add regression coverage for `cd -` across isolated `exec()` calls using persisted `cwd` and `env`

## Why

`exec()` supports per-call `cwd` and `env`, which lets callers preserve shell state between isolated executions. `cd` updates `OLDPWD` in the returned env, but `cd -` reads the interpreter's internal `previousDir` state instead.

That means a caller can pass the returned `PWD` and `OLDPWD` into the next `exec()` call and still get stale `cd -` behavior.

## Verification

- `pnpm --filter just-bash test:run src/Bash.exec-options.test.ts src/interpreter/builtins/cd.test.ts`
- `pnpm --filter just-bash test:run src/spec-tests/bash/spec.test.ts -t 'cd - without OLDPWD'`
- `pnpm lint`
- `pnpm knip`
- `pnpm --filter just-bash build`

### DIFF
--- a/packages/just-bash/src/Bash.exec-options.test.ts
+++ b/packages/just-bash/src/Bash.exec-options.test.ts
@@ -193,6 +193,29 @@ describe("exec options", () => {
       const envResult = await env.exec("echo $MODE");
       expect(envResult.stdout).toBe("dev\n");
     });
+
+    it("should use OLDPWD from per-exec env for cd dash", async () => {
+      const env = new Bash({ cwd: "/" });
+      await env.exec("mkdir -p /tmp/old /tmp/new");
+
+      const firstCd = await env.exec("cd /tmp/old", {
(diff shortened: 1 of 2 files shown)


cramforce (approved):

cramforce:
Please sign your commit

skovranek:
> Please sign your commit

@cramforce Thank you, it's verified now.

I hadn't added the key to GH for signing so it couldn't verify the sig.

Ref for others:
https://docs.github.com/en/authentication/connecting-to-github-with-ssh/adding-a-new-ssh-key-to-your-github-account?tool=cli

PR #295, opened 2026-07-10
docs: model js-exec surface in threat model, align WASM policy docs; ci: coverage report

MauricioPerera:
## Summary

Documentation and CI improvements surfaced by an external review of the repo:

- **THREAT_MODEL.md**
  - Add **§4.11 JavaScript Execution Surface (When Enabled)**, modeling the opt-in `js-exec`/QuickJS surface analogously to Python's §4.7: how it is enabled (`javascript: true` or `invokeTool`), the SharedArrayBuffer host bridge (VFS, sandbox-re-entrant `exec`, gated `fetch`, `invokeTool`), the virtual `child_process` shim, what the isolation rests on (QuickJS WASM + worker + `WorkerDefenseInDepth` + 64MB cap + timeout), and the residual risk (MEDIUM).
  - Qualify the §3.5 `child_process` row: the literal "no code path" claim now notes the *virtual* shim that exists inside QuickJS when `javascript` is enabled (re-enters the sandbox, not the host OS), pointing to §4.11.
  - Make the architectural invariant behind §4.2/§4.3 explicit — **no bash→host-JS code path** — including that it is guarded by the `check-banned-patterns` linter and that those residuals would become critical if it ever broke.
- **CLAUDE.md**: align the WASM policy with reality — "no new WASM dependencies without approval", with sql.js, `quickjs-emscripten`, and the vendored CPython WASM listed as the approved exceptions (previously it said "No WASM dependencies allowed" while two additional WASM runtimes ship in the repo).
- **README (`packages/just-bash`)**: document that `node` is an alias of the `js-exec` runtime (not a real Node.js), and list `tar`, `yq`, `xan` (plus `js-exec`/`node`) among the commands unavailable in browser builds, matching the `__BROWSER__` gate in `registry.ts`.
- **CI**: add `coverage.yml`, a non-gating workflow that runs the existing (currently unused in CI) `test:coverage` script and uploads the report as an artifact (`if: always()`, no thresholds). Setup mirrors `unit-tests.yml` (LFS, pnpm, Node 20).

All doc statements were verified against the code before writing. No production code is touched.

## Notes for reviewers

- The coverage workflow was validated structurally (YAML parse + action versions matching `unit-tests.yml`); it hasn't been executed in a CI run from this fork.
- Happy to split this into separate PRs (threat model / docs / CI) if preferred.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #296, opened 2026-07-10
ci: add weekly deep-fuzz workflow (test:fuzz:long)

MauricioPerera:
## Summary

The fuzzing suite (`src/security/fuzzing/`, 4 fuzz test files + generators + DoS/sandbox oracles) already runs on every PR inside `test:unit` — but only with the default budget (50–100 runs per test, `FUZZ_RUNS` unset). The deep run that the repo already defines, `test:fuzz:long` (`FUZZ_RUNS=10000`), never runs in any workflow.

This adds `.github/workflows/fuzz-deep.yml`:

- **Triggers**: weekly schedule (Mondays 03:00 UTC) + `workflow_dispatch` for manual runs — deliberately not on push/PR (cost).
- **Job**: single ubuntu-latest job, `timeout-minutes: 60`, setup mirroring `unit-tests.yml` (LFS, pnpm, Node 20), running `pnpm --filter just-bash test:fuzz:long`.
- A crash or oracle violation fails the workflow (no `continue-on-error`) — that's the signal.

This gives low-probability DoS/sandbox regressions (only visible with a large budget) continuous coverage without adding PR latency.

Related: #295 (separate, independent change).

### DIFF
--- a/.github/workflows/fuzz-deep.yml
+++ b/.github/workflows/fuzz-deep.yml
@@ -1,0 +1,32 @@
+name: Deep Fuzz
+
+on:
+  schedule:
+    - cron: "0 3 * * 1"
+  workflow_dispatch:
+
+jobs:
+  deep-fuzz:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          lfs: true
+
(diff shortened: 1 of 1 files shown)


cramforce:
Thanks. I don't want more workflows as they increase risk

PR #298, opened 2026-07-13
fix(interpreter): preserve executable script exit status

skovranek:
## Summary

- return an executable script's `exit` status to its parent command list
- add regression coverage for `;` and `&&` after executable scripts exit nonzero

## Why

Executable scripts run in a subshell-like context, but their internal `ExitError` was rethrown after the script state was restored. That terminated the surrounding command list, so:

```sh
./fail.sh ; echo $?

never reached echo.

The script boundary now converts ExitError into a normal command result. This preserves the exit status while allowing the parent list to apply standard ; and && behavior.

## Verification

- pnpm vitest run src/interpreter/executable-script.test.ts

### DIFF
--- a/packages/just-bash/src/interpreter/executable-script.test.ts
+++ b/packages/just-bash/src/interpreter/executable-script.test.ts
@@ -1,0 +1,40 @@
+import { describe, expect, it } from "vitest";
+import { Bash } from "../Bash.js";
+
+describe("executable scripts", () => {
+  it.each([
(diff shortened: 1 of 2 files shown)


cramforce:
Looks good. Take a look at the CI runs

skovranek:
@cramforce bump. I fixed the linting. Thank you.

cramforce (approved):

PR #302, opened 2026-07-16
fix(interpreter): handle unclosed POSIX case patterns

### LINKED ISSUE #301: case pattern with unmatched outer POSIX class bracket throws SyntaxError
### Reproduction

With `just-bash@3.1.0`:

```ts
import { Bash } from "just-bash";

const bash = new Bash();
await bash.exec(`case B in
[[:alpha:]) echo bad ;;
esac`);
```

### Actual behavior

`Bash.exec()` rejects with a host JavaScript exception:

```text
SyntaxError: Invalid regular expression: /^[[:alpha:]$/: error parsing regexp: missing closing ]
```

### Expected behavior

GNU Bash 5.2.15 returns status 0 with no output. The unmatched outer `[` is treated literally while the remaining `[:alpha:]` bracket expression is still parsed. For example, the same pattern matches `[a` but not `B` or `[x`.

A shell syntax edge case should always return a shell result rather than escape as a JavaScript exception.

### Cause

The case-pattern converter uses the first `]` after `[` as the outer class terminator. In `[[:alpha:]]`, that first bracket belongs to the inner POSIX class (`:]`). When the outer bracket is absent, the converter produces an invalid regex instead of applying the normal literal-`[` fallback.

This was found by comparing the FreeBSD sh regression corpus against a native Bash 5.2.15 reference.

obinnanwachukwu1:
Fixes #301.

## Summary

- scan shell character classes without mistaking an inner POSIX `:]`, collating `.]`, or equivalence `=]` terminator for the outer closing bracket
- preserve the Bash literal-`[` fallback when the outer bracket is unmatched
- add a regression covering both the non-match and literal-prefix match behavior
- add a patch changeset

## Verification

- `pnpm --filter just-bash exec vitest run src/interpreter/control-flow.test.ts`
- `pnpm --filter just-bash test:comparison` (560 passed)
- `pnpm --filter just-bash test:unit` (13,187 passed, 97 skipped)
- `pnpm --filter just-bash build`
- `pnpm --filter just-bash lint`
- `pnpm --filter just-bash knip`

### DIFF
--- a/.changeset/fix-unclosed-posix-case-pattern.md
+++ b/.changeset/fix-unclosed-posix-case-pattern.md
@@ -1,0 +1,5 @@
+---
+"just-bash": patch
+---
+
+Treat unmatched outer brackets around POSIX classes as literals during shell pattern matching instead of throwing a JavaScript regular-expression error.
--- a/packages/just-bash/src/interpreter/conditionals.ts
+++ b/packages/just-bash/src/interpreter/conditionals.ts
@@ -621,7 +621,7 @@ function patternToRegexStr(pattern: string, extglob: boolean): string {
     } else if (char === "?") {
       regex += ".";
     } else if (char === "[") {
-      const closeIdx = pattern.indexOf("]", i + 1);
+      const closeIdx = findPatternCharClassEnd(pattern, i);
       if (closeIdx !== -1) {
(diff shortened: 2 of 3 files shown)


cramforce:
Please sign your commit

obinnanwachukwu1:
Signed now, thanks!

PR #303, opened 2026-07-16
Pass original command to custom overrides

boramuyar:
## Summary

- provide the bundled command shadowed by a custom command as `ctx.origCommand(args)`
- preserve lazy loading and pass the original implementation the same guarded, revocable invocation context
- add focused tests and a minor changeset
- remove the previous registry exports and README additions

## Motivation

Custom commands can override bundled commands, but previously could not delegate to the original implementation. Passing that implementation through the command context supports selective wrapping without exposing the built-in registry as public API.

```ts
const cat = defineCommand("cat", async (args, ctx) => {
  // Add custom behavior here.
  return ctx.origCommand!(args);
});
```

## Validation

- focused custom-command, registration, and defense-context suites: 51 passed
- `pnpm --filter just-bash knip`
- Biome check on all changed source and test files


cramforce:
I'm a little sus. What is a concrete example where you need this?

boramuyar:
> I'm a little sus. What is a concrete example where you need this?

I have a custom filesystem wrapper where some folders (nodes) are remote whereas others are local, when running commands like grep, find, tree etc. I need to make sure some built-in commands function differently, specially the recursive ones. Instead of re-implementing all those functions from scratch, with this, I can branch out when it is needed.

boramuyar:
Can i get a response on whether if this would be merged or any alternative solutions to the problem I presented would be provided? @cramforce

cramforce:
How about we pass the built-in to the override if you are defining a custom command that has a built-in?

boramuyar:
> How about we pass the built-in to the override if you are defining a custom command that has a built-in?

That would work for me as well. How would the snippet look like? That is my intention in any case.

cramforce:
```
const cat = defineCommand("cat", (args, ctx) => ctx.origCommand?(…));
```
Something like this?

boramuyar:
So, the original command picker would be available through the context? That works for me.

boramuyar:
@cramforce I updated the implementation with your requested changes. I would appreciate if we can merge the changes, so we can use them upstream, thanks!

boramuyar:
@cramforce Can we get a resolution for this one?

cramforce (approved):

cramforce:
The PR is good to go. Please sign commits!

boramuyar:
> The PR is good to go. Please sign commits!

Done.

cramforce:
Need to rebase after #343

boramuyar:
Rebased, should we merge before the base drifts again? :)

boramuyar:
Would you guys be able to merge this?

cramforce:
@auto-maintain review

cramforce:
@boramuyar Could you take a look at the feedback above?

boramuyar:
Fixed both in 704d787: delegated built-ins now receive the full execution scope and run under their own trust boundary. Added find/xargs and security regression tests.

boramuyar:
Fixed the async follow-up in 6e80f19: forced-untrusted state now survives Promise/timer callbacks, and the original Promise liveness guard is restored. Added regressions for both paths.

boramuyar:
Reverted the broad trusted-scope follow-up in dc4e3cc. The source is back to the narrower 6e80f19 implementation; the current auto-maintain review still references reverted head bbbd413.

boramuyar:
@auto-maintain review

cramforce:
@auto-maintain review

PR #313, opened 2026-07-21
feat(fs): add allowNestedMounts to MountableFs

mutewinter:
I mount an agent's working directory into the sandbox and need one subdirectory inside it hidden. Mounting an empty filesystem over that subdirectory would do it, but `MountableFs` rejects a mount inside an existing mount — so hiding one path means wrapping the entire filesystem in a decorator that reimplements every `IFileSystem` method.

```ts
const fs = new MountableFs({ allowNestedMounts: true });
fs.mount("/work", new ReadWriteFs({ root }));
fs.mount("/work/.private", new InMemoryFs()); // shadows the subtree
```

Off by default, so the existing rejection and its error messages are unchanged. Semantics follow an OS mount: deepest mount wins, the subtree is shadowed while mounted, unmounting reveals the outer contents again, and mount order doesn't matter.

`routePath` already resolves by longest matching prefix — unreachable today, since validation guarantees at most one mount can prefix-match a path. Four sites assumed a path belongs to exactly one mount:

- `realpath` rebased onto the first matching mount, not the deepest
- `getAllPaths` listed outer entries under a nested mount point, which are shadowed and unopenable
- `cp` took a same-filesystem fast path that bypassed the facade and read the shadowed contents
- `mv` did the same on the destination side, landing part of the source beneath a mount where it's unreachable; it also now refuses to move a directory containing a mount, matching `rm`

16 new tests, plus the two existing rejection tests kept as-is for the default. Each of the four fixes was reverted individually to confirm its own test fails. Full `test:run` green, changeset included.

Happy to make nesting the default instead, or to close this if it's not a direction you want in `MountableFs`.

Authored with Opus 4.8.

### DIFF
--- a/.changeset/hungry-moons-mount-nested.md
+++ b/.changeset/hungry-moons-mount-nested.md
@@ -1,0 +1,5 @@
+---
+"just-bash": minor
+---
+
+Add `allowNestedMounts` to `MountableFs`, letting a mount sit inside another mount. With the option set, mounting `/data/private` inside an existing `/data` mount shadows that subtree the way an OS mount does: the deepest mount owning a path wins, so `/data/private/key` routes to the inner filesystem while `/data/notes.txt` stays with the outer one, and unmounting reveals the outer filesystem's contents again. This makes it possible to hide a subdirectory of a mounted filesystem behind a restricted one without wrapping the whole filesystem in a decorator. The option defaults to `false`, which keeps the existing behavior of rejecting overlapping mounts.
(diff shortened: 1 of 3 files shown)


mutewinter:
Closing this one. It was speculative when I opened it, and shadowing a subtree by mounting over it may well not be a direction `MountableFs` should go.

The branch stays up if it is ever useful. The four single-mount assumptions it corrects along the way in `realpath`, `getAllPaths`, `cp`, and `mv` are real whether or not the flag exists, so if nested mounts come up again this is a starting point rather than a rewrite. Happy to reopen it, or for someone else to pick it up.

PR #318, opened 2026-07-26
fix(interpreter): interleave duplicated streams in write order

mutewinter:
## Problem

A duplication operator points both fds at one descriptor, so what it carries is the two streams in the order they were written. Today stderr is appended after all of stdout instead:

```js
await new Bash().exec("{ echo O1; echo E1 1>&2; echo O2; echo E2 1>&2; } 2>&1");

// stdout, before: "O1\nO2\nE1\nE2\n"
// stdout, after:  "O1\nE1\nO2\nE2\n"   // what bash prints
```

Same for `> f 2>&1` into a file, and for `1>&2` onto stderr.

When the two fds reach one file from opposite directions, the same gap drops output rather than reordering it:

```bash
{ echo O1; echo E1 1>&2; echo O2; } 2> f 1>&2

# f, before: "E1\n"           # O1 and O2 are gone
# f, after:  "O1\nE1\nO2\n"   # what bash writes
```

stdout is delivered through the descriptor, then stderr is written to the file from position 0 and truncates it. That clobbering is correct for two independent opens, which is what the delivery path takes these for. There is only one open here.

This matters most when the output is being read by a model rather than a person. `2>&1` is the usual way to capture everything a script did, and an agent reading the result has only the text to reason from, so when every error migrates to the bottom, a diagnostic no longer sits next to the command that produced it. A failure in step one reads as a failure in the last step, and an agent that retries or repairs on that basis fixes the wrong thing. Ordering is the only signal it has for attributing a message to a command.

## Cause

Two things, and the second is why fixing the first alone changes nothing.

`ExecutionOutputAccumulator` appends each statement's output in the order it was produced, then splits it across `stdoutChunks` and `stderrChunks`. The sequence exists at append time and is discarded by that split, so the merge in `applyRedirections` had nothing to follow and concatenated the two strings.

The delivery site already called this out:

> stdout-then-stderr order, not the command's temporal write order: ExecResult accumulates the two streams separately, so interleaving is not recorded anywhere in the interpreter.

So this is filed against a known gap rather than a surprise. If it was left open deliberately, I'm happy to be told so and close it.

The merge then only runs where `applyRedirections` can tell both fds ended up on one descriptor, and it recognized that by sink object identity: a dup used to assign the source fd's sink object to the other fd. Pre-resolved redirections (#336) changed that. A dup now reads its source back from the fd table, so `2>&1` rebuilds an equivalent `live-stdout` sink instead of sharing fd 1's object, and `> f 2>&1` leaves fd 1 on a `file` sink while fd 2 holds a `descriptor` sink for the entry that same open produced. Identity sees two objects, takes the separate-streams path, and that path is also what clobbers `2> f 1>&2`.

## Fix

Record the sequence the accumulator already sees, carry it on `ExecResult` as `outputChunks`, and merge along it when one sink holds both fds. `stdout` and `stderr` stay authoritative: the chunks are those same strings, split by stream, and only add back the relative ordering.

Then recognize one descriptor by what makes it one, rather than by object identity alone:

- a shared sink object, since `&> f` still points both fds at a single open
- the live streams by kind, since the caller has exactly one stdout and one stderr
- a file by the fd-table entry its open produced, whose alias list names the fd that opened it

Two independent redirects to one path each open their own entry and match none of these, so they keep clobbering.

The merge falls back to stdout-then-stderr unless the recorded chunks reconstruct both pending strings exactly. A single command records no order, and the redirection list can rewrite a stream before delivery (an ambiguous-redirect message, a target that failed to open). So the merge can reorder, never invent or drop.

## Scope

Statement-level ordering only. A single command hands back two already-separated strings, so its own interleaving is still lost and `cmd 2>&1` merges that command stdout-first. There's a test pinning that as intended rather than implying it's fixed.

Two independent redirects naming one path (`> f 2> f`) are separate descriptors and keep clobbering, now keyed on the entry each open produced rather than on path equality.

Not addressed, and unchanged by this: a delivery writes its whole string from position 0 rather than at the descriptor's offset, so independent opens clobber wholesale where bash overwrites byte ranges. `> f 2> f` gives `E1` where bash gives `E1 O2`, and `exec 2> f` followed by `{ ...; } > f` gives `O1` where bash gives `E1`. Both behave identically before and after this change; they want offset-tracking sinks, which is a different piece of work.

## Tests

Eight cases in `redirections.dup-ordering.test.ts`. Five fail on `main`, checked by copying the file onto a clean worktree at `244f73c` and running it there, not inferred. The other three pin behaviour this must not change: separate streams when nothing duplicates them, the single-command fallback, and `> f 2> f` still clobbering.

Checked the dup matrix by hand against real bash as well (3.2.57, so the ordering semantics rather than anything version-specific). File contents, newlines shown as spaces:

```
                     bash         after        before
> f 2>&1             O1 E1 O2     O1 E1 O2     O1 O2 E1
2> f 1>&2            O1 E1 O2     O1 E1 O2     E1
&> f                 O1 E1 O2     O1 E1 O2     O1 O2 E1
>> f 2>&1            O1 E1 O2     O1 E1 O2     O1 O2 E1
2>&1 > f             O1 O2        O1 O2        O1 O2        (E1 to stdout, all three)
> f 2> f             E1 O2        E1           E1
exec 2> f; > f       E1           O1           O1
```

The last two are the offset gap noted above, matching `main` in both columns.

`test:run` on this branch: 570 files, 15249 passing, 97 skipped, nothing failing, plus 72 passing in `just-bash-executor`. Typecheck, Biome and knip clean, changeset included.

One thing worth naming: `src/syntax/loops.test.ts > should detect infinite for loop and error` failed once while I had two full suites running concurrently, and passes on its own and in the clean run above. It's a timing-based limit test and nothing here touches loop protection, but it does look load-sensitive.

---

Model: Claude Opus 5

### DIFF
--- a/packages/just-bash/src/execution-output.ts
+++ b/packages/just-bash/src/execution-output.ts
@@ -1,7 +1,7 @@
 import { utf8ByteLength } from "./encoding.js";
 import type { ExecutionScope } from "./execution-scope.js";
 import { ControlFlowError } from "./interpreter/errors.js";
-import type { ExecResult } from "./types.js";
+import type { ExecResult, OutputChunk } from "./types.js";
 
 /**
  * Chunked interpreter-output sink backed by the one top-level execution
@@ -11,6 +11,9 @@ import type { ExecResult } from "./types.js";
 export class ExecutionOutputAccumulator {
   private readonly stdoutChunks: string[] = [];
   private readonly stderrChunks: string[] = [];
+  // The same pieces in arrival order, which is the order the shell produced
+  // them. Splitting them across the two arrays above is what discards it.
+  private readonly orderedChunks: OutputChunk[] = [];
   private stdoutBytes = 0;
   private stderrBytes = 0;
   private readonly attachedErrors = new WeakSet<ControlFlowError>();
@@ -41,6 +44,7 @@ export class ExecutionOutputAccumulator {
     }
     if (chunk) {
       (stream === "stdout" ? this.stdoutChunks : this.stderrChunks).push(chunk);
+      this.orderedChunks.push({ stream, text: chunk });
     }
     if (stream === "stdout") this.stdoutBytes += bytes;
     else this.stderrBytes += bytes;
@@ -65,6 +69,7 @@ export class ExecutionOutputAccumulator {
         : "text";
     const stdoutBytes =
       stdoutKind === "bytes" ? stdout.length : utf8ByteLength(stdout);
+    const orderMark = this.orderedChunks.length;
     this.append(
       "stdout",
       stdout,
@@ -76,6 +81,21 @@ export class ExecutionOutputAccumulator {
       result.stderr,
       result.internalOutputAccounting?.stderr ?? 0,
     );
+    // A child that recorded its own ordering knows better than the two entries
+    // just appended, which say stdout-then-stderr for the whole child. Swap
+    // them for the finer sequence so nesting does not flatten it a level at a
+    // time. Skipped when the caller overrode stdout, since the child's chunks
+    // then describe content this result no longer carries. Byte accounting is
+    // untouched either way: it was charged by the appends above.
+    const childChunks =
+      stdout === result.stdout ? result.outputChunks : undefined;
+    if (childChunks?.length) {
+      this.orderedChunks.splice(
+        orderMark,
+        this.orderedChunks.length - orderMark,
+        ...childChunks,
+      );
+    }
   }
 
   build(exitCode: number, extra?: Partial<ExecResult>): ExecResult {
@@ -84,6 +104,9 @@ export class ExecutionOutputAccumulator {
       stderr: this.stderrChunks.join(""),
       exitCode,
       ...extra,
+      ...(this.orderedChunks.length > 0 && {
+        outputChunks: this.orderedChunks,
+      }),
       internalOutputAccounting: {
         stdout: this.stdoutBytes,
         stderr: this.stderrBytes,
--- a/packages/just-bash/src/execution-output.ts
+++ b/packages/just-bash/src/execution-output.ts
@@ -1,6 +1,7 @@
-import { utf8ByteLength } from "./encoding.js";
+import { type OutputKind, utf8ByteLength } from "./encoding.js";
 import type { ExecutionScope } from "./execution-scope.js";
 import { ControlFlowError } from "./interpreter/errors.js";
+import { chunksDescribe } from "./output-chunks.js";
 import type { ExecResult, OutputChunk } from "./types.js";
 
 /**
@@ -27,7 +28,7 @@ export class ExecutionOutputAccumulator {
     stream: "stdout" | "stderr",
     chunk: string,
     alreadyAccountedBytes = 0,
-    kind: "text" | "bytes" = "text",
+    kind: OutputKind = "text",
   ): void {
     let bytes: number;
     try {
@@ -44,7 +45,7 @@ export class ExecutionOutputAccumulator {
     }
     if (chunk) {
       (stream === "stdout" ? this.stdoutChunks : this.stderrChunks).push(chunk);
-      this.orderedChunks.push({ stream, text: chunk });
+      this.orderedChunks.push({ stream, text: chunk, kind });
     }
     if (stream === "stdout") this.stdoutBytes += bytes;
     else this.stderrBytes += bytes;
@@ -59,12 +60,27 @@ export class ExecutionOutputAccumulator {
     if (!(error instanceof ControlFlowError)) return;
     if (this.attachedErrors.has(error)) return;
     this.attachedErrors.add(error);
-    error.prependOutput(this.stdout, this.stderr);
+    error.prependOutput(this.stdout, this.stderr, this.orderedChunks);
+  }
+
+  /**
+   * Absorb the output an aborted child carried out on a control-flow error,
+   * the way a subshell swallows `exit` and keeps what ran before it.
+   */
+  appendError(error: ControlFlowError): void {
+    const orderMark = this.orderedChunks.length;
+    this.append("stdout", error.stdout, error.internalOutputAccounting.stdout);
+    this.append("stderr", error.stderr, error.internalOutputAccounting.stderr);
+    this.adoptChunks(orderMark, error.outputChunks, error.stdout, error.stderr);
   }
 
   appendResult(result: ExecResult, stdout: string = result.stdout): void {
-    const stdoutKind =
-      result.stdoutKind === "bytes" || result.stdoutEncoding === "binary"
+    // The shape of what is appended, not of what the producer emitted: a
+    // caller that hands over decoded text in place of the result's own byte
+    // buffer is appending Unicode whatever the producer's shape was.
+    const stdoutKind: OutputKind =
+      stdout === result.stdout &&
+      (result.stdoutKind === "bytes" || result.stdoutEncoding === "binary")
         ? "bytes"
         : "text";
     const stdoutBytes =
@@ -81,22 +97,33 @@ export class ExecutionOutputAccumulator {
       result.stderr,
       result.internalOutputAccounting?.stderr ?? 0,
     );
-    // A child that recorded its own ordering knows better than the two entries
-    // just appended, which say stdout-then-stderr for the whole child. Swap
-    // them for the finer sequence so nesting does not flatten it a level at a
-    // time. Skipped when the caller overrode stdout, since the child's chunks
-    // then describe content this result no longer carries. Byte accounting is
-    // untouched either way: it was charged by the appends above.
-    const childChunks =
-      stdout === result.stdout ? result.outputChunks : undefined;
-    if (childChunks?.length) {
-      // Truncate and push rather than splicing the child's chunks in: a spread
-      // passes them as arguments, which overflows the stack once the array is
-      // long enough. Nothing bounds its length, so it must not be spread.
-      this.orderedChunks.length = orderMark;
-      for (const chunk of childChunks) {
-        this.orderedChunks.push(chunk);
-      }
+    this.adoptChunks(
+      orderMark,
+      result.internalOutputChunks,
+      stdout,
+      result.stderr,
+    );
+  }
+
+  /**
+   * Replace the two coarse entries a child's append just added with the finer
+   * sequence the child recorded, so nesting does not flatten the ordering a
+   * level at a time. Byte accounting is untouched: those bytes were charged by
+   * the appends this refines.
+   */
+  private adoptChunks(
+    orderMark: number,
+    chunks: OutputChunk[] | undefined,
+    stdout: string,
+    stderr: string,
+  ): void {
+    if (!chunksDescribe(chunks, stdout, stderr)) return;
+    // Truncate and push rather than splicing the child's chunks in: a spread
+    // passes them as arguments, which overflows the stack once the array is
+    // long enough. Nothing bounds its length, so it must not be spread.
+    this.orderedChunks.length = orderMark;
+    for (const chunk of chunks) {
+      this.orderedChunks.push(chunk);
     }
   }
 
@@ -107,7 +134,9 @@ export class ExecutionOutputAccumulator {
       exitCode,
       ...extra,
       ...(this.orderedChunks.length > 0 && {
-        outputChunks: this.orderedChunks,
+        // A copy: the array behind it keeps growing if this accumulator is
+        // appended to again, and a result already handed out must not change.
+        internalOutputChunks: [...this.orderedChunks],
       }),
       internalOutputAccounting: {
         stdout: this.stdoutBytes,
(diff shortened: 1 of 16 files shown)


packages/just-bash/src/execution-output.ts
```diff
@@ -76,6 +81,21 @@ export class ExecutionOutputAccumulator {
...
+    // then describe content this result no longer carries. Byte accounting is
+    // untouched either way: it was charged by the appends above.
+    const childChunks =
+      stdout === result.stdout ? result.outputChunks : undefined;
+    if (childChunks?.length) {
+      this.orderedChunks.splice(
+        orderMark,
+        this.orderedChunks.length - orderMark,
+        ...childChunks,
+      );
```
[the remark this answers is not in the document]
mutewinter: Good catch, applied in b76c222 — truncate to the mark and push, as suggested.

Confirmed the failure in isolation: `[].splice(0, 0, ...arr)` with a 200k-element array throws `RangeError: Maximum call stack size exceeded`, while the push loop handles it fine.

I couldn't construct a repro through the public API, for what it's worth: a loop's output coalesces into a single chunk (`{ { for i in $(seq 1 100000); do echo x; done; }; }` yields `outputChunks.length === 1`), so the count tracks statements rather than iterations, and `seq` caps at 100k anyway. Reaching the limit would take ~125k literal statements in one script.

Applying it regardless, since nothing bounds the array's length by construction and the guard is free — and if chunk granularity ever gets finer, the spread would start overflowing silently.

mutewinter:
@cramforce interested in this PR? If not, I can close it. Since it's more structural I understand if you want to do it yourself.

cramforce:
@privatenumber What do you think?

privatenumber:
Yeah, I think this is worth pursuing. Its a real gap reproducible on main.

packages/just-bash/src/execution-output.ts
```diff
@@ -84,6 +106,9 @@ export class ExecutionOutputAccumulator {
       stderr: this.stderrChunks.join(""),
       exitCode,
       ...extra,
+      ...(this.orderedChunks.length > 0 && {
+        outputChunks: this.orderedChunks,
```
privatenumber: Could we add nested group and function coverage here?

applyRedirections builds a new result without forwarding outputChunks, so an inner scope can flatten before an outer 2>&1 sees it

packages/just-bash/src/interpreter/redirections.ts
```diff
@@ -1338,27 +1403,38 @@ export async function applyRedirections(
...
-      // ExecResult accumulates the two streams separately, so interleaving
-      // is not recorded anywhere in the interpreter. This matches the
-      // convention of the live-stream merge (`stdout += stderr`) used for a
-      // bare `2>&1`.
-      const combined = pendingStdout + pendingStderr;
+    if (fdsShareOneDescriptor()) {
+      // One descriptor carries both streams, so they arrive at it in the order
+      // they were written.
+      const combined = mergeInRecordedOrder(
+        result.outputChunks,
```
privatenumber: Could we carry this representation through control flow and pipelines too?

It seems if/loops/errors/|& can currently bypass the ordering?

packages/just-bash/src/types.ts
```diff
@@ -48,6 +54,18 @@ export interface ExecResult {
...
+   * produced, when the interpreter observed that order. `stdout` and `stderr`
+   * remain authoritative -- they are these pieces filtered by stream -- so this
+   * only adds back the relative ordering that keeping two strings discards.
+   *
+   * Set for a result the interpreter assembled from more than one statement: a
+   * group, a subshell, a function body. Absent for a single command, whose two
+   * streams arrive already separated, so consumers fall back to
+   * stdout-then-stderr there.
+   */
+  outputChunks?: OutputChunk[];
```
privatenumber: WDYT about keeping this metadata internal and recording encoding per chunk? 

ExecResult is public, and { stream, text } cant distinguish binary stdout from Unicode stderr when both target one sink

cramforce:
@auto-maintain review

packages/just-bash/src/execution-output.ts
mutewinter: Done in fa95a5e. `applyRedirections` now re-tags what stayed on the caller's streams and forwards it, so a scope with no redirection of its own stops flattening.

It records nothing where the upstream recorded nothing: a single command's two strings are not an observation of order, and fabricating one there broke `stdout_stderr.py |& cat` in the spec suite.

Covered by `redirections.dup-ordering.test.ts > fd duplication ordering through nested scopes`: nested groups, sibling groups, a subshell, a function body, a redirection on a function definition, repeated calls, and a nested group into a file.

packages/just-bash/src/interpreter/redirections.ts
mutewinter: You were right on all four. Done in fa95a5e.

- **if/loops/case**: `CompoundOutput` kept two strings only. It now carries the sequence, and drops to undefined the moment anything appends without one.
- **errors**: a scope left on `break`, `exit` or `return` carried its output out on the error, flattened. `ControlFlowError` carries the sequence too, and `routeControlFlowError` writes it back after merging.
- **`|&`**: put all of stderr ahead of all of stdout on the pipe. It now interleaves.
- `ConditionResult` was a `Pick` that dropped the field, so an `if` condition's own output flattened separately from the body.

New file `redirections.dup-ordering.control-flow.test.ts`: for, C-style for, while, until, if (with condition output), case, break, exit, return, subshell exit, `|&`, and `2>&1 | cat`.

packages/just-bash/src/types.ts
mutewinter: Both done in fa95a5e.

`outputChunks` is now `internalOutputChunks`, matching the other `internal*` fields.

Chunks carry `kind`, so each piece is written by its own shape. The mixed case was real: `{ cat utf8-file; printf 'grün\n' 1>&2; } > f 2>&1` wrote `café` as latin1 `0xE9` because the whole merge took stdout's encoding. It is `0xC3 0xA9` now, matching bash byte for byte, on the file, descriptor, `|&` and caller-stdout paths.

That also turned up `appendResult` tagging decoded text with the producer's shape when the caller overrode stdout, which is what made the merge re-encode.

Chunks are never synthesized from the two strings alone, since `kind` is not recoverable from the text. `build()` also hands out a copy now rather than the live array (VADE bot's nit).

New file `redirections.dup-ordering.encoding.test.ts`.

mutewinter:
Pushed fa95a5e addressing @privatenumber's three comments, plus the two auto-maintain findings. Replies are on the threads; the short version:

**Ordering now survives every layer that relays output.** It was recorded by one accumulator and read by one merge, and everything in between flattened it: `applyRedirections` rebuilding its result, `CompoundOutput` in control flow, `ControlFlowError` on the way out of a scope that exits early, `ConditionResult`'s `Pick`, and `|&`.

**`internalOutputChunks`, with `kind` per chunk.** The two streams reaching one descriptor need not agree on a shape, and merging them under stdout's encoding rewrote one of them. `{ cat utf8-file; printf 'grün\n' 1>&2; } > f 2>&1` wrote `café` as latin1 `0xE9`; it is `0xC3 0xA9` now, matching bash byte for byte.

**Descriptor sharing keys on the open, not on a path match.** auto-maintain was right that the alias list can name an fd a scoped redirection has since replaced. Worse than it flagged: after `exec 3>f; exec 1>&3`, `{ echo O1; echo E1 >&2; echo O2; } >f 2>&3` merged two independent opens and delivered through fd 1's file sink, truncating away the `O1`/`O2` already in the file. It gives `O1 O2 E1` now. Still not bash's `E1 O2`, which wants offset-tracking sinks, but it stops losing output.

Two things worth naming.

An order the upstream never observed has to stay unobserved. My first pass had `applyRedirections` fall back to two coarse chunks, which made a single command's two strings claim to be stdout-then-stderr and broke `pipeline.test.sh [L115] |&: stdout_stderr.py |& cat` in the spec suite, where bash's answer is stderr-first for buffering reasons. Only a genuinely recorded sequence is forwarded now.

fd-table entries are decoded fresh on every lookup, so object identity alone does not survive `exec 3>f` then `>&3 2>&3`. Sharing is recognized three ways: the sink object, the entry an open in this list produced, and the alias group two dups of an inherited descriptor land in.

## Tests

30 cases across three files, 19 of which fail on `af897d7`, checked by copying them onto a worktree at that commit and running them there. The other 11 are regression guards.

The dup matrix in the description is unchanged before and after, verified by running it on both. One correction to it: the `exec 2> f; > f` row reads `O1` where the three-statement body actually gives `O1 O2`.

`test:run` on this branch: 572 files, 15279 passing, 97 skipped, nothing failing, three runs in a row, plus 72 in `just-bash-executor`. Typecheck, Biome, knip and the banned-pattern linter clean.

Two timing-only tests flaked under full-suite load and pass in isolation and on clean runs: `find.perf.test.ts > should measure pure evaluation overhead`, a pure in-memory micro-benchmark that hit vitest's 5s default, and the `loops.test.sh` limit test noted earlier. Neither is reachable from this change.

---

Model: Claude Opus 5

mutewinter:
@privatenumber rebased onto `main` at #436 (28 commits, no conflicts; unit 14,849, comparison 872, spec 3,834, typecheck and biome clean). Your three threads were taken in fa95a5e back in August and the replies are on each; nothing has changed in the approach since. Whenever you have a moment for another look.

mutewinter:
Both findings from the re-run held, and 0ddaeac2 takes them. Neither was introduced by the rebase; the earlier review passes had not tried an fd reopened within one list.

**Reopened descriptor.** Alias overlap was standing in for open identity, and an fd number reused within a list defeats it: `1>&3 3>b 2>&3` snapshots fd 3 twice, once on each open, and both snapshots say 3. The prepare pass now records the entry object it created or copied for every fd it opened or re-pointed and hands that object to every later dup of the fd in the same list, marking the ones that are opens the list made itself. An open the list made is its own identity, so two different objects are two opens; an open that predates the list keeps its alias group as identity, since the table decodes it afresh on every lookup. `exec 3>a; { …; } 1>&3 3>b 2>&3` now writes `O` to `a` and `E` to `b`, and the alias-across-reopen shapes still merge where bash does: `exec 3>a; exec 4>&3; … 1>&3 3>b 2>&4` puts both on `a`, and `… 4>&3 3>b 1>&3 2>&4` puts `E` on `a` and `O` on `b`. Five unit cases and six comparison fixtures recorded against bash 3.2.57 pin those; two of each fail on the previous head.

**Exit in a pipeline stage.** The result rebuilt from an `ExitError` or `ErrexitError` dropped the chunks, so `{ echo O; echo E >&2; exit; } |& cat` printed `E` first. The chunks now ride with the strings they describe; three cases (`exit`, `exit 3`, errexit) fail on the previous head. `2>&1 |` was already right, since the group's own redirection had merged before the error reached the pipeline.

Unit 14,859 pass, comparison 878, spec 3,834; `typecheck`, `lint`, `biome check`, and `knip` clean.

mutewinter:
Three of the five findings on the re-run held, and 0053e77d takes them; one is answered by a boundary change; one I could not reproduce.

**Fd re-pointed at another pre-existing open.** Right, and the previous commit's rule did not reach it: `1>&3 3>&4 2>&3` re-points fd 3 by a dup rather than an open, so neither snapshot is an open the list made and both fall through to the alias group, which lists 3 twice. The prepare pass now counts how many times it has re-pointed each fd, a snapshot records the count beside each fd it lists, and two snapshots overlap only through an fd whose counts agree. `exec 3>a; exec 4>b; … 1>&3 3>&4 2>&3` writes `O` to `a` and `E` to `b`, and `… 1>&3 3>&1 2>&3` still merges, since fd 1 and fd 3 are one open by then. Two unit cases and a bash 3.2.57 fixture pin it; both fail on 0ddaeac2.

**`executeScript` and the errexit statement.** Both sites prepended without chunks, as did the `break`/`continue` and `return` sites beside them; all four now pass `output.chunks` like the other propagation sites. `{ eval '…; exit'; } 2>&1`, an `eval` leaving on `return`, `break`, or `continue`, and `set -e; { … && false; } |& cat` keep their order; six cases and a fixture, all failing on 0ddaeac2. One note on the example as written: `eval '…; exit' 2>&1` with the dup on the `eval` itself does not route the `ExitError` through the command's own redirections at all, on this branch or on `main` (`stderr` still carries `E`), which is a separate gap from ordering and untouched here.

**Retention.** Fair on the public surface: `Bash.exec` now drops `internalOutputChunks` from what it returns, so a caller gets the two streams once and a serialized result no longer repeats them. Inside an execution a compound result does hold its pieces beside the two joined strings, so a scope retains at most twice the output budget rather than once; the budget still bounds it, and the body now says so. Recording offsets instead of pieces would bring it to once, which I would rather do as its own change than fold in here; say the word.

**stderr re-encoding.** I could not make this happen. A mixed byte/text merge landing on live stderr (`{ printf 'grün\n' | cat; echo grün >&2; } 1>&2`) comes out as `grün` twice, on `main` and here, and so does the same through `2>&1 | cat`, `|& cat`, `2> f`, a subshell, a function, and `$(…)`. The boundary decodes stderr (`decodeBinaryToUtf8`) rather than encoding it, so a byte-shaped merged string is decoded once. If you have a shape that double-encodes I will add it as a case.

Unit 14,867 pass, comparison 880, spec 3,834; `typecheck`, `lint`, `biome check`, and `knip` clean. The earlier CI failure on this branch was the two `timeout.test.ts` cancellation cases on the Node 24 runner, which pass locally on repeat and had passed on the same branch thirty minutes earlier; this push reruns them.

Issue #323, opened 2026-07-29
Pipelines drain the shell's shared stdin to EOF — `while read …; do … | …; done < file` runs only once (regression from #285)


trieloff:
## Summary
Any **pipeline** consumes/closes the enclosing shell's stdin, draining it to EOF — even a pipeline where **neither** stage reads stdin (e.g. `true | true`). This breaks the extremely common `while read …; do … | …; done < file` idiom: the loop runs its body once and then exits, because the pipeline has drained the loop's input.

This appears to be a side-effect of the single shared StdinStream model introduced in #285 ("reading is consuming"): the pipeline machinery wires its input to the shared stdin and drains/closes it, regardless of whether any command in the pipeline actually reads.

## Environment
- just-bash `3.2.0` (bash-compat reports `5.1.0(1)-release`)
- Running inside the SLICC browser-native agent runtime

## Repro (minimal)
**A pipeline drains the shell's stdin even outside any loop:**
```bash
printf 'L1\nL2\nL3\n' > /tmp/loop.txt
{ read a; echo "a=[$a]"; true | true; read b; echo "b=[$b]"; } < /tmp/loop.txt
```
Actual:
```
a=[L1]
b=[]        # <-- 'true | true' consumed the rest of stdin
```
Expected (bash):
```
a=[L1]
b=[L2]
```

**The practical fallout — `while read` loop with any pipeline in the body runs once:**
```bash
printf 'L1\nL2\nL3\nL4\nL5\n' > /tmp/loop.txt
n=0; while read a; do n=$((n+1)); echo hi | head -1 >/dev/null; done < /tmp/loop.txt
echo "iterations: $n"     # prints 1; bash prints 5
```

## Scope (what triggers it)
Confirmed with pure builtins (no `node`, no external command) — it is the **pipeline construct itself**:

| Loop body | iters (file has 5 lines) | bash |
|---|---|---|
| *(empty)* | 5 | 5 |
| `{ echo x; }` (command group, no pipe) | 5 | 5 |
| `y=$(cat)` (single cmd, no pipe) | 5 | 5 |
| `true \| true` | **1** | 5 |
| `echo x \| echo y` | **1** | 5 |
| `echo abc \| tr a-z A-Z` | **1** | 5 |
| `echo hi \| head -1` | **1** | 5 |
| `printf 'x\ny\n' \| sed 1q` | **1** | 5 |

[...]

PR #326, opened 2026-07-30
feat(interpreter): support numeric file descriptors (exec N< file, read -u N, done N< file, <&N)

### LINKED ISSUE #321: Numeric input file descriptors (fd ≥3) unsupported: `read -u N`, `done N< file`, `exec N< file` + `<&N` all fail
## Summary
User-defined numeric file descriptors (fd ≥ 3) for **input** are not supported. `read -u N`, compound-command redirects `done N< file`, and `exec N< file` + `<&N` all fail — you cannot open and read from a secondary input fd.

## Environment
- just-bash `3.2.0` (bash-compat reports `5.1.0(1)-release`)
- Running inside the SLICC browser-native agent runtime

## Repro
**1) `read -u FD` silently reads nothing:**
```bash
printf 'one\ntwo\nthree\n' > /tmp/r.txt
n=0; while read -u 3 line; do n=$((n+1)); echo "got: $line"; done 3< /tmp/r.txt
echo "iterations: $n"
```
Actual: `iterations: 0` (no output). Expected: 3 iterations printing one/two/three.

**2) `<&N` → Bad file descriptor:**
```bash
printf 'alpha\nbeta\n' > /tmp/r.txt
exec 3< /tmp/r.txt          # exits 0
read line <&3               # bash: 3: Bad file descriptor
```

**3) Compound-command input redirect to a numeric fd is not honored:**
```bash
cnt=0; while IFS= read -r x <&3; do cnt=$((cnt+1)); done 3< /tmp/r.txt
echo "$cnt"                 # prints 0; bash prints 2
```

## Expected vs Actual
- **Expected (bash):** `N< file` opens `file` on fd `N`; `read -u N` / `read <&N` read from it; `done N< file` keeps fd `N` open for the loop body. This is the canonical way to iterate a file with `while read` while still using stdin inside the loop.
- **Actual:** fd `3` is never actually opened — `exec 3< file` returns 0 but `<&3` reports `Bad file descriptor`, `read -u 3` reads nothing, and `done 3< file` feeds the loop nothing.

## Impact
The idiomatic, robust pattern for looping over a file *while a command inside the loop also consumes stdin* is `while read -u 3 …; done 3< file` (or `read <&3`). With the single-shared-stdin model (#285), a command in the loop body that reads stdin now drains the loop's input; the standard fix is a secondary fd — but secondary input fds don't work, so there is no clean workaround.

## Likely area
Interpreter/redirection: allocate real fd table entries for `N< file` (and `N> file`), honor them in `exec`, `<&N`/`>&N` duplication, `read -u N`, and compound-command redirections.

— filed by Sliccy 🍦 (SLICC agent)

trieloff:
## Summary

User-defined file descriptors (fd >= 3) were not modelled. `exec 3< file` returned 0 without opening anything, `read -u 3` silently read nothing, `done 3< file` fed the loop nothing, and `<&3` reported `Bad file descriptor`. This PR gives the interpreter a real descriptor table and routes every construct that can name a descriptor by number through it.

Closes #321.

#321 frames itself as the missing escape hatch for the single-shared-stdin model discussed in #285 (which was closed unmerged — there is no `StdinStream`; stdin is still `state.groupStdin`). The problem it describes is real under the current model too: a command inside a `while read` body consumes the loop's stdin, and the idiomatic fix is a secondary descriptor (`while read -u 3 …; done 3< file`, or `read <&3`). That pattern now works. This PR deliberately does not touch fd 0 — stdin resolution is unchanged, so it composes with whatever shape the stdin model ends up taking, and does not overlap #323.

### One descriptor table

`src/interpreter/fd-table.ts` owns the encoding of `ctx.state.fileDescriptors`. The map stays a `Map<number, string>` — it is part of the public `CommandContext` surface and is documented as "fd -> content" — with typed entries layered on top:

| kind | opened by |
| --- | --- |
| `input` | `N< file`, `N<<EOF`, `N<<<word` |
| `output` | `N> file`, `N>> file` |
| `readwrite` | `N<> file` |
| `dup-in` / `dup-out` | `N<&0`, `N>&1`, `N>&2` |

The module also exposes `readFd` (returns the unread remainder, or `not-open` / `write-only`), `advanceFd` (moves the shared position), `dupFd`, and `rememberFd` / `restoreFds`.

Scope of the centralisation, precisely: **every fd >= 3 path goes through the table.** Two families still decide for themselves *which* entry to build — `processFdVariableRedirections` (the `{var}>file` allocator) and the fd 0/1/2 duplication branches in `applyRedirections` and `exec`, which construct `dup-in` / `dup-out` entries for the standard streams. Both now hand typed entries to `setFdEntry` / `dupFd`, so no file outside `fd-table.ts` writes a raw marker string any more.

`src/interpreter/numeric-fd-redirects.ts` applies the redirections whose effective descriptor is >= 3 — `N< file`, `N> file`, `N>> file`, `N<> file`, `N<<EOF`, `N<<<word`, `N<&M`, `N>&M`, `N<&M-`, `N>&M-`, `N<&-`, `N>&-` — and hands back a `restore` closure. It is called from exactly one place per construct: `executeSimpleCommandInner` (which also covers `exec`), `withNumericFds` wrapping `if`/`for`/`c-style for`/`while`/`until`/`case`/group/subshell, and `callFunction` for definition-level redirections.

### Duplicated descriptors share one read offset

`N<&M` duplicates the descriptor, not the file: both names refer to one open file description and therefore one read offset. `state.fdAliases` records those groups, and `advanceFd` moves every member together, so `exec 3< f; exec 4<&3; read -u 3; read -u 4` reads the first and then the *second* line. Opening a descriptor (`exec 4< other`) gives it a fresh description and detaches it from the group; closing one member leaves the survivor where the last read left it; and a scoped `4<&3` correctly leaves fd 3 advanced after the body has read through fd 4, because restoring a descriptor restores the descriptor, never the offset behind it. Verified against bash across eleven shapes, including a chain of duplicates, a move (`4<&3-`), a loop-scoped dup, and a scoped dup over an already-open descriptor.

### Content is never mistaken for a marker

Because the table stores strings, a file whose first line reads `__file__:/tmp/x` used to decode as an *output* descriptor. `state.inputFds` records which descriptors hold verbatim content, so those are never re-parsed; the prefix sniffing in `decodeFdEntry` is now only a fallback for values written by code outside the module. This also closes the write side, which was wrong before this PR as well: `exec 3< marker-file; echo BOOM >&3` now reports a bad descriptor instead of writing `BOOM` into the path spelled inside the file. The classification travels with the value through `dupFd`, through scoped save/restore, and through the subshell state copy.

### Semantics matched against real bash

Every behavior below was checked with `pnpm dev:exec --real-bash` before it was encoded in a test:

- **Reading is consuming.** An `input` entry keeps only what has not been read, so `read -u 3` / `read <&3` advance one shared position. `read` advances by exactly what it consumed.
- **`exec` persists, everything else is scoped.** `exec 3< file` stays open until closed. `done 3< file`, `} 3< file`, `fn 3< file` and `cmd 3< file` restore the enclosing descriptor afterwards — including its read position, so `exec 3< f; read -u 3; while … done 3< g; read -u 3` reads `a` then `b`. The early-exit paths for a command word that expands to nothing (`x=''; $x 3< f`) close their descriptors too.
- **Closing.** `N<&-` / `N>&-` close; closing an unopened descriptor is a no-op with status 0, in both `exec` and command position. A command-scoped close is undone after the command (spec case "3>&- << EOF"), but the source of a *move* (`: 6>&7-`) deliberately stays closed — bash does not restore it, and the spec suite pins that ("1>&2- (Bash bug: fail to restore closed fd)").
- **Errors.** `read -u N` on an unopened descriptor → `bash: read: N: invalid file descriptor: Bad file descriptor`, status 1; on a write-only one (including `read -u 1` / `read -u 2`) → `bash: read: read error: N: Bad file descriptor`. `<&N` and `N<&M` on an unopened descriptor → `bash: N: Bad file descriptor`, status 1. `exec N< missing` reports the missing file, returns 1, and does not end the shell.
- **`exec 3<&0`** (the "save stdin" idiom) makes `read -u 3` and `read <&3` read stdin.

### Output-side fixes that fell out of the same table

`applyRedirections` no longer treats a duplication onto a user descriptor or onto stdin as stdout/stderr routing (`<&3` used to emit a spurious `Bad file descriptor`; `3>&1` used to merge stderr into stdout). `>&N` now resolves through the typed table, which fixes two standing bugs: `exec 4>> file; echo x >&4` reported a bad descriptor instead of appending, and `echo x 5> file >&5` did not open fd 5 at all. Writing to a read-side descriptor reports a bad descriptor.

### Deferred / known deviations

- **`<&N` drains the descriptor for any command other than `read`.** `interpreter.ts` advances the descriptor by the whole buffer it handed over, so `true <&3`, `: <&3` and `echo hi <&3` consume it even though they read nothing, and `head -1 <&3` consumes all of it where bash leaves the offset just past what `head` actually read. This is a deliberate approximation: a command's stdin is a string, so there is no signal for how much of it was consumed, and draining matches bash for the common readers (`cat`, `grep`, `wc`). Before this PR the same expressions were wrong in the other direction — the descriptor never advanced at all, plus a spurious `Bad file descriptor`. A precise fix needs consumed-count plumbing through the command interface.
- **Subshells copy the descriptor table rather than sharing the open file description.** In bash, `exec 3< f; ( read -u 3 ); read -u 3` reads `l1` then `l2`; here the parent still reads `l1`. Aliasing is correct *within* a shell — `exec 4<&3` shares an offset, see above — but `beginIsolatedShellState` clones the table and the alias groups so a subshell's reads stay inside it. just-bash already copies stdin the same way, so this stays consistent with the existing model. Not covered by a test, to avoid pinning behavior we intend to change.
- `mapfile` / `readarray` have no `-u` option at all (pre-existing); only `read` takes `-u`.
- Writing to a read-side descriptor reports `bash: N: Bad file descriptor` where bash says `bash: echo: write error: Bad file descriptor` and still emits the text on stdout. The status (1) and the untouched target file match.
- `read line <&3` with fd 3 write-only fails before running `read` (bash runs it and fails inside). Status (1) matches.

## Tests

| File | Tests | Lines |
| --- | --- | --- |
| `packages/just-bash/src/interpreter/fd-table.encoding.test.ts` | 16 | 153 |
| `packages/just-bash/src/interpreter/fd-table.access.test.ts` | 22 | 257 |
| `packages/just-bash/src/interpreter/numeric-fds.input.test.ts` | 21 | 294 |
| `packages/just-bash/src/interpreter/numeric-fds.marker-content.test.ts` | 4 | 77 |
| `packages/just-bash/src/interpreter/numeric-fds.output.test.ts` | 12 | 179 |
| `packages/just-bash/src/interpreter/numeric-fds.scoping.test.ts` | 17 | 273 |
| `packages/just-bash/src/interpreter/numeric-fds.dup-aliasing.test.ts` | 9 | 178 |
| `packages/just-bash/src/comparison-tests/numeric-fds.comparison.test.ts` | 20 | 220 |

121 new tests. `numeric-fds.input.test.ts` opens with the three exact repros from #321. All assertions compare full stdout and stderr strings, and every file is under the repo's 300-line limit.

Comparison fixtures were recorded against real bash — **GNU bash 3.2.57(1)-release**, `/bin/bash` on macOS (`RECORD_FIXTURES=1`). Three of the twenty are marked `"locked": true`: they are the cases whose bash run also writes a diagnostic to stderr, and only the argv0 / `line N:` prefix of those messages differs between bash 3.2 and bash 5.x. The fields the test compares (stdout, exit code) are identical across versions, but an unlocked re-record on the CI runner would rewrite the file and fail the comparison workflow's no-diff check.

Verification on this branch:

- `pnpm typecheck` — clean
- `pnpm lint` (biome + `lint:banned`) — clean
- `pnpm knip` — clean
- `pnpm test:comparison` — **587 passed, 0 failed** (567 before this PR + 20 new). The 3 locked entries pass in playback and survive `RECORD_FIXTURES=1` with no fixture diff.
- spec tests — `src/spec-tests/bash/spec.test.ts` **2336 passed, 0 failed**; the whole `src/spec-tests` directory **3825 passed, 1 skipped, 0 failed**. Identical to `upstream/main`: no fd spec case was previously skipped, so none newly passes — the value here is that all 105 `redirect*` cases still pass through the rewritten path.
- `pnpm test:unit` — **13922 passed, 6 failed, 97 skipped**
- `pnpm test:run --exclude src/spec-tests` — **11153 passed, 214 failed, 98 skipped** (11465 total), against **11041 passed, 220 failed, 98 skipped** (11359 total) on `upstream/main` in the same checkout. (Measured before the descriptor-aliasing commit; the unit, comparison and spec suites above were re-run after it and the failing set is unchanged.)

Every failure on both sides is the CPython-WASM suite: `packages/just-bash/vendor/cpython-emscripten/python.wasm` is an unmaterialized git-LFS pointer (132 bytes) here, so `python: true` tests time out loading the worker. The failing files are identical modulo two that flake between runs (`js-exec.initialization-failure`, `python3.queue-timeout-exploit`), which is why the branch shows six *fewer* failures than main rather than the same number. **No non-WASM test fails on either side**, and the branch adds 106 tests to the total.


trieloff:

Issue #297, opened 2026-07-13
grep: multiple -e patterns are not OR-combined — only the last one is applied


jesse-neurofusion:
**Description**

When `grep` is given multiple `-e` options, POSIX requires all patterns to be combined with OR semantics: a line matches if it matches *any* of the patterns. just-bash currently applies only the last `-e` pattern, silently dropping matches for the others.

**Reproduction** (justbash.dev, also reproduced with `just-bash@3.1.0` on Node 22 / Bun 1.3.5)

```bash
printf 'aaa Q1\nbbb\nccc Q4\n' > t.txt
grep -e 'Q1' -e 'Q4' t.txt
```

**Expected** (GNU grep / POSIX)

```
aaa Q1
ccc Q4
```

**Actual**

```
ccc Q4
```

Exit code is `0` and stderr is empty in both cases, so there is no signal that matches were dropped.

**Why this matters for agent use cases**

Since the command "succeeds" with plausible output, an LLM agent has no way to notice the missing matches — it will typically conclude the data simply doesn't contain them and carry on with a wrong analysis. A hard error would be recoverable (the model retries); a silent partial result is not.

**Reference:** POSIX grep spec, `-e pattern_list`: "Specify one or more patterns to be used during the search for input. […] multiple `-e` and `-f` options are accepted and grep uses all of the patterns it is given while matching input text lines." — the fix is presumably to collect all `-e` values and join them as alternation, rather than keeping the last.

**Workaround:** `grep -E 'Q1|Q4'` behaves correctly.

Thanks for the project — the agent-oriented design (in-memory fs, network off by default, re2js) is exactly what we needed. Happy to provide more test cases if helpful.

PR #184, opened 2026-04-09
Add grep -f flag to read patterns from file

ericjypark:
`grep -F -f a.txt b.txt` fails with "invalid option -- 'f'":

```
$ echo "hello" > a.txt
$ echo "hello2" > b.txt
$ grep -F -f a.txt b.txt
grep: invalid option -- 'f'
```

This adds the `-f FILE` flag, which reads patterns from a file
(one per line) and treats them as alternation.

Fixes #166

PR #327, opened 2026-07-30
feat(grep): support -f/--file to read patterns from file

### LINKED ISSUE #322: `grep -f PATTERNFILE` (patterns from file) unsupported: `invalid option -- 'f'`
## Summary
`grep -f PATTERNFILE` (read patterns from a file, one per line) is not supported — the flag is rejected.

## Environment
- just-bash `3.2.0` (bash-compat reports `5.1.0(1)-release`)
- Running inside the SLICC browser-native agent runtime

## Repro
```bash
printf 'apple\nbanana\n' > /tmp/pat.txt
printf 'apple pie\ncherry\nbanana split\n' > /tmp/hay.txt
grep -f /tmp/pat.txt /tmp/hay.txt
```
```
grep: invalid option -- 'f'
```

## Expected vs Actual
- **Expected:** patterns are read from `/tmp/pat.txt` (one per line); output:
  ```
  apple pie
  banana split
  ```
- **Actual:** `grep: invalid option -- 'f'`, exit 1.

## Impact
`grep -f` is the standard way to filter one list against another set of patterns/fixed strings (especially with `-F -x -f`, `-v -f`). Without it, set-difference/intersection over lines requires ad-hoc `comm`/loops — and `comm` needs sorted input plus process substitution (`<(…)`), which is also currently unavailable (see companion issue), compounding the gap.

Related but distinct existing grep issues: #297 (`-e` OR-combination), #166. This is specifically the missing `-f` (patterns-from-file) option.

## Likely area
`grep` command implementation: accept `-f FILE` (and `--file=FILE`), reading newline-separated patterns; compose with `-F`, `-x`, `-v`, `-i`.

— filed by Sliccy 🍦 (SLICC agent)

trieloff:
## Summary

Adds GNU-compatible `grep -f FILE` / `--file=FILE` — patterns read from a file, one per line — so `grep -F -x -f`, `grep -v -f` and friends work as set operations over lines.

Closes #322.

This **builds on #184** by @ericjypark, which is still open and still applies cleanly to `main` — the problem is correctness, not staleness. Behaviours where #184 diverges from GNU grep 3.12:

1. **`-x` produces wrong results.** #184 joins the alternatives into `(?:a)|(?:b)` and hands that to `buildRegex`, whose `lineRegexp` wrapper produced `^(?:a)|(?:b)$` — "starts with a" OR "ends with b", not "the whole line is a or b". `-F -x -f` is the primary use case in #322, so this matters.
2. **`-fFILE` and `-vxfFILE` read the wrong file.** The cluster loop's `f` case always takes `args[++i]`, so `grep -fpat.txt hay.txt` reads patterns from `hay.txt` and is then left with no input file. (`-f FILE` and `-vxf FILE` do work.)
3. **Only one `-f` is kept** (`patternFile` is a single slot, so `-f a -f b` silently drops `a`), and `-f` does not combine with `-e` — whichever is written last wins.
4. **Empty pattern files are wrong in both directions.** #184 returns `exit 1` immediately whenever the file is empty, so `grep -v -f EMPTY` prints nothing where GNU prints every line, and `grep -c -f EMPTY` is only accidentally right.
5. **A malformed pattern can absorb its neighbour.** Joining alternatives textually means a pattern file of `a\` + `banana` compiles as the single literal `a|banana` instead of erroring; under `-v` that turns a blocklist into a pass-everything filter.
6. **`-f -`** (patterns from stdin) is unsupported, and would double-read stdin if it were.
7. **`-P` accepts multiple patterns**, which GNU refuses outright.

### What changed vs #184

- **Fixed `-x`**: `buildRegex` now wraps as `^(?:PATTERN)$`, mirroring what the `-w` wrapper already does. Split into its own commit.
- **All `-f` spellings**: `-f FILE`, `-fFILE`, `-vxf FILE`, `-vxfFILE`, `--file=FILE`, `--file FILE`, and `-f -`.
- **Repeated `-f` unions**, and unions with `-e` in either order.
- **Empty pattern list handled like GNU** rather than as an early error (details below).
- **Each pattern is compiled standalone before joining** (basic/extended only — `-F` patterns are escaped and can never be malformed), so a syntax error is reported instead of silently swallowing the next pattern.
- **`-P` with more than one distinct pattern is refused**: `grep: the -P option only supports a single pattern`, exit 2.
- **`-f -` models stream consumption**: only the first `-f -` reads stdin; a repeated `-f -` sees EOF and contributes nothing, so `printf 'a\n' | grep -P -f - -f - FILE` stays a single-pattern search instead of duplicating it.
- **Pattern-file errors are GNU's**: `grep: FILE: No such file or directory` / `Is a directory`, exit 2, raised before any input file is searched — #184 has no directory case and always reports "No such file or directory".
- BRE alternatives are joined with `\|` instead of being lifted to ERE, so a BRE pattern file stays a BRE (#184's `-f` silently switched the whole pattern to extended regex, changing what `+`, `?`, `{n}` and `(` mean).
- **Non-option operands are files when `-f` is present** — `grep -f pat.txt hay.txt` must not treat `hay.txt` as the pattern.
- Pattern count from files is bounded by `maxArrayElements`, per the repo's runaway-compute rule.
- Tests rewritten and expanded (see below); every expectation re-derived from GNU grep 3.12 rather than assumed.

Credit to @ericjypark for the original implementation and tests (#184). The `-e`/`-i`/`--file=`/blank-line/empty-file test scenarios there are all preserved here, with the expected values corrected where they diverged from GNU.

## Details

**GNU edge cases matched** (each verified against GNU grep 3.12, `ggrep` from Homebrew):

| case | GNU behaviour |
| --- | --- |
| blank line in pattern file | empty pattern, matches every line |
| blank line + `-x` | anchors to `^$`, matches only blank lines |
| trailing newline in pattern file | terminator, does **not** add an empty pattern |
| trailing newline in a `-e` operand | separator, **does** add an empty pattern |
| empty pattern file / `-f /dev/null` | selects nothing, exit 1 |
| `-c -f EMPTY` | prints **nothing** (not `0`), exit 1 |
| `-f EMPTY missing_file` | no diagnostic at all, exit 1 — GNU never opens the inputs |
| `-v -f EMPTY` | selects every line, exit 0 |
| `-c -v -f EMPTY` | per-file counts as usual |
| missing pattern file | `grep: FILE: No such file or directory`, exit 2 |
| directory pattern file | `grep: FILE: Is a directory`, exit 2 |
| `-f a -f missing` | reports only `missing`, exit 2, nothing searched |
| `-f` with no argument | `grep: option requires an argument -- 'f'`, exit 2 |
| `--file` with no argument | `grep: option '--file' requires an argument`, exit 2 |
| `-f ""` / `--file=` | `grep: : No such file or directory`, exit 2 (never resolved against the cwd) |
| `-f -` | patterns from stdin; stdin is then drained, so `printf x \| grep -f -` exits 1 |
| repeated `-f -` | only the first reads stdin; later ones hit EOF and add no patterns |
| malformed pattern among several | rejected, exit 2 — never joined into the alternation |
| `-P` with more than one *distinct* pattern | `grep: the -P option only supports a single pattern`, exit 2 (duplicates are folded first, so `-P -f dup.pat` is accepted) |

The empty-pattern-list short-circuit is the subtle one: GNU exits 1 without visiting the inputs when the list is empty and `-v` is absent, but `-v` and `-L` still need the normal path. Implemented as an early return for the former and an unmatchable regex (`[^\s\S]`) for the latter, so `-x`/`-w` wrapping keeps it unmatchable.

**Malformed patterns.** Alternatives are concatenated textually, so a pattern that is invalid on its own can otherwise swallow the separator and merge with its neighbour — `a\` + `banana` becoming the literal `a|banana`, `[a` + `b]c` matching `ac`. Each pattern is therefore compiled standalone first (skipped for `-F`, whose patterns are escaped) and a failure is reported as `grep: invalid regular expression: PATTERN`, exit 2. GNU words these differently (`grep: FILE:1: Trailing backslash`); this keeps just-bash's existing phrasing. Residual gap: `a\{1` still compiles here, because incomplete BRE braces are treated as literals — a pre-existing engine difference already recorded in the grep spec-test skip patterns, so `a\{1` + `banana` matches where GNU errors.

**Composition with `-e` / relationship to #314.** GNU treats `PATTERNS` (positional, `-e`, or `-f`) as a newline-separated list, and unions everything. This PR builds that union and adds a `combinePatterns()` helper that OR-combines while preserving each mode's semantics (`\|` for BRE, `(?:...)|(?:...)` for ERE, escape-then-lift for `-F`). It deliberately **does not** change `-e`'s "last one wins" behaviour — that is #297, owned by #314.

The two PRs are complementary in intent but **do overlap in code**: #314 adds a near-identical `combinePatterns()` at the same insertion point, makes the same `^(?:...)$` anchor fix, and edits the same spec-test skip list. `git merge-tree` reports conflicts in all three files (`grep.ts`, `search-engine/regex.ts`, `spec-tests/grep/skips.ts`) — roughly 70% of the non-test diff is shared. Whichever lands second should rebase; the resolution is mechanical (keep one `combinePatterns`, keep one anchor fix, take the union of the skip-list removals) since the two implementations agree on semantics.

Newline-separated `PATTERNS` operands are split here as well, because `-f` is defined in terms of that same rule and the common `grep -F "$(cat file)"` workaround depends on it.

**Spec tests unskipped** (5, all now passing):

- `grep can read regexps from stdin` (`-f -`)
- `grep -v -f EMPTY_FILE`
- `grep -vxf EMPTY_FILE`
- `grep PATTERN can be a newline-delimited list`
- `grep -e PATTERN can be a newline-delimited list`

## Tests

New files:

- `packages/just-bash/src/commands/grep/grep.patterns-from-file.test.ts` — 28 tests: reading forms, `-f`+`-e` union, repeated `-f`, no trailing newline, `-f -`, stdin drain, repeated `-f -` (plain and under `-P`), blank/mixed/empty pattern files, `-f /dev/null`, `-c`/`-v` on an empty list, the four error cases, and the exact repro from #322.
- `packages/just-bash/src/commands/grep/grep.patterns-from-file-combos.test.ts` — 21 tests: `-F -x -f`, `-F -x -v -f`, `-F -x -c -f`, `-x` alternative anchoring, `-v -f`, `-i`, `-w`, `-F` literals, BRE/ERE modes, `-o`, `-c`, `-l`, `-h`, `-m`, `-A`, `-r`, `-q`.
- `packages/just-bash/src/commands/grep/grep.patterns-from-file-validation.test.ts` — 16 tests: malformed patterns (trailing backslash, unmatched `[`, unmatched `(` under `-E`, and the same backslash accepted under `-F`), the `-P` single-pattern restriction and what it still allows (single pattern, single empty pattern, duplicates, empty file), and the four argument-shape cases (`-f ""`, `--file=`, `--file` and `-f` with no argument).
- `packages/just-bash/src/comparison-tests/grep-patterns-from-file.comparison.test.ts` + `fixtures/grep-patterns-from-file.comparison.fixtures.json` — 26 fixtures.

All assertions are full-string `stdout`/`stderr` plus exit code.

**Fixture recording.** macOS ships BSD grep 2.6.0-FreeBSD, so the fixtures were recorded against **GNU grep 3.12** (Homebrew `grep` formula) by putting its `gnubin` first on `PATH`, and every entry is marked `"locked": true` so a re-record on a BSD box can't silently replace GNU behaviour with BSD behaviour. The re-record command is documented at the top of the test file:

```
PATH=/opt/homebrew/opt/grep/libexec/gnubin:$PATH \
  RECORD_FIXTURES=force pnpm test:run \
  src/comparison-tests/grep-patterns-from-file.comparison.test.ts
```

**Verification results:**

- `pnpm typecheck` — clean
- `pnpm lint:fix` — clean (no fixes applied)
- `pnpm lint:banned` — clean
- `pnpm knip` — clean (only the two pre-existing configuration hints)
- `pnpm test:comparison` — 593 passed, 36 files (26 of them new)
- `pnpm test:run src/commands/grep src/commands/rg src/commands/search-engine src/spec-tests/grep` — 1195 passed, 90 skipped, 35 files
- `pnpm test:run` (full) — 14378 passed, 97 skipped; the 6 failures are pre-existing on `main` in this environment (python3/CPython-WASM and bundled-binary tests that need `pnpm build` and the vendored WASM), confirmed by re-running them on a clean tree

**Known gap (pre-existing, not touched here):** `grep -L`'s exit code is inverted relative to GNU — GNU exits 1 when `-L` prints names and 0 when it prints none, since the status reflects whether a line was selected. `-L` is already listed as unimplemented in the spec-test skips, so no `-L` assertion is included.

### DIFF
--- a/.changeset/witty-moons-shave.md
+++ b/.changeset/witty-moons-shave.md
@@ -1,0 +1,5 @@
+---
+"just-bash": minor
+---
+
+Add `grep -f FILE` / `--file=FILE` to read patterns from a file (one per line). Patterns from `-f` OR-combine with `-e` patterns and with each other, `-f -` reads patterns from stdin, empty pattern lines match every line, and an empty pattern file selects nothing (exit 1). Newline-separated `PATTERNS` operands are now split into individual patterns, and `-x` groups alternatives correctly (`^(?:a|b)$`).
--- a/packages/just-bash/src/commands/grep/grep.patterns-from-file-combos.test.ts
+++ b/packages/just-bash/src/commands/grep/grep.patterns-from-file-combos.test.ts
@@ -1,0 +1,221 @@
+import { describe, expect, it } from "vitest";
+import { Bash } from "../../Bash.js";
+
+/**
+ * `grep -f FILE` composed with the other selection flags. The `-F -x -f` and
+ * `-v -f` combinations are the set-intersection / set-difference idioms called
+ * out in issue #322. Expectations verified against GNU grep 3.12.
+ */
+const setFiles = {
+  "/list.txt": "apple\nbanana\ncherry\n",
+  "/keep.txt": "banana\ndate\napple\n",
+};
+
+const searchFiles = {
+  "/pat.txt": "apple\nbanana\n",
+  "/p1.txt": "apple\n",
+  "/hay.txt": "apple pie\ncherry\nbanana split\n",
+  "/hay2.txt": "x\ny\n",
+};
+
+describe("grep -f set operations", () => {
+  it("intersects two line sets with -F -x -f", async () => {
+    const env = new Bash({ files: setFiles });
+    const result = await env.exec("grep -F -x -f /keep.txt /list.txt");
+    expect(result.stdout).toBe("apple\nbanana\n");
+    expect(result.stderr).toBe("");
+    expect(result.exitCode).toBe(0);
+  });
+
+  it("subtracts one line set from another with -F -x -v -f", async () => {
+    const env = new Bash({ files: setFiles });
+    const result = await env.exec("grep -F -x -v -f /keep.txt /list.txt");
+    expect(result.stdout).toBe("cherry\n");
+    expect(result.stderr).toBe("");
+    expect(result.exitCode).toBe(0);
+  });
+
+  it("counts the intersection with -F -x -c -f", async () => {
+    const env = new Bash({ files: setFiles });
+    const result = await env.exec("grep -F -x -c -f /keep.txt /list.txt");
+    expect(result.stdout).toBe("2\n");
+    expect(result.stderr).toBe("");
+    expect(result.exitCode).toBe(0);
+  });
+
+  it("anchors every alternative with -x, not just the first", async () => {
+    const env = new Bash({
+      files: {
+        "/pat.txt": "foo\nbar\n",
+        "/data.txt": "foo\nbar\nfoobar\nfoo suffix\nprefix bar\n",
+      },
+    });
+    const result = await env.exec("grep -x -f /pat.txt /data.txt");
+    expect(result.stdout).toBe("foo\nbar\n");
+    expect(result.stderr).toBe("");
+    expect(result.exitCode).toBe(0);
+  });
+
+  it("inverts the union of all file patterns with -v", async () => {
+    const env = new Bash({ files: searchFiles });
+    const result = await env.exec("grep -v -f /pat.txt /hay.txt");
+    expect(result.stdout).toBe("cherry\n");
+    expect(result.stderr).toBe("");
+    expect(result.exitCode).toBe(0);
+  });
+});
+
+describe("grep -f with matching modifiers", () => {
+  it("applies -i to every pattern", async () => {
+    const env = new Bash({
+      files: { "/p1.txt": "apple\n", "/caps.txt": "APPLE PIE\nApple\nnope\n" },
+    });
+    const result = await env.exec("grep -i -f /p1.txt /caps.txt");
+    expect(result.stdout).toBe("APPLE PIE\nApple\n");
+    expect(result.stderr).toBe("");
+    expect(result.exitCode).toBe(0);
+  });
+
+  it("applies -w to every pattern", async () => {
+    const env = new Bash({
+      files: {
+        "/p1.txt": "apple\n",
+        "/words.txt": "apple\npineapple\napple pie\n",
+      },
+    });
+    const result = await env.exec("grep -w -f /p1.txt /words.txt");
+    expect(result.stdout).toBe("apple\napple pie\n");
+    expect(result.stderr).toBe("");
+    expect(result.exitCode).toBe(0);
+  });
+
+  it("keeps every -F pattern literal", async () => {
+    const env = new Bash({
+      files: {
+        "/meta.txt": "a.b\nx*y\n",
+        "/metahay.txt": "a.b\naxb\nx*y\nxy\n",
+      },
+    });
+    const result = await env.exec("grep -F -f /meta.txt /metahay.txt");
+    expect(result.stdout).toBe("a.b\nx*y\n");
+    expect(result.stderr).toBe("");
+    expect(result.exitCode).toBe(0);
+  });
+
+  it("treats every pattern as a BRE by default", async () => {
+    const env = new Bash({
+      files: {
+        "/meta.txt": "a.b\nx*y\n",
+        "/metahay.txt": "a.b\naxb\nx*y\nxy\n",
+      },
+    });
+    const result = await env.exec("grep -f /meta.txt /metahay.txt");
+    expect(result.stdout).toBe("a.b\naxb\nx*y\nxy\n");
+    expect(result.stderr).toBe("");
+    expect(result.exitCode).toBe(0);
+  });
+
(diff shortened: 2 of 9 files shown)


trieloff:
I'll take care of the rebases, depending on which order you prefer for the #314 overlap.

cramforce (approved):

PR #329, opened 2026-07-30
fix(interpreter): apply output redirections on while/until loops

trieloff:
## Summary

Output redirections attached to a `while` or `until` loop were silently dropped:

```bash
while true; do echo x; break; done >/dev/null; echo end
# just-bash: "x\nend"
# bash:      "end"
```

Everything the loop wrote went straight to the caller — `> file`, `>> file`, `2> file`, `2>&1`, `&>` and `>|` on a loop all had no effect. `for`, C-style `for` and `case` were unaffected; only `while`/`until` were missing the wiring.

Found while verifying #328 (pipeline stdin). No issue was filed for it.

### Root cause

`executeFor`, `executeCStyleFor` and `executeCase` bracket their body with the compound-command redirection pair:

* `preOpenOutputRedirects(ctx, node.redirections)` before the body — expands the targets (so `> "f$((n++)).txt"` bumps `n` first) and truncates `>` targets before anything in the loop can read them.
* `applyRedirections(ctx, bodyResult, node.redirections, targets)` after the body — delivers the collected stdout/stderr to the fds' final sinks.

`executeWhile` did neither. It walked `node.redirections` only to pull *input* redirections (`< file`, here-docs, here-strings) into `ctx.state.groupStdin` — which is why `while read line; do ...; done < file` has always worked — and then returned `output.build(exitCode)` directly, so every output redirection on the loop was ignored. `executeUntil` did not look at `node.redirections` at all, so both its input and its output redirections were dropped.

### The fix

A new `prepareLoopRedirections()` helper in `control-flow.ts` runs the loop's redirection prologue once, and both loops now use it:

1. Input redirections (`<`, `<<`, `<<-`, `<<<`) are consumed to seed the loop's stdin — this is the existing `while` logic, moved.
2. Every remaining redirection is handed to `preOpenOutputRedirects` before the loop and to `applyRedirections` after it, exactly like `for`/`case`.

Two deliberate details:

* **Input redirections are kept out of the pre-open/apply list.** They are already fully consumed in step 1; routing them through `preOpenOutputRedirects` would expand their targets a second time and would apply `expandRedirectTarget`'s globbing to `<<<`, which bash does not do (`done <<< *` must yield a literal `*`). There is a comparison test pinning this.
* **Steps 1 and 2 interleave strictly left to right** (see below), so a failed redirection leaves exactly the redirections to its left applied.

`executeUntil` also gained the `stdin` parameter `executeWhile` already had (dispatch in `interpreter.ts`), so `printf 'a\nb\n' | until ! read l; do ...; done` reads from the pipe. That is the same loop-level redirection code path, and without it `until ... done < in > out` would have been only half fixed.

### Strict left-to-right redirection order

Thanks to the **Vercel review bot**, which flagged on `control-flow.ts:134` that a failing `< file` aborted before any output redirect was pre-opened — so an output redirect standing *to the left* of the failing input was never applied. Bash processes the list in order, so it truncates that target and only then fails.

`prepareLoopRedirections` now walks the list once, in order. Output redirections accumulate into a pending run that is pre-opened the moment an input redirection (or the end of the list) is reached; an input redirection is resolved where it stands, and on failure the diagnostic is routed through exactly the redirections opened so far. Because `ExpandedRedirectTargets` is keyed by position in the array handed to `preOpenOutputRedirects`, each run's keys are rebased onto the accumulated `opened` array before merging, so `opened`/`targets` stay index-coherent for `applyRedirections` — that is what keeps `done > "f$((n++))" < nosuch` expanding its target exactly once (verified: `n=1`, `f0` created, `f1` not).

Measured on `/bin/bash` 3.2.57 **and** `/opt/homebrew/bin/bash` 5.3.15 — byte-identical on both:

| script | bash | before | after |
| --- | --- | --- | --- |
| `done > o < nosuch` (o pre-exists) | `o` truncated, rc=1 | `o` unchanged | matches |
| `done > o < nosuch` on `until` | `o` truncated, rc=1 | `o` unchanged | matches |
| `done >> o < nosuch` (o pre-exists) | `o` kept, rc=1 | `o` kept | matches |
| `done >> o < nosuch` (o missing) | `o` created empty, rc=1 | `o` not created | matches |
| `done > a < nosuch > b` (both exist) | `a` truncated, `b` kept | both kept | matches |
| `done > a < nosuch > b` (neither exists) | `a` created, `b` not | neither created | matches |
| `done 2> e < nosuch` | diagnostic lands **in `e`** | `e` untouched, message to caller | matches |
| `done > "s$((n++))" < nosuch` | `n=1`, `s0` created | `n=0`, nothing created | matches |
| `done > f < f` | `f` truncated first, loop reads nothing, rc=0 | loop echoed the old contents back | matches |
| `done > o <<EOF < nosuch` | `o` truncated, rc=1 | `o` unchanged | matches |
| `done < nosuch > o` | `o` untouched, rc=1 | `o` untouched | unchanged |
| `done > a < in > b` (success) | `a` empty, `b` gets the output | already correct | unchanged |

### Stdin ownership (`resolveLoopStdin`)

Threading `until` through the same prologue exposed a trap in the pre-existing stdin handling: the loops installed `ctx.state.groupStdin` and then restored it *unconditionally* in `finally`, even when they had never installed anything. Restoring a stream the loop merely INHERITED rewinds the shared read position, so the next `read` re-reads a consumed line.

`resolveLoopStdin()` makes ownership explicit. A loop installs — and therefore restores — stdin only when it brought its own: its own input redirection, or the stdin handed to it as a pipeline stage. An inherited stream is left alone so reads inside the body advance it naturally. `ownStdin: ""` still counts as ownership, so `done < empty-file` gives the body an empty stream rather than the enclosing one.

The guard lives in the shared helper, so **`while` gets it too**. That fixes four `while` behaviours that were already wrong on `main` (verified against bash, now covered by tests):

| case | main | branch / bash |
| --- | --- | --- |
| `printf 'a\nb\nc\n' \| { while read l; do break; done; read r; }` | `r=a` | `r=b` |
| two sequential piped `while` reads | `x=a y=a` | `x=a y=b` |
| piped `while` drained to EOF, then `read r` | `r=a`, code 0 | `r=`, code 1 |
| nested `while` sharing one stream (4 lines) | `a-b b-c c-d` | `a-b c-d` |
| `while ... done < empty.txt` inside a group | ate the enclosing stream | leaves it untouched |

Nothing regressed from the guard: `{ while ...; done \| cat; read r; }` and `while` inside a function still behave exactly as on `main` (both are clobbered by the pipeline/function scope layers, not by the loop).

### Bash matrix verified

Every case below was run through `pnpm dev:exec --real-bash` against /bin/bash 3.2.57 before any expectation was written. All now match bash byte-for-byte on stdout:

| case | before | after |
| --- | --- | --- |
| `>/dev/null`, `> file`, `>> file` (while + until) | ignored | matches bash |
| `2> file`, `> out 2>&1`, `2>&1 > out`, `&>` | ignored | matches bash |
| `>\|` and noclobber refusal | ignored | matches bash |
| `> a > b` (two stdout targets) | ignored | matches bash |
| `> dir` → `Is a directory`, exit 1 | body ran, exit 0 | matches bash |
| target truncated before body reads it (`done > f` with `cat f` inside) | stale content | matches bash |
| target side effects (`> "f$((n++)).txt"`) | not expanded | matches bash |
| `\| cat` after the loop, `2>&1 \| cat` | already worked / stderr leaked | matches bash |
| `done < file` (classic read loop) | worked | still works |
| left-to-right ordering of a failing redirection (see table above) | earlier output redirects skipped | matches bash |
| `done < in > out` (while + until) | out empty | matches bash |
| here-doc / here-string + `> out` | out empty | matches bash |
| `done <<< *` (no globbing of here-strings) | literal `*` | still literal `*` |
| `done < nosuch > out` (out not truncated) | out kept | out kept |
| `until ... done < in`, `until` here-doc/here-string, piped `until` | produced nothing | matches bash |
| nested loops, inner and outer redirects | ignored | matches bash |
| `break` / `continue` inside a redirected loop | ignored | matches bash |
| `break 2` / `continue 2` where the redirect is on the loop they TARGET | ignored | matches bash |
| inherited-stream read positions (see table above) | wrong on `while`, would have been wrong on `until` | matches bash |
| `done <&3` (bad fd) | ran the body, exit 0 | `Bad file descriptor`, exit 1 (same as `for`) |

Known deviations left in place. All are pre-existing and shared with `for`, so they are out of scope here:

* `exit`, `return` and errexit unwinding out of a redirected loop still deliver the loop's output to the caller instead of the file (`for` behaves identically — the unwinding error carries the output past `applyRedirections`).
* Same class: `break 2` / `continue 2` out of an INNER loop that carries its own redirect leaks that inner loop's output. `for` does the same (`for a in 1; do for b in 1; do echo x; break 2; done > f; done` leaks `x` on both). Only the redirect on the loop the jump targets is fixed here; that is the case the tests cover.
* `for` and `case` still process their redirection list as "expand and validate everything, then truncate everything", so they keep the ordering divergence that `while`/`until` no longer have: `for i in 1; do echo x; done > out < nosuch` leaves `out` untouched where bash truncates it. Fixing that means reworking `executeFor`/`executeCStyleFor`/`executeCase` and is outside this PR's scope. The same all-then-truncate rule still applies *within* a single run of consecutive output redirections on any compound command (`> a > adir` validates both before truncating `a`) — unchanged here.
* `for ... done < file` still does not feed the body; only `while`/`until` consume loop-level input redirections.
* `done < f > f` echoes the file's old contents back into `f`, where bash ends up with an empty `f`. Our `<` snapshots the file's bytes when the redirection is set up, so a later `> f` cannot affect what the body reads; bash hands the loop a descriptor and reads lazily, after the truncation. The reverse order, `done > f < f`, does now match bash. Unchanged by this PR.
* `done < somedir` reports `bash: somedir: No such file or directory` and exit 1, where bash lets `read` fail with a read error and the loop exits 0. `while` has always behaved this way; `until` previously matched bash only by accident (it ignored the redirection entirely) and now shares `while`'s behaviour. Fixing it means changing how loop input redirection reports a directory, which is `while`'s long-standing path.
* `break` after a failing command reports the failing command's status rather than 0.
* `done < "nosuch$((n++))"` expands its target once (bash expands it twice, once for the failed open and once for the diagnostic, so bash reports `nosuch1` and leaves `n=2`). The branch is the more principled of the two; the divergence is obscure and untested either way.
* `done 3< in` is consumed as the loop's stdin rather than being bound to fd 3. That is a pre-existing wart in `while`'s input handling which `until` now inherits; numeric fds are #326's area.

### Coordination with open PR #326

This branch conflicts semantically (not textually in every hunk) with **#326 (numeric fds)**, which rewrites the same loop functions in `control-flow.ts`:

* #326 wraps the loop bodies in `withNumericFds` and adds `isNumericFdRedirection` skips inside the very input-redirection loop that this PR moved into `prepareLoopRedirections`.
* #326 threads a third `fdTargets` argument through `preOpenOutputRedirects`.
* #326's `executeUntilBody` has no `stdin` parameter, while this PR adds one to `executeUntil`.

Whoever lands second must: re-apply the `isNumericFdRedirection` skip inside `prepareLoopRedirections` (otherwise `done 3< in` keeps being swallowed as loop stdin), thread `fdTargets` through the `preOpenOutputRedirects` call in the helper, and reconcile the `executeUntil`/`executeUntilBody` signature so the pipeline stdin still reaches the loop.

## Tests

* `packages/just-bash/src/interpreter/control-flow.loop-redirections.test.ts` (221 lines, 17 tests) — the output/input redirect matrix, full-string stdout/stderr assertions, including the exact repro (`while true; do echo x; break; done >/dev/null; echo end` → `end\n`).
* `packages/just-bash/src/interpreter/control-flow.loop-redirections-flow.test.ts` (95 lines, 6 tests) — `break 2`/`continue 2` on the targeted loop, plus piped loops.
* `packages/just-bash/src/interpreter/control-flow.loop-stdin-ownership.test.ts` (159 lines, 13 tests) — the ownership rule: inherited streams keep advancing (the three regression cases above, for both `while` and `until`), own streams are installed and restored.
* `packages/just-bash/src/interpreter/control-flow.loop-redirection-order.test.ts` (211 lines, 15 tests) — the left-to-right ordering table above, both directions, plus the success paths.
* `packages/just-bash/src/comparison-tests/loop-redirections.comparison.test.ts` (18 tests — the output-redirection matrix), `loop-redirections-stdin.comparison.test.ts` (15 tests — input redirection, nesting, pipelines) and `loop-redirection-order.comparison.test.ts` (10 tests — the ordering cases), plus their three fixture files. Split to stay under the 300-line house limit. Fixtures recorded with `RECORD_FIXTURES=1` against **GNU bash 3.2.57(1)-release (arm64-apple-darwin25)**.

Every expectation in all seven files was checked against real bash before being written, and replays byte-identically under `/bin/bash` 3.2.57.

**Locked fixtures (13 of 43).** The runner compares only stdout and exit code, but it still *records* stderr into the fixture JSON, and the comparison-tests workflow fails on any git diff after re-recording. Every locked entry is one where a redirection FAILS and bash emits a diagnostic carrying its shell path and `line N:` script position:

| fixtures | case |
| --- | --- |
| `240730055fc830ec` | noclobber refusal (`cannot overwrite existing file`) |
| `eb83615d48785472` | `> adir` (`Is a directory`) |
| `98da27b468bc4899`, `ad1c8cf5e7b450d8` | `done < nosuch.txt` (`No such file or directory`) |
| 9 entries in `loop-redirection-order.comparison.fixtures.json` | the failing-redirection ordering cases |

This was **pre-validated against bash 5.x** rather than guessed: pointing `fixture-runner.ts`'s hardcoded `shell: "/bin/bash"` at `/opt/homebrew/bin/bash` (5.3.15) and re-recording all 43 loop fixtures with `RECORD_FIXTURES=force` changed **stderr on exactly those 13 entries** (`/bin/bash: nosuch:` → `/opt/homebrew/bin/bash: line 1: nosuch:`, `line 1:` → `line 2:`) and left **stdout and exit code byte-identical on all 43** — i.e. just-bash matches both bash versions everywhere, and only the unstable strings needed locking. The runner change was reverted and the bash-3.2 recordings restored. The 30 unlocked entries re-record cleanly, confirmed by a final `RECORD_FIXTURES=1` pass producing no fixture diff.

The one ordering fixture that is deliberately *not* locked is the `2> e.txt < nosuch` case: its diagnostic goes into the redirect target instead of stderr, so its recorded stderr is empty and stable. Its stdout asserts `grep -c 'No such file or directory' e.txt` rather than the message text, so it survives version changes.

* `.changeset/loop-output-redirections.md` — patch bump.

### Verification numbers

Baseline measured on a clean `upstream/main` in the same worktree (`git stash`), with `dist/` rebuilt via `pnpm build` first so the bundle tests behave the same in both runs.

| suite | baseline (upstream/main) | this branch |
| --- | --- | --- |
| `pnpm test:run` (just-bash) | 5 files failed / 518 passed (523); **6 tests failed** / 14287 passed / 97 skipped (14390) | 5 files failed / 525 passed (530); **6 tests failed** / 14381 passed / 97 skipped (14484) |
| `pnpm test:run` (just-bash-executor) | 4 files / 72 passed | 4 files / 72 passed |
| spec tests (`src/spec-tests`) | 3825 passed / 1 skipped | 3825 passed / 1 skipped |
| `pnpm test:comparison` | 35 files / 567 tests | 38 files / 610 tests, all passing |

**Zero new failures.** The 6 failures are identical before and after — all CPython-WASM tests failing on pre-existing LFS-pointer artifacts (`just-bash.bundle.test.ts` python3 lazy-load, `code-exec-exploit-regression`, `defense-in-depth-independence`, `python-sqlite-information-disclosure` ×2, `worker-protocol-runtime-desync`).

Two other tests flaked once each under parallel load while measuring — `find.perf.test.ts` "pure evaluation overhead" (a 5 s-timeout micro-benchmark) and `defense-in-depth-exploit-regression.test.ts` "compat mode (audit)" (cross-file global state). Both pass in isolation, neither appeared in the run reported above, and neither touches loops or redirections.

Also clean: `pnpm typecheck`, `pnpm lint` (biome + workflow security + `lint:banned`), `pnpm lint:fix`, `pnpm knip`.

### DIFF
--- a/.changeset/loop-output-redirections.md
+++ b/.changeset/loop-output-redirections.md
@@ -1,0 +1,5 @@
+---
+"just-bash": patch
+---
+
+Apply output redirections attached to `while` and `until` loops. `while true; do echo x; break; done >/dev/null` no longer leaks its output to the caller, and `> file`, `>>`, `2>`, `2>&1`, `&>` and `>|` now behave on loops the way they already did on `for` and `case`. `until` loops also gained the input-redirection handling `while` loops already had, so `until ! read l; do ...; done < file` reads from the file. A loop now only restores stdin it owns, so reading inside a loop no longer rewinds an enclosing group's read position: `printf 'a\nb\n' | { while read x; do break; done; read y; }` sees `y=b`.
--- a/packages/just-bash/src/comparison-tests/fixtures/loop-redirection-order.comparison.fixtures.json
+++ b/packages/just-bash/src/comparison-tests/fixtures/loop-redirection-order.comparison.fixtures.json
@@ -1,0 +1,96 @@
+{
+  "029ef9a4af22f501": {
+    "command": "while true; do echo x; break; done > o.txt < nosuch\necho \"rc=$?\"\necho \"o=[$(cat o.txt)]\"",
+    "files": {
+      "o.txt": "keep\n"
+    },
+    "stdout": "rc=1\no=[]\n",
+    "stderr": "/bin/bash: nosuch: No such file or directory\n",
+    "exitCode": 0,
+    "locked": true
+  },
+  "4056dc47122a0f15": {
+    "command": "while true; do echo x; break; done > a.txt < nosuch > b.txt\necho \"rc=$?\"\necho \"a=[$(cat a.txt)] b=[$(cat b.txt)]\"",
+    "files": {
+      "a.txt": "keepa\n",
+      "b.txt": "keepb\n"
+    },
+    "stdout": "rc=1\na=[] b=[keepb]\n",
+    "stderr": "/bin/bash: nosuch: No such file or directory\n",
+    "exitCode": 0,
+    "locked": true
+  },
+  "55c304dbea2f9c5f": {
+    "command": "until false; do echo u; break; done > o.txt < nosuch\necho \"rc=$?\"\necho \"o=[$(cat o.txt)]\"",
+    "files": {
+      "o.txt": "keep\n"
+    },
+    "stdout": "rc=1\no=[]\n",
+    "stderr": "/bin/bash: nosuch: No such file or directory\n",
+    "exitCode": 0,
+    "locked": true
+  },
+  "5ba4d3c97b929674": {
+    "command": "while true; do echo x; break; done > a.txt < nosuch > b.txt\necho \"rc=$?\"\necho \"a=$([ -e a.txt ] && echo yes || echo no) b=$([ -e b.txt ] && echo yes || echo no)\"",
+    "files": {},
+    "stdout": "rc=1\na=yes b=no\n",
+    "stderr": "/bin/bash: nosuch: No such file or directory\n",
+    "exitCode": 0,
+    "locked": true
+  },
+  "625853946873d829": {
+    "command": "while true; do echo x; break; done 2> e.txt < nosuch\necho \"rc=$?\"\necho \"hits=$(grep -c 'No such file or directory' e.txt)\"",
+    "files": {
+      "e.txt": "keep\n"
+    },
+    "stdout": "rc=1\nhits=1\n",
+    "stderr": "",
+    "exitCode": 0
+  },
+  "91da5cb96c38ee37": {
+    "command": "while read l; do echo \"h:$l\"; done > o.txt <<EOF < nosuch\nalpha\nEOF\necho \"rc=$?\"\necho \"o=[$(cat o.txt)]\"",
+    "files": {
+      "o.txt": "keep\n"
+    },
+    "stdout": "rc=1\no=[]\n",
+    "stderr": "/bin/bash: line 2: nosuch: No such file or directory\n",
+    "exitCode": 0,
+    "locked": true
+  },
+  "c37167907026f157": {
+    "command": "while true; do echo x; break; done >> o.txt < nosuch\necho \"rc=$?\"\necho \"o=[$(cat o.txt)]\"",
+    "files": {
+      "o.txt": "keep\n"
+    },
+    "stdout": "rc=1\no=[keep]\n",
+    "stderr": "/bin/bash: nosuch: No such file or directory\n",
+    "exitCode": 0,
+    "locked": true
+  },
+  "d15827827b1e9040": {
+    "command": "while true; do echo x; break; done >> o.txt < nosuch\necho \"rc=$?\"\necho \"exists=$([ -e o.txt ] && echo yes || echo no) o=[$(cat o.txt)]\"",
+    "files": {},
+    "stdout": "rc=1\nexists=yes o=[]\n",
+    "stderr": "/bin/bash: nosuch: No such file or directory\n",
+    "exitCode": 0,
+    "locked": true
+  },
+  "e6247bbd96681489": {
+    "command": "while true; do echo x; break; done < nosuch > o.txt\necho \"rc=$?\"\necho \"o=[$(cat o.txt)]\"",
+    "files": {
+      "o.txt": "keep\n"
+    },
+    "stdout": "rc=1\no=[keep]\n",
+    "stderr": "/bin/bash: nosuch: No such file or directory\n",
+    "exitCode": 0,
+    "locked": true
+  },
+  "f77f9201e0a8db50": {
+    "command": "n=0\nwhile true; do echo x; break; done > \"s$((n++)).txt\" < nosuch\necho \"rc=$? n=$n\"\necho \"s0=$([ -e s0.txt ] && echo yes || echo no) s1=$([ -e s1.txt ] && echo yes || echo no)\"",
+    "files": {},
+    "stdout": "rc=1 n=1\ns0=yes s1=no\n",
+    "stderr": "/bin/bash: line 1: nosuch: No such file or directory\n",
+    "exitCode": 0,
+    "locked": true
+  }
+}
--- a/packages/just-bash/src/comparison-tests/fixtures/loop-redirections-stdin.comparison.fixtures.json
+++ b/packages/just-bash/src/comparison-tests/fixtures/loop-redirections-stdin.comparison.fixtures.json
@@ -1,0 +1,106 @@
+{
+  "1bb58c023a7e3d11": {
+    "command": "printf 'p1\\np2\\n' | until ! read l; do echo \"u:$l\"; done > out.txt\ncat out.txt",
+    "files": {},
+    "stdout": "u:p1\nu:p2\n",
+    "stderr": "",
+    "exitCode": 0
+  },
+  "1e0e9742faa27b5b": {
+    "command": "while read l; do echo \"s:$l\"; done <<< \"hello\" > out.txt\ncat out.txt",
+    "files": {},
+    "stdout": "s:hello\n",
+    "stderr": "",
+    "exitCode": 0
+  },
+  "39035298fab339d9": {
+    "command": "while read l; do echo \"h:$l\"; done > out.txt <<EOF\nalpha\nbeta\nEOF\ncat out.txt",
+    "files": {},
+    "stdout": "h:alpha\nh:beta\n",
+    "stderr": "",
+    "exitCode": 0
+  },
+  "40fe41c114c96981": {
+    "command": "while read l; do echo \"s:[$l]\"; done <<< *",
+    "files": {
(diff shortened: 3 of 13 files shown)


trieloff:

PR #330, opened 2026-07-30
fix(interpreter): groups no longer rewind stdin they never replaced

trieloff:
## Summary

A command group put the shell's stdin read position back even when it never replaced it, throwing away everything its body had read. With a two-line file:

```bash
{ { read a; }; read b; echo "b=[$b]"; } < file
# bash:      b=[L2]
# just-bash: b=[L1]
```

The inner group has no redirection of its own, so it shares the shell's fd 0 — the `read` inside it consumes the first line and the next `read` must get the second. just-bash restored the pre-group position instead, so the line came back and every read after a group replayed input. The same restore sits on the path that runs function bodies (a function body *is* a group) and on `eval`, where it was worse: `{ eval 'while read l; do echo "E:$l"; done'; } < file` printed **nothing at all**, because `eval` also installed an empty stdin over the shell's before running its body.

No issue is filed for this. It was found while verifying **#328**, which fixes the pipeline half of the same root pattern: a construct writing to the shared `state.groupStdin` that it does not own. #328 fixes `executePipeline` (a pipeline drained stdin to EOF); this PR fixes the group/function/`eval` side (they rewound it). Different constructs, different files — `pipeline-execution.ts` there, `subshell-group.ts` + `builtins/eval.ts` + the ownership plumbing here — so the two compose. This branch is cut from `upstream/main`, which does **not** contain #328 yet.

Two cases need both patches and are still wrong on this branch alone: `{ { read a; echo "a=[$a]"; } | cat; read b; }` and `printf … | eval '…'` inside a redirected group. With #328's `finally` hand-back in place they resolve — #328 makes the pipeline give the position back, this PR makes the group inside the stage stop rewinding it.

### Root cause

`executeGroup` saved and restored unconditionally, but installed conditionally:

```ts
const savedGroupStdin = ctx.state.groupStdin;
if (effectiveStdin) {                        // only when the group had its own stdin
  ctx.state.groupStdin = effectiveStdin;
}
…
ctx.state.groupStdin = savedGroupStdin;      // always
```

`eval` had the same shape plus a second defect: `const effectiveStdin = stdin ?? ctx.state.groupStdin` — the builtin dispatcher hands `eval` a plain string, so a command with no redirection arrives as `""`, not `undefined`. `"" ?? …` is `""`, so `eval` installed an empty stdin (immediate EOF) over the shell's position on *every* call, then rewound it on the way out.

### The rule

A construct restores only what it actually replaced.

- **It owns fd 0** — `{ …; } < file`, a heredoc or here-string on the group, a pipe into it, a redirection on a function definition or call, `eval '…' < file`. The enclosing shell's position must come back untouched by whatever the body read.
- **It shares fd 0** — no redirection, no pipe. Reads inside move the one shared position and the command after the construct must see where the body left off.

`executeGroup` now tracks that as an explicit `ownsStdin` flag and routes both the normal and the error path through one `restoreGroupStdin()`. `eval` uses the same rule.

### Ownership is a fact, not an inference from the content

An empty stdin is ambiguous: `f < empty-file` and an unredirected `f` both reach the callee as `""`, and they mean opposite things — EOF versus "inherit the shell's stdin". Inferring ownership from `stdin !== ""` gets the second right and the first wrong, for a whole class of constructs, with a loop-count failure mode:

```bash
f() { read a; }; while read x; do f < /dev/null; done < five-line-file
# shielding the body from the loop's stdin is the idiom; inferring gives 3 turns, not 5
```

So the fact travels beside the string instead. `executeSimpleCommandInner` already knows it — it is the branch that assigned `stdin` — and passes `stdinRedirected` through `runCommand` → `dispatchBuiltin` to `handleEval` and `callFunction`, alongside the existing `stdinSourceFd`. `callFunction` combines it with its own definition-level redirections (`processInputRedirections` now reports `{ stdin, redirected }` rather than a bare string) and hands the result to `executeCommand(func.body, stdin, stdinOwned)`, which reaches `executeGroup` / `executeSubshell`. Every new parameter is optional and defaults to `false`, so no caller that does not pass one changes behavior.

That closes the class: `f < empty.txt`, `f() { …; } < empty.txt`, `f < /dev/null`, `eval '…' < empty.txt`, `eval '…' < /dev/null` and `{ …; } < empty.txt` all mean EOF inside now, and all five loop-shield idioms run once per line.

The wrapper builtins re-enter dispatch with the *same* stdin, so they forward the flag as well: `command`, `builtin` and `exec` each pass it to the command they wrap. Without that, `command eval '…' < empty-file` looked unredirected to the wrapped `eval` — it read the enclosing shell's line instead of EOF, and `while read q; do command eval 'read x' < /dev/null; done < file` ran three times instead of five. `command`/`builtin` still drop `stdinSourceFd` when they re-dispatch, which is the pre-existing `<&N`-through-a-wrapper gap and belongs with #326's descriptor work.

### Why `undefined` is not a position

Both restore points keep one guard: if the body left `ctx.state.groupStdin === undefined` where the construct inherited a defined stream, the inherited position is restored rather than propagated.

A shared stdin can be consumed down to `""` — that is EOF, and it *must* propagate — but it can never be taken away; `undefined` means "this scope has no stdin at all", which is not a read position. Today the only thing that produces it is a pipeline stage clearing shared state it does not own, i.e. exactly the bug #328 fixes. Without the guard, `while read a; do { echo x | cat; }; done < file` — a pipeline nested in a group in a loop body — would run once instead of five times on a branch without #328, because the group would faithfully propagate the pipeline's `undefined`. With #328 merged the guard never fires; four unit tests pin the behavior from the outside so it stays correct either way.

### Bash matrix

90 shapes, each run in `/bin/bash` 3.2.57 and in just-bash, before and after. **45 → 71 matching: 26 fixed, 0 regressed**, 64 unchanged.

| Case | bash | before | after |
|---|---|---|---|
| `{ { read a; }; read b; }` | `L1`,`L2` | `L1`,**`L1`** | `L1`,`L2` |
| `{ { { read a; }; }; read b; }` | `L1`,`L2` | `L1`,**`L1`** | `L1`,`L2` |
| `{ { { read a; }; read c; }; read b; }` | `L1`,`L2`,`L3` | `L1`,`L2`,**`L1`** | `L1`,`L2`,`L3` |
| `{ { 5×read }; read b; echo $? }` | `""`, 1 | **`L1`**, **0** | `""`, 1 |
| `{ { read a; } >/dev/null; read b; }` | `L1`,`L2` | `L1`,**`L1`** | `L1`,`L2` |
| `{ { read a; }; cat; }` | `L2…L5` | **`L1…L5`** | `L2…L5` |
| `{ { read a; }; read b; } <<EOT` / `<<<` | `H1`,`H2` / `S1`,`""` | **replayed** | `H1`,`H2` / `S1`,`""` |
| `{ read a; { read x; } < other; read b; }` | `O1`, `L1`,`L2` | `O1`, `L1`,`L2` | unchanged |
| `{ read a; { read x; } < empty; read b; }` | `""`, `L1`,`L2` | **`L2`**, `L1`,`L2` | `""`, `L1`,`L2` |
| `f() { read a; }; { f; f; read b; }` | `L1`,`L2`,`L3` | `L1`,**`L1`**,**`L1`** | `L1`,`L2`,`L3` |
| `f() { read a; } < other; { f; read b; }` | `O1`,`L1` | `O1`,`L1` | unchanged |
| `{ f < other; read b; }` | `O1`,`L1` | `O1`,`L1` | unchanged |
| `{ read a; f < empty; read b; }` | `""`, `L1`,`L2` | **`L2`**, `L1`,`L2` | `""`, `L1`,`L2` |
| `f() { …; } < empty; { read a; f; read b; }` | `""`, `L1`,`L2` | **`L2`**, `L1`,`L2` | `""`, `L1`,`L2` |
| `{ read a; f < /dev/null; read b; }` | `""`, `L1`,`L2` | **`L2`**, `L1`,`L2` | `""`, `L1`,`L2` |
| `while read a; do …; f; done` (f reads) | 3 turns | **5 turns, replayed** | 3 turns |
| `while read x; do f < /dev/null; done` | 5 turns | 5 turns | 5 turns |
| `while read x; do f < empty; done` | 5 turns | 5 turns | 5 turns |
| `while read x; do eval 'read q' < /dev/null; done` | 5 turns | 5 turns | 5 turns |
| `while read x; do { read q; } < /dev/null; done` | 5 turns | 5 turns | 5 turns |
| `{ eval 'read a'; read b; }` | `L1`,`L2` | **`""`**,**`L1`** | `L1`,`L2` |
| `{ eval 'read a; read c'; read b; }` | `L1`,`L2`,`L3` | **`""`,`""`,`L1`** | `L1`,`L2`,`L3` |
| `{ eval 'while read l; …'; } < file` | `L1…L5` | **(nothing)** | `L1…L5` |
| `{ eval '…' < other; read b; }` | `O1`,`L1` | `O1`,`L1` | unchanged |
| `{ read a; eval 'read x' < empty; read b; }` | `""`, `L1`,`L2` | `""`, `L1`,`L2` | unchanged |
| `{ command eval 'read x' < empty; read y; }` | `""`,`L1` | `""`,`L1` | unchanged |
| `{ command eval 'read x'; read y; }` | `L1`,`L2` | **`""`**,**`L1`** | `L1`,`L2` |
| `{ builtin eval 'read x'; read y; }` | `L1`,`L2` | **`""`**,**`L1`** | `L1`,`L2` |
| `while read q; do command eval 'read x' < /dev/null; done` | 5 turns | 5 turns | 5 turns |
| `while read a; do { read b; }; done` | 3 turns | **5 turns, replayed** | 3 turns |
| `while read a; do …; done < file` (bare) | 5 turns | 5 | 5 |
| loop body: group / subshell / function / eval | 5 turns | 5 | 5 |
| loop body: pipeline in group / function / eval / subshell | 5 turns | 5 | 5 |
| `{ { echo x \| cat; }; read b; }` | `L1` | `L1` | `L1` |
| `{ { true \| true; }; read b; }` | `L1` | `L1` | `L1` |
| `for` / `until` / `case` / `if` body reads | advance | advance | unchanged |

Two rows still differ from bash on both sides but changed value, both for the better:

- `{ eval 'while read l; …'; read b; } < file` — the loop body now prints all five lines instead of nothing. The row stays DIFF only because of the `while` residual below (`b=[L1]` where bash has `b=[]`).
- `{ { mapfile -t arr; }; echo ${#arr[@]}; read b; }` — `b` went from `L1` to `""`, which is what bash 5 prints. The recorded reference is bash 3.2.57, which has no `mapfile` at all, so this row cannot be judged against it; it is not covered by a test.

### Subshells: deliberately not touched

`( read a )` in bash advances the parent's offset for a seekable stdin (`{ ( read a ); read b; }` gives `b` the second line); just-bash gives it the first, because `executeSubshell` runs the body under `beginIsolatedShellState`, which snapshots and restores `groupStdin` along with the rest of the namespace. That is the model, not this bug: #326's *Deferred* section states it for descriptors — "subshells copy the descriptor table rather than sharing the open file description … just-bash already copies stdin the same way, so this stays consistent with the existing model. Not covered by a test, to avoid pinning behavior we intend to change." Changing it here would mean making subshell state selectively non-isolated, which is a model change, not a restore-rule fix. No test in this PR pins subshell stdin behavior in either direction.

### Known residual divergences — all measured, none introduced or changed by this PR

Each was run on `upstream/main` and on this branch and produced byte-identical just-bash output on both.

- **`while`/`until` loops have the same unconditional restore** in `control-flow.ts`: `{ while read a; do …; done; read b; }` gives `b` the first line where bash gives EOF. It is the same two-line rule, but `control-flow.ts` is being rewritten by a sibling PR for loop *output* redirections; applying it here would guarantee a conflict for a case that only shows up when a loop shares stdin with a following command. Left for a follow-up. The core `done < file` idiom is unaffected and covered by tests here.
- **Non-`read` commands do not advance the shared position**: `{ read a; cat; read b; }` — `cat` prints the rest but leaves the position, so the following `read` gets a line bash does not. #326 documents the same gap for `<&N`.
- **`read` is the one member of the empty-fd-0 class left open.** `{ read a; read x < empty.txt; read b; }` gives `x=[L2]`, `b=[L3]` where bash gives `x=[]`, `b=[L2]`: `handleRead` falls back to `groupStdin` whenever its own stdin is empty. The `stdinRedirected` fact now reaches `dispatchBuiltin`, so closing it is a small follow-up, but `read` is the most heavily exercised builtin in the suite and the change deserves its own PR and its own matrix.
- **`( … ) < file` never installs the file as stdin.** `executeSubshell` ignores input redirections entirely, so `( read x ) < other.txt` reads the *outer* stdin and `( { read a; }; read b; ) < loop.txt` reads nothing at all.
- **`exec < file` never installs stdin.** `{ exec < other.txt; read a; }` reads `L1` from the enclosing redirect and bare `exec < other.txt; read a` reads nothing, where bash gives `O1` for both. `exec`'s redirection handling only populates `fileDescriptors`; fd 0 is not routed to `groupStdin`.
- **`exec` runs builtins that bash refuses to exec.** `exec eval '…'` reports `exec: eval: not found` (status 127) in bash — `exec` only replaces the shell with an external program — where just-bash's `exec` stub runs it. Unrelated to stdin; the ownership flag is forwarded through `exec` anyway so the wrapped command is correct on the day that stub is replaced.
- **A group's `<` branch ignores the descriptor number.** `{ read x; } 3< other.txt` feeds `other.txt` to *stdin*: `{ read a; { read x; } 3< other.txt; read b; }` gives `x=[O1]`, `b=[L2]` where bash gives `x=[L2]`, `b=[L3]`. The heredoc branch beside it checks `redir.fd ?? 0`; the `<` and `<<<` branches do not. This overlaps open PR **#326**, which routes every fd ≥ 3 construct through a typed descriptor table, so the fix belongs there rather than in a conflicting parallel edit.

## Tests

| File | Tests | Lines |
| --- | --- | --- |
| `packages/just-bash/src/interpreter/group-stdin.test.ts` | 22 | 243 |
| `packages/just-bash/src/interpreter/group-stdin.constructs.test.ts` | 34 | 264 |
| `packages/just-bash/src/interpreter/group-stdin.wrappers.test.ts` | 11 | 87 |
| `packages/just-bash/src/comparison-tests/group-stdin.comparison.test.ts` | 25 | 265 |

92 new tests. `group-stdin.test.ts` opens with the exact repro. All assertions compare full stdout and stderr strings plus the exit code; every file is under the 300-line limit. **27 of the 67 unit tests fail on `upstream/main`**; the other 40 are guards that must keep passing (read-loop idioms, loop-body shields, redirections that *do* own stdin, pipelines in a shared-stdin body, and the wrapper cases `upstream/main` happens to get right).

Comparison fixtures were recorded against **GNU bash 3.2.57(1)-release**, `/bin/bash` on macOS (`RECORD_FIXTURES=1`). None are locked: every command is plain POSIX `read`/`echo`/`cat` with no diagnostics on stderr, so a re-record on the CI runner's bash 5 produces byte-identical entries. Re-recording on this machine produces no diff.

Verification on this branch, measured against `upstream/main` in the same checkout with `pnpm build` run first in both cases:

| Suite | `upstream/main` | this branch |
| --- | --- | --- |
| `pnpm test:run` | 6 failed, 14287 passed, 97 skipped (14390) | 6 failed, **14379** passed, 97 skipped (14482) |
| `pnpm test:comparison` | 567 passed (35 files) | **592** passed (36 files) |
| `src/spec-tests` | 3825 passed, 1 skipped | 3825 passed, 1 skipped |

Zero new failures. The 6 failures are identical on both sides and are the CPython-WASM suite: `vendor/cpython-emscripten/python.wasm` is an unmaterialized git-LFS pointer in this checkout (132 bytes), so `python: true` tests time out loading the worker (`just-bash.bundle`, `code-exec-exploit-regression`, `defense-in-depth-independence`, `python-sqlite-information-disclosure` ×2, `worker-protocol-runtime-desync`). They fail in isolation here too, on main and on the branch, for the same reason. No existing test asserted the old behavior, so none was changed.

`pnpm typecheck`, `pnpm lint:fix` (biome), `pnpm lint:banned` and `pnpm knip` are clean.

One CI run of this branch failed `packages/just-bash-executor` › "should list all items from a discovered tool" — a js-exec worker returning truncated stdout with status 0 and empty stderr — on the Node 24 job while the **same commit passed the Node 22 job**, which runs that suite too. The failure is in `commands/js-exec` + `commands/worker-bridge`, both byte-identical to `upstream/main` here (this PR touches only `src/interpreter/`), and the signature is only reachable inside `BridgeHandler`. 180 repeat runs of that suite on this branch (including 70 on Node 24, under CPU saturation and against a concurrently running just-bash unit suite) and 80 on `upstream/main` produced 0 failures on either side, so it looks like a pre-existing worker-output race rather than anything this PR introduces.

### DIFF
--- a/.changeset/eighty-jars-attack.md
+++ b/.changeset/eighty-jars-attack.md
@@ -1,0 +1,5 @@
+---
+"just-bash": patch
+---
+
+Stop command groups, function bodies and `eval` from rewinding stdin they never replaced, so `{ { read a; }; read b; }` gives `b` the second line instead of replaying the first.
--- a/packages/just-bash/src/comparison-tests/fixtures/group-stdin.comparison.fixtures.json
+++ b/packages/just-bash/src/comparison-tests/fixtures/group-stdin.comparison.fixtures.json
@@ -1,0 +1,234 @@
+{
+  "07b73d845debc1bd": {
+    "command": "{ { { read a; }; }; read b; echo \"a=[$a] b=[$b]\"; } < loop.txt",
+    "files": {
+      "loop.txt": "L1\nL2\nL3\nL4\nL5\n"
+    },
+    "stdout": "a=[L1] b=[L2]\n",
+    "stderr": "",
+    "exitCode": 0
+  },
+  "0c9e0ea31e2c8880": {
+    "command": "{ read a; while read x; do echo \"X:$x\"; done < other.txt; read b; echo \"a=[$a] b=[$b]\"; } < loop.txt",
+    "files": {
+      "loop.txt": "L1\nL2\nL3\nL4\nL5\n",
+      "other.txt": "O1\nO2\nO3\n"
+    },
+    "stdout": "X:O1\nX:O2\nX:O3\na=[L1] b=[L2]\n",
+    "stderr": "",
+    "exitCode": 0
+  },
+  "124dbc51aae2bf1d": {
+    "command": "{ eval 'read a; echo \"a=[$a]\"' < other.txt; read b; echo \"b=[$b]\"; } < loop.txt",
+    "files": {
+      "loop.txt": "L1\nL2\nL3\nL4\nL5\n",
+      "other.txt": "O1\nO2\nO3\n"
+    },
+    "stdout": "a=[O1]\nb=[L1]\n",
+    "stderr": "",
+    "exitCode": 0
+  },
+  "28a816d2d3632e61": {
+    "command": "f() { read x; echo \"x=[$x]\"; } < empty.txt; { read a; f; read b; echo \"a=[$a] b=[$b]\"; } < loop.txt",
+    "files": {
+      "loop.txt": "L1\nL2\nL3\nL4\nL5\n",
+      "empty.txt": ""
+    },
+    "stdout": "x=[]\na=[L1] b=[L2]\n",
+    "stderr": "",
+    "exitCode": 0
+  },
+  "2ccccd0e98d641dd": {
+    "command": "{ eval 'while read l; do echo \"E:$l\"; done'; } < loop.txt",
+    "files": {
+      "loop.txt": "L1\nL2\nL3\nL4\nL5\n"
+    },
+    "stdout": "E:L1\nE:L2\nE:L3\nE:L4\nE:L5\n",
+    "stderr": "",
+    "exitCode": 0
+  },
+  "3a0c611cda10a437": {
+    "command": "{ eval 'read a; read c'; read b; echo \"a=[$a] c=[$c] b=[$b]\"; } < loop.txt",
+    "files": {
+      "loop.txt": "L1\nL2\nL3\nL4\nL5\n"
+    },
+    "stdout": "a=[L1] c=[L2] b=[L3]\n",
+    "stderr": "",
+    "exitCode": 0
+  },
+  "583d99b0e8a38d6f": {
+    "command": "{ read a; { read x; echo \"x=[$x]\"; } < empty.txt; read b; echo \"a=[$a] b=[$b]\"; } < loop.txt",
+    "files": {
+      "loop.txt": "L1\nL2\nL3\nL4\nL5\n",
+      "empty.txt": ""
+    },
+    "stdout": "x=[]\na=[L1] b=[L2]\n",
+    "stderr": "",
+    "exitCode": 0
+  },
+  "5b9b4e1d1a1ebc39": {
+    "command": "{ { read a; echo \"a=[$a]\"; } <<< \"S1\"; read b; echo \"b=[$b]\"; } < loop.txt",
+    "files": {
+      "loop.txt": "L1\nL2\nL3\nL4\nL5\n"
+    },
+    "stdout": "a=[S1]\nb=[L1]\n",
+    "stderr": "",
+    "exitCode": 0
+  },
+  "5bb969bda6c97a80": {
+    "command": "{ { read a; echo \"a=[$a]\"; } <<EOT\nH1\nEOT\nread b; echo \"b=[$b]\"; } < loop.txt",
+    "files": {
+      "loop.txt": "L1\nL2\nL3\nL4\nL5\n"
+    },
+    "stdout": "a=[H1]\nb=[L1]\n",
+    "stderr": "",
+    "exitCode": 0
+  },
+  "6723871c6848c7a5": {
+    "command": "n=0; while read a; do n=$((n+1)); { echo x | cat; } >/dev/null; done < loop.txt; echo \"iterations: $n\"",
+    "files": {
+      "loop.txt": "L1\nL2\nL3\nL4\nL5\n"
+    },
+    "stdout": "iterations: 5\n",
+    "stderr": "",
+    "exitCode": 0
+  },
+  "6b84ec4bde6c8e9a": {
+    "command": "{ read a; { read x; echo \"x=[$x]\"; } < other.txt; read b; echo \"a=[$a] b=[$b]\"; } < loop.txt",
+    "files": {
+      "loop.txt": "L1\nL2\nL3\nL4\nL5\n",
+      "other.txt": "O1\nO2\nO3\n"
+    },
+    "stdout": "x=[O1]\na=[L1] b=[L2]\n",
+    "stderr": "",
+    "exitCode": 0
+  },
+  "731214683b5fc3a6": {
+    "command": "{ { read a; }; cat; } < loop.txt",
+    "files": {
+      "loop.txt": "L1\nL2\nL3\nL4\nL5\n"
+    },
+    "stdout": "L2\nL3\nL4\nL5\n",
+    "stderr": "",
+    "exitCode": 0
+  },
+  "760a355a7212adc4": {
+    "command": "{ read a; eval 'read x; echo \"x=[$x]\"' < empty.txt; read b; echo \"a=[$a] b=[$b]\"; } < loop.txt",
+    "files": {
+      "loop.txt": "L1\nL2\nL3\nL4\nL5\n",
+      "empty.txt": ""
+    },
+    "stdout": "x=[]\na=[L1] b=[L2]\n",
+    "stderr": "",
+    "exitCode": 0
+  },
+  "7a0dda5553691c92": {
+    "command": "{ { read a; } >/dev/null; read b; echo \"a=[$a] b=[$b]\"; } < loop.txt",
+    "files": {
+      "loop.txt": "L1\nL2\nL3\nL4\nL5\n"
+    },
+    "stdout": "a=[L1] b=[L2]\n",
+    "stderr": "",
+    "exitCode": 0
+  },
+  "9521f508de0a5238": {
+    "command": "f() { read a; echo \"in=[$a]\"; }; { f < other.txt; read b; echo \"b=[$b]\"; } < loop.txt",
+    "files": {
+      "loop.txt": "L1\nL2\nL3\nL4\nL5\n",
+      "other.txt": "O1\nO2\nO3\n"
+    },
+    "stdout": "in=[O1]\nb=[L1]\n",
+    "stderr": "",
+    "exitCode": 0
+  },
+  "97e5ad35e4af1b3d": {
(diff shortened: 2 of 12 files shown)


trieloff:

mdjahid11978-design (commented):
Fixed All set up not my repo

PR #333, opened 2026-07-30
combo: numeric file descriptors (#326) + loop redirections (#329), pre-resolved

trieloff:
## What this is

A **pre-resolved merge** of #326 (numeric file descriptors) and #329 (loop output redirections, now including its strictly-left-to-right ordering fix), offered purely for your convenience. The two branches conflict in `packages/just-bash/src/interpreter/control-flow.ts`, and the conflict is semantic rather than textual — whoever lands second has to reconcile four separate things by hand, one of which is a silent index-space bug. That work is already done here.

> **Rebuilt on #329's second commit** `b8f87f6` ("process loop redirections strictly left to right"), which reworked `prepareLoopRedirections` into an ordered pass after the Vercel review bot flagged that a failing `< file` aborted before pre-opening output redirects standing to its **left**. That commit touches exactly the machinery this merge has to reconcile, and it introduced a second index-rebasing scheme that has to compose with #326's — see decision 3.

**How to use it:**

* Want both? Merge **this** PR and close #326 and #329.
* Prefer to review them one at a time? Merge #326 and #329 individually and close this — the reconciliation notes below are then a checklist for whoever lands second. Closing this costs nothing; nothing else depends on it.

Nothing new is invented here. This branch is `feat/numeric-fds-321` with `fix/while-loop-redirect` merged into it: **both branches' original signed commits are preserved as parents**, and the entire reconciliation lives in the single merge commit on top. `git log --graph` shows two intact histories and one merge; `git diff <this> feat/numeric-fds-321` is exactly #329's contents, and `git diff <this> fix/while-loop-redirect` is exactly #326's.

## The reconciliation, decision by decision

#329's PR body listed three things the second-lander must do. All three are done, plus the index-space composition that only exists once both changes are in the same function.

### 1. `prepareLoopRedirections()` stays the front door; `withNumericFds` stays the outer layer

#326 wraps every compound command in `withNumericFds(...)`; #329 introduced `prepareLoopRedirections()` as the single redirection prologue for `while`/`until`, now an ordered left-to-right walk. Neither was dropped. The merged shape is:

```
executeWhile / executeUntil          <- public entry, keeps #329's `stdin` param
  └─ withNumericFds(node.redirections, fdTargets => ...)   <- #326: fd table up
       └─ executeWhileBody / executeUntilBody(ctx, node, stdin, fdTargets)
            └─ prepareLoopRedirections(ctx, node.redirections, fdTargets)  <- #329
```

The fd table is therefore installed **before** the loop's redirection prologue runs and torn down **after** the loop finishes, which is what makes `done 3< file` visible to `read -u 3` inside the body and restored afterwards.

### 2. `isNumericFdRedirection` re-applied inside the ordered walk

#329 moved `while`'s input-redirection scan into `prepareLoopRedirections`, which is exactly the loop #326 had added its `isNumericFdRedirection(redir) continue;` guard to. The guard is back, and its placement in the new ordered walk matters:

```ts
for (let index = 0; index < redirections.length; index++) {
  const redir = redirections[index];
  if (!LOOP_STDIN_OPERATORS.has(redir.operator)) {
    pending.push(redir);
    pendingSources.push(index);
    continue;
  }

  // `done 3< file` names a descriptor, not this loop's stdin — the fd table
  // already holds it. It is not an ordering barrier either, since the whole
  // numeric pass ran before this walk, so it does not close the pending run.
  if (isNumericFdRedirection(redir)) continue;

  const openError = await openPending();
  ...
```

Two properties, both required:

* A `3<`-family redirection is **not loop stdin** — without the guard, `done 3< file` is swallowed as the loop's stdin, the wart #329 documented and explicitly deferred to #326's area.
* It is **not an ordering barrier**. `withNumericFds` has already run to completion by the time this walk starts, so a numeric fd sitting between two output redirects must not split them into two pending runs.

Numeric redirections with output operators (`3>`, `3>>`, `3<>`, `3>&1`, `4<&3`) ride along in the pending run, where `preOpenOutputRedirects` and `applyRedirections` both ignore them by descriptor number — keeping the list index-coherent and matching how `for`/`case` pass their full redirection array.

### 3. `fdTargets` threaded through the walk — composing TWO index spaces

This is the piece neither PR could have written on its own, and the most likely place for a silent bug.

#326 added a third `alreadyExpanded` argument to `preOpenOutputRedirects` so a redirect target with side effects (`3> "f$((n++))"`) is expanded exactly once — the numeric pass expands it, and the stdout/stderr pass reuses the value. #329's new ordered walk calls `preOpenOutputRedirects` **once per pending run**, and rebases each run's returned target keys by `opened.length` before merging them into the accumulated map.

`ExpandedRedirectTargets` is a `Map<number, string>` **keyed by position in the array handed to `preOpenOutputRedirects`**, so there are now three index spaces to keep straight:

| space | keyed by | who owns it |
| --- | --- | --- |
| **original** | position in `node.redirections` | `fdTargets`, produced by #326's `withNumericFds` |
| **run-local** | position in the current pending `run` | the argument to and result from `preOpenOutputRedirects` |
| **opened** | position in the accumulated `opened` array | `targets`, consumed by `applyRedirections` |

#329 handles run-local → opened. #326 assumes original → run-local is the identity, which it is for `for`/`case` (they pass the whole array) but **is not** for a pending run. The merge adds `pendingSources`, a parallel array of original positions, and re-keys `fdTargets` onto each run just before the call:

```ts
const runFdTargets: ExpandedRedirectTargets = new Map();
for (let i = 0; i < sources.length; i++) {
  const expanded = fdTargets.get(sources[i]);
  if (expanded !== undefined) runFdTargets.set(i, expanded);
}
const prepared = await preOpenOutputRedirects(ctx, run, runFdTargets);
// ... then #329's existing rebase:
const offset = opened.length;
for (const [index, target] of prepared.targets) targets.set(offset + index, target);
opened.push(...run);
```

**What goes wrong without it.** `preExpandRedirectTargets` starts from `new Map(alreadyExpanded)`, so passing `fdTargets` unrebased leaks original-space keys straight into a run-local map. For `done 3> a < in > b` the original indices are `0 = 3> a`, `1 = < in`, `2 = > b`; the `< in` closes the first run, so `> b` is alone in run 2 at run-local index **0** — where the leaked `fdTargets` entry `{0: "a"}` already sits. `preExpandRedirectTargets` sees `targets.has(0)`, skips expanding `b`, and `applyRedirections` writes the loop's entire stdout into **`a`**. Silent, no error, wrong file.

Verified fixed, byte-identical to bash (case C1 in the table below):

```bash
while read x; do echo "body:$x"; echo "tofd3:$x" >&3; done 3> a.txt < in.txt > b.txt
# a.txt = tofd3:i1 / tofd3:i2      (fd 3 writes)
# b.txt = body:i1 / body:i2        (loop stdout)
```

### 4. `executeUntil` gets both halves of its signature

#329 added a `stdin` parameter to `executeUntil` (so `printf 'a\nb\n' | until ! read l; ...` works) and a matching dispatch change in `interpreter.ts`. #326 split the function into `executeUntil` / `executeUntilBody` with **no** stdin parameter. Merged:

```ts
export async function executeUntil(ctx, node, stdin = "") {
  return withNumericFds(ctx, node.redirections, (fdTargets) =>
    executeUntilBody(ctx, node, stdin, fdTargets));
}
async function executeUntilBody(ctx, node, stdin: string, fdTargets: ExpandedRedirectTargets) { ... }
```

`executeWhileBody` carries the same four arguments. `interpreter.ts` merged cleanly and already passes `stdin` to both.

## What else comes along: #329's left-to-right ordering fix

Because this branch is built on #329's current head, it also carries `b8f87f6`, which makes the loop's redirection list process strictly left to right: a failing `< file` now leaves every redirection to its **left** already applied, and routes its own diagnostic through them. So `done 2> err < nosuch` lands the message in `err`, and `done > out < nosuch` truncates `out` — while `done < nosuch > out` still leaves `out` alone. That behavior is #329's; it is listed here only so you know the merge contains it and that the numeric-fd reconciliation was done against the ordered walk, not the older two-phase version.

## Bonus the combination delivers: `done 3< file` on `while` **and** `until`

#329 listed this as a known deviation:

> `done 3< in` is consumed as the loop's stdin rather than being bound to fd 3. That is a pre-existing wart in `while`'s input handling which `until` now inherits; numeric fds are #326's area.

In the merged tree it is simply fixed, and it falls out of the reconciliation rather than being bolted on: #326's fd machinery exists, #329's helper is the only place that decides what a loop's input redirections mean, and the `isNumericFdRedirection` guard sits between them. `until` gets it for free even though **neither branch alone** could give it — #326 never taught `until` about loop-level input redirections at all, and #329 had `until` consuming `3<` as stdin.

Verified against `/bin/bash` 3.2.57, byte-identical:

```bash
while read -u 3 line; do echo "got:$line"; done 3< /tmp/in.txt   # got:l1 got:l2 got:l3
until ! read -u 3 l;  do echo "u:$l";      done 3< /tmp/in.txt   # u:l1 u:l2 u:l3
while read line; do echo "body:$line"; break; done 3< /tmp/in.txt <<< "fromhere"
                                                                 # body:fromhere  (fd 3 is NOT the loop's stdin)
```

The first of those is precisely the idiom #321 asked for, now working on a redirected loop.

### Both branches' tests, together

| suite | result |
| --- | --- |
| #326 unit files (`fd-table.encoding`, `fd-table.access`, `numeric-fds.input`, `.marker-content`, `.output`, `.scoping`, `.dup-aliasing`) | **7 files, 101/101 passed** |
| #329 unit files (`control-flow.loop-redirections`, `.loop-redirections-flow`, `.loop-redirection-order`, `.loop-stdin-ownership`) | **4 files, 51/51 passed** |
| `pnpm test:comparison` | **39 files, 630/630 passed** — 567 base + 20 (#326) + 43 (#329), all fixture sets replaying together |
| spec suite (`src/spec-tests`) | **3825 passed, 1 skipped, 0 failed** — identical to both branches and to base |
| `pnpm test:run --exclude src/spec-tests` (just-bash) | 5 files failed / 528 passed (533); **6 tests failed** / 10677 passed / 96 skipped (10779) |
| `pnpm test:run` (just-bash-executor) | **4 files, 72/72 passed** |

The 6 failures are the documented CPython-WASM set, identical on both branches and on `main`: `just-bash.bundle.test.ts` (python3 lazy-load), `code-exec-exploit-regression`, `defense-in-depth-independence`, `python-sqlite-information-disclosure` ×2, `worker-protocol-runtime-desync`. `packages/just-bash/vendor/cpython-emscripten/python.wasm` is a **132-byte git-LFS pointer** in this checkout, so every `python: true` test times out loading the worker. **No non-WASM test fails.**

The totals reconcile exactly as a union: 10779 + 3826 = **14605** = 14390 (base) + 94 (#329's new tests across both its commits) + 121 (#326's new tests). Nothing was lost in the merge and nothing was double-counted.

### Fixtures re-record clean

`RECORD_FIXTURES=1` over all four new comparison files (`numeric-fds`, `loop-redirections`, `loop-redirections-stdin`, `loop-redirection-order`) against real bash: **63 passed, and `git status` shows no modification to any fixture JSON.** The merged implementation reproduces both branches' recordings byte-for-byte, so the comparison workflow's no-diff check will pass. The locked fixtures from #326 and #329 are untouched and still locked.

### Static checks

`pnpm typecheck`, `pnpm lint` (workflow security + biome + `lint:banned`), `pnpm lint:fix` (no fixes applied), `pnpm knip` — all clean. Knip reports only the two pre-existing configuration hints that are present on `main`.

### Spot-checked against real bash

Every case below via `pnpm dev:exec --real-bash` against **GNU bash 3.2.57(1)-release**, stdout and stderr byte-identical:

| case | source |
| --- | --- |
| the three #321 repros (`while read -u 3 … done 3< f`, `exec 3< f` persisting across `read -u 3`, `read line <&3`) | #326 |
| dup-aliasing chain `exec 3< f; exec 4<&3; exec 5<&4` → reads `a`, `b`, `c` on one shared offset | #326 |
| `printf 'a\nb\nc\n' \| until ! read l; do …; done` | #329 (#323-adjacent) |
| `printf 'a\nb\nc\n' \| { while read l; do break; done; read r; }` → `r=b` | #329 (#323-adjacent) |
| loop output matrix: `>/dev/null`, `> f`, `>> f`, `2> f`, `> out 2>&1`, `done < in > out` — for `while` **and** `until` | #329 |
| `done 3< file` binds fd 3 on `while` and on `until`; does not feed the body's stdin | the combination |
| `done 3> file` + `echo >&3` inside the loop | the combination |
| `done 3> "f$((n++)).txt"` expands its target exactly once (`n=2`) | the fdTargets remap |

Plus the ordering cases **with numeric fds in play**, which neither parent tests and which only exist once both changes are in the same walk:

| case | result (identical to bash) |
| --- | --- |
| **C1** `done 3> a < in > b` — numeric fd left of the input redirect, output redirect right of it | `a` gets the fd-3 writes, `b` gets the loop's stdout. This is the case that breaks if the two index rebasings do not compose. |
| **C2** `done > o 3< file < nosuch` | `o` truncated (it is left of the failure), exit 1, diagnostic on stderr |
| **C3** `done 3> f > o < nosuch` | both `f` and `o` opened/truncated, exit 1 |
| **C4** `done 3< file` still binds fd 3 on `while` **and** `until` | `w:L1 w:L2` / `u:L1 u:L2` |
| **C5** `done > "s$((n++))" < nosuch` | expanded exactly once — `n=2`, `s1` created, `s2` absent |
| **C6** `done 3> "t$((m++))" < in > b2` | `m=2`, `t1` holds the fd-3 write, `b2` holds stdout |

## One divergence you should know about (inherited from #326, not created here)

`done <&3` with descriptor 3 closed. #329's PR body lists it as fixed (`Bad file descriptor`, exit 1, "same as `for`"). **On the merged tree it is not**: the loop body runs and the loop exits 0.

This is not a merge artifact — it is #326's behavior standing alone, and it applies equally to `for`, `case` and `{ }`. #326 routes `<&N` through the fd table and makes `applyRedirections` skip duplications onto fd 0, which is correct for simple commands (`executeSimpleCommandInner` reports the bad descriptor) but leaves compound commands with no `<&N` pass at all. Measured on all three heads with the same probe:

| | `for … done <&3` | `while … done <&3` | `{ …; } <&3` | `case … esac <&3` | `echo s <&3` |
| --- | --- | --- | --- | --- | --- |
| `main` / #329 branch | rc=1 ✓ | rc=1 ✓ | rc=1 ✓ | rc=1 ✓ | rc=1 ✓ |
| **#326 branch** | rc=0 ✗ | rc=0 ✗ | rc=0 ✗ | rc=0 ✗ | rc=1 ✓ |
| **this merge** | rc=0 ✗ | rc=0 ✗ | rc=0 ✗ | rc=0 ✗ | rc=1 ✓ |
| bash 3.2.57 | rc=1 | rc=1 | rc=1 | rc=1 | rc=1 |

So merging #326 alone, or #326 then #329, or this branch, all land in the same place. The merge is faithful; it just cannot un-do a change made on one of its parents. The positive side of the same gap is also unimplemented on every head: `exec 3< f; while read x; do …; done <&3` reads nothing (bash reads `a`, `b`). Binding a compound command's stdin to a user descriptor needs a `<&N` pass for compounds analogous to the one `executeSimpleCommandInner` has — that belongs in #326 or a follow-up, and was deliberately left out of this merge commit so that this branch contains **nothing but the reconciliation**.

No spec test, comparison fixture or unit test on either branch covers the case, which is why both branches are green.

## Risks

* **The reconciliation is concentrated in one function.** `prepareLoopRedirections` now carries #329's ordered walk and stdin ownership rule, #326's numeric-fd skip, and the original→run-local re-keying. It is exercised by all 152 unit tests plus 63 comparison cases — but it is the place to look if something is off.
* **The `pendingSources` re-keying is the only genuinely new logic**, and it is the highest-risk line in the branch: it has no direct review upstream, and getting it wrong is silent (wrong output file, no error). It is covered by C1 and C6 above; if you review one thing here, review that.
* **This branch will go stale** if #326 or #329 are updated again. It is cheap to regenerate — the conflict resolution is mechanical once the four decisions above are known. It has already been rebuilt once, for #329's `b8f87f6`.
* The `done <&3` divergence documented in the section above.
* Both changesets (`.changeset/numeric-file-descriptors.md`, `.changeset/loop-output-redirections.md`) are present and merged cleanly; the combined release is a patch bump from both.


trieloff:
Superseded by #336, which carries #326 + #329 + #330 in one pre-resolved branch.

#326 conflicts with **both** #329 and #330, so this combo and its sibling were mutually exclusive — merging either one stranded the other. #336 replaces both. Closing to keep the queue unambiguous; every finding documented here is carried forward in #336's description.

PR #334, opened 2026-07-30
combo: grep -f (#327) + repeated -e (#314), pre-resolved

### LINKED ISSUE #297: grep: multiple -e patterns are not OR-combined — only the last one is applied

trieloff:
**Title suggestion:** `combo: grep -f (#327) + repeated -e (#314), pre-resolved`

## What this is

A **convenience merge**, not a new feature. #327 (`grep -f/--file`) and #314 (repeated `-e`) touch the same three files in near-identical ways and conflict. This branch is those two PRs already merged, with the overlap resolved and verified.

**Merge this if you want both #327 and #314 without doing the resolution yourself. Or merge them individually and close this — nothing is lost either way.** The merge commit has both PRs as parents, so this is purely a labour-saving shortcut.

Draft on purpose. No new work of its own: every line comes from one of the two PRs, except the grafting described below.

## Credit

**#314 is [@tylergibbs1](https://github.com/tylergibbs1)'s work** (closes #297). Their three commits are in this branch's history with their authorship intact:

```
57e1912 fix(grep): combine repeated -e patterns          Tyler Gibbs
2da38fe fix(grep): group line regexp alternatives        Tyler Gibbs
ccfa48e chore: add grep changeset                        Tyler Gibbs
```

Both changesets are preserved, so both PRs still get their release notes.

## The overlap

Both PRs independently added:

- a `combinePatterns` helper, at the same insertion point, with the same semantics
- the same `^(?:...)$` fix in `search-engine/regex.ts` so `-x` anchors bind around an alternation
- overlapping removals from the grep spec skip list (~70% the same entries)

Conflicts were in `commands/grep/grep.ts`, `commands/search-engine/regex.ts`, and `spec-tests/grep/skips.ts`.

## Resolution decisions

**`combinePatterns` — kept #327's copy.** The two implementations are functionally equivalent; #327's carries comments explaining why BRE joins with `\|` and why `-F` escapes-then-lifts into extended mode. Nothing of #314's version is lost.

**Repeated `-e` — took #314's behaviour.** This is #314's whole point and it works in the merged tree. `-e` accumulates into a list rather than last-one-wins:

```
$ grep -e apple -e cherry hay.txt
apple
cherry
```

#327 had deliberately left `-e` on last-one-wins, so this had to be grafted in rather than merged.

**Pattern collection is now the union of both features.** Every `-e` operand in argument order (each of which may itself be a newline-separated PATTERNS list, per #327), then every `-f` file in order. All of #327's guards moved up to cover `-e`-collected patterns too:

- `-P` refuses more than one *distinct* pattern, duplicates folded first (GNU accepts `-P -e a -e a`, rejects `-P -e a -e b`)
- an empty pattern list short-circuits to exit 1, except under `-v`/`-L`
- each pattern pre-compiles standalone, so a malformed one is reported rather than silently absorbing its neighbour across the `|` separator
- `maxArrayElements` now bounds `-e`-expanded patterns as well as `-f` ones

**`regex.ts` anchor fix — one copy.** Byte-identical on both sides; kept #327's text because it has the explanatory comment.

**`skips.ts` — union of both removals.** All 9 busybox-grep cases unskipped by either branch pass in the merged tree:

```
✓ grep handles multiple regexps                                    (#314)
✓ grep -F handles multiple expessions                              (#314)
✓ grep -x -v -e EXP1 -e EXP2 finds nothing if either EXP matches   (#314)
✓ grep can read regexps from stdin                                 (#327)
✓ grep -f EMPTY_FILE                                               (#327)
✓ grep -v -f EMPTY_FILE                                            (#327)
✓ grep -vxf EMPTY_FILE                                             (#327)
✓ grep PATTERN can be a newline-delimited list                     (#327)
✓ grep -e PATTERN can be a newline-delimited list                  (#327)
```

## Test expectations adjusted

**None.** #314's four `grep.multiple-patterns.test.ts` cases and #327's three patterns-from-file files all pass unmodified. No behaviour asserted by one PR conflicts with the other.

For the record, all four of @tylergibbs1's assertions were re-checked against GNU grep 3.12 and are exactly right:

```
$ grep --version | head -1
grep (GNU grep) 3.12

$ grep -e Q1 -e Q4 test1.txt          $ grep -F -e 'a.b' -e 'x*y' test2.txt
aaa Q1                                a.b
ccc Q4                                x*y

$ grep -x -e foo -e bar test3.txt     $ grep one.txt -e needle two.txt
foo                                   one.txt:needle one
bar                                   two.txt:needle two
```

## Verification

Ground truth throughout is GNU grep 3.12 (`/opt/homebrew/opt/grep/libexec/gnubin/grep`), never BSD grep.

**Differential vs GNU grep 3.12 — 65/69 cases match exactly** (stdout, stderr, exit code), across two rounds:

- round 1, 42/46: repeated `-e` in every mode (`-F -E -P -x -w -i -v -c -n -q -l`), repeated `-e` + `-f` together, empty pattern file with `-c`/`-v`/`-L`, `-f -`, malformed-pattern rejection, `-P` guards, newline-separated PATTERNS
- round 2, 23/23: `-e` after a file operand, empty `-e`, `-P` duplicate folding via `-f`, `-fFILE`/`-vxfFILE` attached forms, `-m`, `-o`, `--` terminator, multi-file `-l`

The 4 non-matching cases are **pre-existing gaps, not merge regressions** — each reproduces identically with a *single* pattern on the unmerged baseline, on code paths this merge never touches:

| Case | GNU | just-bash | Note |
|---|---|---|---|
| `-L` exit code | 1 | 0 | reproduces with `grep -L zzz f`; `fix/grep-L-exit-code` branch already exists |
| BRE `\+` extension | matches | no match | reproduces with `grep 'a\+pple' f`; GNU BRE extension unsupported |
| malformed `-E` message ×2 | `Unmatched ( or \(` | `invalid regular expression: a(` | wording only; both exit 2 and reject. Pre-existing house style |

**Suites on the merged tree:**

| Check | Result |
|---|---|
| `src/commands/grep/` (both PRs' files) | 11 files, **276 passed**, 8 skipped |
| `src/commands/search-engine/` + `src/commands/rg/` | 24 files, **591 passed**, 82 skipped |
| `src/spec-tests/` | 5 files, **3825 passed**, 1 skipped |
| `pnpm test:comparison` | 36 files, **593 passed** |
| `pnpm test:run` (full) | **14382 passed**, 97 skipped, 6 failed |
| `pnpm typecheck` | pass |
| `pnpm lint:fix` | clean, no fixes applied |
| `pnpm lint:banned` | no banned patterns |
| `pnpm knip` | clean |

The 6 full-suite failures are all python3/CPython-WASM and sqlite worker tests, unrelated to grep. Verified pre-existing: checking out #327's tip unmerged and running the same files reproduces 5 of them identically (the 6th is `just-bash.bundle.test.ts > should lazy-load commands (python3 with CPython Emscripten)`, same family). Looks environmental to the sandbox this was built in.

## About the red CI on this PR

**Only one job actually failed a test, and it is a known Node-24 flake unrelated to grep.** The matrix looks like "all three Node versions failed", but that is `fail-fast` masking:

| Job | What the log actually says |
|---|---|
| `unit-tests (24)` | **1 failed** — `js-exec-host-runtime-breakout-probes.test.ts` |
| `unit-tests (22)` | **`496 passed`** — every unit file, plus executor and dist tests — then `The operation was canceled.` |
| `unit-tests (20)` | **`478 passed`** and still running when `The operation was canceled.` |

`.github/workflows/unit-tests.yml` sets no `fail-fast: false`, so the Node 24 failure cancelled the other two mid-flight. **Node 22 passed the entire unit suite on this exact tree.**

Why it is not the merge:

1. **No causal path.** `git diff 861f201 c31307f` is 4 files: `grep.ts`, `skips.ts`, `grep.multiple-patterns.test.ts`, one changeset. Nothing under `src/security`, `src/commands/js-exec`, or the worker/bridge. The failing test does not use grep.
2. **The signature rules out resource starvation.** The failure is truncated stdout with exit 0 and empty stderr. A js-exec deadline cannot produce that — `js-exec.ts:687-695` returns `exitCode: 124` with a `js-exec: ...` message on stderr when the worker is cancelled. Truncation with exit 0 is the bridge losing buffered `console.log` output as the worker completes, i.e. an output-draining race.
3. **Load does not reproduce it.** 25 CPU-saturating hogs on 16 cores, 5 consecutive runs of that test file: 5/5 pass.
4. **This suite already flakes on main.** 5 of the last 25 `Unit Tests` runs on `main` itself failed — including on `6df692f`, this PR's own merge base (that one on `state-transaction.test.ts`).
5. **Same test, same Node version, same shape as PR #330's flake**, which was investigated and proven Node-24-only.

Worth doing separately: adding `fail-fast: false` to the unit-tests matrix would stop one Node's flake from cancelling the other two and reporting as a three-version failure. Deliberately not included here — it is a CI change with nothing to do with grep.

## Risks

- **Pattern ordering.** Patterns are collected as all `-e` first, then all `-f`, rather than strictly interleaved in argument order. Unobservable today — order within an alternation cannot change which lines are selected, and `-o` matched GNU in testing — but it would matter if leftmost-alternative semantics ever became visible (e.g. `--color`).
- **`-P` refusal is a deliberate divergence from silence, not from GNU.** GNU also refuses multiple patterns under `-P`; the guard now sees `-e`-collected patterns, so `-P -e a -e b` errors where before the merge it would have silently used only `b`. This is the GNU-correct behaviour but it is a behaviour change for anyone relying on the old last-one-wins.
- Nothing here is independently reviewed work — it is two already-reviewed PRs plus a mechanical graft. The graft itself (repeated `-e` feeding #327's guarded collection path) is the only genuinely new code and is where review attention is best spent.

### DIFF
--- a/.changeset/quick-geckos-grep.md
+++ b/.changeset/quick-geckos-grep.md
@@ -1,0 +1,5 @@
+---
+"just-bash": patch
+---
+
+Combine repeated `grep -e` patterns and correctly group alternatives in line-regexp modes.
--- a/.changeset/witty-moons-shave.md
+++ b/.changeset/witty-moons-shave.md
@@ -1,0 +1,5 @@
+---
+"just-bash": minor
+---
+
+Add `grep -f FILE` / `--file=FILE` to read patterns from a file (one per line). Patterns from `-f` OR-combine with `-e` patterns and with each other, `-f -` reads patterns from stdin, empty pattern lines match every line, and an empty pattern file selects nothing (exit 1). Newline-separated `PATTERNS` operands are now split into individual patterns, and `-x` groups alternatives correctly (`^(?:a|b)$`).
--- a/packages/just-bash/src/commands/grep/grep.multiple-patterns.test.ts
+++ b/packages/just-bash/src/commands/grep/grep.multiple-patterns.test.ts
@@ -26,6 +26,18 @@ describe("grep with multiple -e patterns", () => {
     expect(result.exitCode).toBe(0);
   });
 
+  it("anchors every pattern with -x", async () => {
+    const env = new Bash({
+      files: { "/test.txt": "foo\nbar\nfoobar\nfoo suffix\nprefix bar\n" },
+    });
+
+    const result = await env.exec("grep -x -e foo -e bar /test.txt");
+
+    expect(result.stdout).toBe("foo\nbar\n");
+    expect(result.stderr).toBe("");
+    expect(result.exitCode).toBe(0);
+  });
+
   it("treats non-option operands as files when -e is present", async () => {
     const env = new Bash({
       files: {
--- a/packages/just-bash/src/commands/grep/grep.multiple-patterns.test.ts
+++ b/packages/just-bash/src/commands/grep/grep.multiple-patterns.test.ts
@@ -1,0 +1,43 @@
+import { describe, expect, it } from "vitest";
+import { Bash } from "../../Bash.js";
+
+describe("grep with multiple -e patterns", () => {
+  it("matches lines selected by any pattern", async () => {
+    const env = new Bash({
+      files: { "/test.txt": "aaa Q1\nbbb\nccc Q4\n" },
+    });
+
+    const result = await env.exec("grep -e Q1 -e Q4 /test.txt");
+
+    expect(result.stdout).toBe("aaa Q1\nccc Q4\n");
+    expect(result.stderr).toBe("");
+    expect(result.exitCode).toBe(0);
+  });
+
+  it("keeps each pattern literal with -F", async () => {
+    const env = new Bash({
+      files: { "/test.txt": "a.b\naxb\nx*y\nxy\n" },
+    });
+
+    const result = await env.exec("grep -F -e 'a.b' -e 'x*y' /test.txt");
+
+    expect(result.stdout).toBe("a.b\nx*y\n");
+    expect(result.stderr).toBe("");
+    expect(result.exitCode).toBe(0);
+  });
+
+  it("treats non-option operands as files when -e is present", async () => {
+    const env = new Bash({
+      files: {
+        "/one.txt": "needle one\n",
+        "/two.txt": "needle two\n",
+      },
+    });
+
+    const result = await env.exec("grep /one.txt -e needle /two.txt");
+
+    expect(result.stdout).toBe("/one.txt:needle one\n/two.txt:needle two\n");
+    expect(result.stderr).toBe("");
+    expect(result.exitCode).toBe(0);
+  });
+});
--- a/packages/just-bash/src/commands/grep/grep.patterns-from-file-combos.test.ts
+++ b/packages/just-bash/src/commands/grep/grep.patterns-from-file-combos.test.ts
@@ -1,0 +1,221 @@
+import { describe, expect, it } from "vitest";
+import { Bash } from "../../Bash.js";
+
+/**
+ * `grep -f FILE` composed with the other selection flags. The `-F -x -f` and
+ * `-v -f` combinations are the set-intersection / set-difference idioms called
+ * out in issue #322. Expectations verified against GNU grep 3.12.
+ */
+const setFiles = {
+  "/list.txt": "apple\nbanana\ncherry\n",
+  "/keep.txt": "banana\ndate\napple\n",
+};
+
+const searchFiles = {
+  "/pat.txt": "apple\nbanana\n",
+  "/p1.txt": "apple\n",
+  "/hay.txt": "apple pie\ncherry\nbanana split\n",
+  "/hay2.txt": "x\ny\n",
+};
+
+describe("grep -f set operations", () => {
+  it("intersects two line sets with -F -x -f", async () => {
+    const env = new Bash({ files: setFiles });
+    const result = await env.exec("grep -F -x -f /keep.txt /list.txt");
+    expect(result.stdout).toBe("apple\nbanana\n");
+    expect(result.stderr).toBe("");
+    expect(result.exitCode).toBe(0);
+  });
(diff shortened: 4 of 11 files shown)


trieloff:
no longer needed, closing.

PR #335, opened 2026-07-30
combo: numeric file descriptors (#326) + group stdin ownership (#330), pre-resolved

trieloff:
# combo: numeric file descriptors (#326) + group stdin ownership (#330), pre-resolved

## What this is

A convenience branch, not a third feature. It is **#326 and #330, already merged**, so
whoever lands them does not have to resolve the conflict by hand.

- `combo/numeric-fds-group-stdin` = `feat/numeric-fds-321` + `fix/group-stdin-restore`
- exactly one merge commit; both PR histories are intact as its parents
- **nothing here is new work.** Every line comes from one of the two PRs except the
  conflict resolution described below.

**Close this freely.** If you would rather merge #326 and #330 individually, do that —
this branch exists only so you do not have to do the merge yourself. Review the two PRs;
this page only argues that they compose.

### Why they conflict

Both thread new state through the *same* call chain:

```
executeSimpleCommandInner → runCommand → dispatchBuiltin → callFunction / handleEval
executeCommand            → executeGroup / executeSubshell
```

\#326 adds a descriptor table and routes every fd ≥ 3 construct through it; #330 adds a
`stdinRedirected` / `stdinOwned` ownership flag beside `stdin`. Both land as optional
trailing parameters on the same functions, and both rewrite the same redirection loop in
`executeSimpleCommandInner`. Git leaves five conflicts in `interpreter.ts` and — worse —
**auto-merges `subshell-group.ts` into something that typechecks nowhere and would have
silently dropped ownership if it had.** See "the trap" below.

### Parameter order

\#326 adds no parameters to `runCommand`, `dispatchBuiltin` or `callFunction`, so #330's
flag keeps its place as the trailing parameter after `stdinSourceFd` on every one of
them. Nothing had to be renumbered:

| function | signature after the merge |
| --- | --- |
| `runCommand` | `(name, args, quotedArgs, stdin, skipFunctions, useDefaultPath, stdinSourceFd, stdinRedirected)` |
| `dispatchBuiltin` | `(dispatchCtx, name, args, quotedArgs, stdin, skipFunctions, useDefaultPath, stdinSourceFd, stdinRedirected)` |
| `callFunction` | `(ctx, func, args, stdin, callLine, stdinRedirected)` |
| `executeCommand` | `(node, stdin, stdinOwned)` |
| `executeGroup` / `executeSubshell` | `(ctx, node, stdin, executeStatement, stdinOwned)` |

Only #326's private `executeGroupBody` / `executeSubshellBody` take a sixth parameter,
`fdTargets`, after `stdinOwned` — the fd plumbing is internal to those two files, so it
stays behind the public flag rather than in front of it.

### The trap

\#326 splits `executeGroup` and `executeSubshell` into a `withNumericFds` wrapper plus a
`*Body` function. #330 adds `stdinOwned` to those same two exported signatures. Git
merges both edits cleanly: the wrapper grows the parameter, the body still uses it, and
the wrapper **never passes it**. That resolution is what a `--strategy=recursive` merge
produces. Here the flag is threaded to the body, which is where `ownsStdin` reads it.

The rest of the call sites were audited argument-by-argument rather than trusted to the
type checker — every caller of `runCommand` (5), `dispatchBuiltin` (1), `callFunction`
(2), `executeCommand` (3) and the group/subshell helpers (2+2). An optional boolean in
the wrong slot typechecks fine.

### Redirection loop in `executeSimpleCommandInner`

\#326's `isNumericFdRedirection(redir) continue` runs first, so a redirection naming fd ≥ 3
never touches stdin and never sets #330's flag. That is the correct composition: `cmd 3< f`
opens a descriptor, it does not give the command an fd 0.

The remaining fd-0 branches (`<<`, `<<<`, `<`) set both #326's `stdinSourceFd = -1` and
#330's `stdinRedirected = true`.

`<&N` needed a judgement call, because #326 rewrote that branch onto the typed table and
it now has six exits. Ownership is claimed on exactly two of them:

| exit | owns fd 0? | why |
| --- | --- | --- |
| `<&-` | **yes** | a closed fd 0 is EOF, not the shell's stdin |
| `readFd` success | **yes** | an exhausted descriptor hands over `""`, and that is EOF |
| `<&0` | no | explicit no-op, stdin untouched |
| `<&N` where N duplicates stdin (`exec 3<&0`) | no | reads the shell's stdin by construction |
| unopened standard descriptor | no | falls through with stdin unchanged |
| unopened user descriptor / write-only | n/a | returns `Bad file descriptor` |

That decision is load-bearing and verified against bash — see the matrix below.

### `callFunction`

\#326 makes `processInputRedirections` skip numeric descriptors; #330 makes it report
`{ stdin, redirected }`. Composed, a definition-level `f() { …; } 3< file` reports
`redirected: false` — correct, it opens a descriptor without giving the body fd 0 — while
`f() { …; } < file` still reports `true` and shields the body from the caller's stdin.

## Bonus: two cases neither PR fixes alone

\#330's *Known residual divergences* listed "a group's `<` branch ignores the descriptor
number" and handed it to #326. The merged tree fixes it for free, because #326 removes
the redirection from the stdin loop and #330 stops the group rewinding the shared
position:

```bash
{ read a; { read x; } 3< other.txt; read b; }  < loop.txt
#  bash:   x=[L2]  a=[L1] b=[L3]
#  #326:   x=[L2]  a=[L1] b=[L2]      ← fd 3 bound, but the group rewinds
#  #330:   x=[O1]  a=[L1] b=[L2]      ← no rewind, but other.txt fed stdin
#  merged: x=[L2]  a=[L1] b=[L3]      ✓
```

Same shape with an output descriptor, also only correct on the merge:

```bash
{ read a; { read x; echo "seen:$x" >&4; } 4> out.txt 3< other.txt; read b; }  < loop.txt
#  bash / merged: a=[L1] b=[L3]     #326 alone: a=[L1] b=[L2]
```

`{ read -u 3 x; } 3< other.txt` is correct on #326 and on the merge, and wrong on #330
alone (`x=[]`), so the merge is not a regression in the other direction either.

## Verification

Everything below was run on the merge commit.

### Suites

| suite | result |
| --- | --- |
| the 10 unit files from both PRs (7 fd + 3 group-stdin) | **168 passed, 0 failed** (#326's 101 + #330's 67) |
| `pnpm test:comparison` | **612 passed, 37 files** (567 base + 20 from #326 + 25 from #330) |
| `RECORD_FIXTURES=1` on both new fixture files | re-records with **no diff** |
| `src/spec-tests` | **3825 passed, 1 skipped, 0 failed** — identical to both branches |
| `pnpm test:run --exclude src/spec-tests` | 10675 passed, **6 failed**, 96 skipped (10777) |
| `packages/just-bash-executor` `pnpm test:unit` | 72 passed — and **0 failures in 20 consecutive runs** |
| `pnpm typecheck` (both packages) | clean |
| `pnpm lint` (workflow security + biome + `lint:banned`) | clean |
| `pnpm knip` | clean |

The 6 failures are the CPython-WASM suite both PRs already report:
`vendor/cpython-emscripten/python.wasm` is an unmaterialised git-LFS pointer in this
checkout, so `python: true` tests time out loading the worker. Same six files, same
reason, on both parent branches.

\#330 reported one CI flake in `just-bash-executor` on Node 24. 20 repeat runs on the
merge produced zero failures, so the merge does not make it deterministic-bad.

### Bash matrix

58 shapes, each run in `/bin/bash` 3.2.57 and in just-bash via `pnpm dev:exec --real-bash`:
**47 match, 11 differ.** Every one of the 11 produces byte-identical just-bash output on a
parent branch — nothing is introduced here. The 11 group into six known items, five of
them named in one PR body or the other:

- **`read`'s empty-stdin fallback** (3 shapes) — #330's documented open item: `handleRead`
  falls back to `groupStdin` when its own stdin is empty, so `read b <&-` and `read b <&3`
  on an exhausted descriptor still see the shell's line. The ownership fact now *reaches*
  `dispatchBuiltin`; wiring it into `read` is the follow-up #330 already scopes.
- **`exec 3<&0`** binds "the current stdin" rather than the stdin open at `exec` time —
  \#326's documented approximation.
- **`( … ) < file` never installs the file as stdin** — #330's documented gap.
- **`{ …; } <&3`** — `executeGroup` has no `<&` branch on any branch; the group reads its
  inherited stdin.
- **`command read x <&3`** — the wrapper drops `stdinSourceFd`, so fd 3 is drained rather
  than advanced by one line. Both PRs name this: #330 called it "the pre-existing
  `<&N`-through-a-wrapper gap [that] belongs with #326's descriptor work", and #326 does
  not close it. **`command read -u 3 x` and `builtin read -u 3 x` are correct** — `-u`
  reads the table directly and `read` advances it itself — so only the `<&N` spelling is
  affected.
- **`bash: line N:` prefix** on two `read`-on-a-closed-descriptor diagnostics; stdout and
  exit code match. This is the divergence #326 locks three fixtures for.

Shapes that pass on the merge and are the point of the exercise:

```
#326  while read -u 3 l; do …; done 3< fd.txt        exec 3< f; read -u 3        read <&3
      exec 3< f; exec 4<&3; exec 5<&4  (one offset)  scoped 4<&3                 done 3< g scoping
#330  { { read a; }; read b; } < file → b=[L2]       { eval 'while read l…'; } < file
      f < empty / eval '…' < empty / { …; } < empty  5 loop-shield idioms × 5 turns
      command eval / builtin eval, shared and empty-redirected
combo { read -u 3 a; } sharing vs owning stdin       f 3< file, f() { …; } 3< file
      eval 'read -u 3 …' inside a group              exec 3< f; { exec 4<&3; read -u 4; }
      f <&3 / eval '…' <&3 on an exhausted fd → EOF, not the shell's line
```

The last row is the `<&N` ownership decision paying off: on #326 alone `f <&3` with fd 3 at
EOF reads the enclosing shell's line (`v=[L2]`); on the merge it sees EOF (`v=[]`), which
is what bash does.

## Risks

- **Concentrated in one file.** Four of the five textual conflicts were one-line additions
  to `interpreter.ts`; the fifth was the `<&` rewrite, and the ownership decision there is
  the only place where a *choice* was made rather than a union taken. It is covered by six
  purpose-built shapes above, all matching bash.
- **`subshell-group.ts` auto-merged wrong.** If you re-do this merge yourself, that is the
  spot to check: the wrapper must forward `stdinOwned` to the body.
- **No new tests.** The 213 tests (168 unit + 45 comparison) are exactly the two PRs' own. The composition cases in
  the matrix are verified but not pinned; if both PRs land, the `{ …; } 3< file` bonus
  deserves a test in a follow-up.
- **Not rebased on `main`.** Both branches are cut from `1e06ce6`; the merge is too. If
  `main` has moved, the result needs the same rebase either PR would.


trieloff:
Superseded by #336, which carries #326 + #329 + #330 in one pre-resolved branch.

#326 conflicts with **both** #329 and #330, so this combo and its sibling were mutually exclusive — merging either one stranded the other. #336 replaces both. Closing to keep the queue unambiguous; every finding documented here is carried forward in #336's description.

PR #338, opened 2026-08-06
fix(sqlite3): report SQL errors on stderr and exit 1 without -bail

mutewinter:
## Problem

A failed statement is written to **stdout** as `Error: ...` and the command exits `0`. Real `sqlite3` writes it to stderr and exits `1`.

```bash
sqlite3 data.db "SELECT * FROM nonexistent" > rows.csv && echo ok

# before: rows.csv contains "Error: no such table: nonexistent", prints "ok", $? = 0
# after:  rows.csv is empty, error on stderr, no "ok", $? = 1
```

`-bail` is not the switch for this. In real `sqlite3` it decides whether the remaining statements still run; the exit status is `1` either way:

```bash
sqlite3 real.db "SELECT * FROM nope"          # $? = 1, stderr: Error: in prepare, no such table: nope
sqlite3 -bail real.db "SELECT * FROM nope"    # $? = 1
```

(checked against sqlite3 3.51.0)

This lands hardest on an agent. `cmd && next-step` is how a model sequences work, and a query is the one step whose failure it cannot see: the command succeeded, so it proceeds, and the error text is now sitting in the middle of the data it just "collected". A redirect turns it into a corrupt file, a pipe into `jq` turns it into a parse error attributed to the wrong command. The only way to detect a bad query today is to pass `-bail` — which also gives up every statement after the first failure, so correct error handling and running a multi-statement script are mutually exclusive.

## Cause

Two branches in the same loop. The `-bail` path set `bailError` and returned it on stderr with exit `1`; the non-bail path appended `Error: ...` to `stdout` and left the status to `hadError && options.bail`, which is only ever reachable when `bail` is false. So the non-bail case could not report a failure through either channel.

## Fix

Collect errors on a `stderr` string in statement order, and set the exit status from `hadError` alone. `-bail` keeps its meaning as `break`.

That removes the `bailError` variable and its return branch — the `break` plus the shared exit status already covers it.

The two early returns that were already writing to stderr (format failure, writeback failure) now prepend what the loop collected, so a SQL error is no longer dropped when something fails after it.

## Scope

Unchanged: successful runs, `-bail` stopping at the first failure, the partial stdout produced before a failure, and writeback of a partially-successful script.

Not addressed here — real `sqlite3` continues past an error when statements arrive on stdin but stops at the first one when they arrive as an argv string. This command continues in both. That's a separate change to statement dispatch and it doesn't affect the exit status either way.

Also not addressed: dot commands. `.tables` and `.schema` are the first thing a model reaches for and they're still a parse error, but #215 and #249 were closed, so I've read that as deliberate rather than missing.

## Tests

Five expectations updated and two cases added, in `sqlite3.errors.test.ts`, `sqlite3.test.ts`, and `sqlite3.write-ops.test.ts`. The two new ones assert the consequences rather than the mechanism: the error staying out of a redirected stdout, and `&&` short-circuiting. The two `load_extension` security assertions move from stdout to stderr.

The updated ones were pinning the old behaviour explicitly, e.g. `should continue after error and return exit code 0` asserting `Error:` on stdout — worth a look, since flipping them is the whole behavioural surface of this PR.

`sqlite3` suite: 163 passing, 4 failing. The same 4 fail on `main` — both statement-splitting cases (trigger bodies, bracket identifiers), the row-limit case, and `incremental_vacuum` writeback. Verified by stashing this change and re-running, not inferred. Nothing new fails.

Two notes on what I could not verify, rather than leaving them implied:

- `main` doesn't typecheck at `d97425d` — `stdinOwned` is used at `interpreter.ts:599,601` but is a parameter of a different function. So `pnpm build` fails and the bundle tests can't run. `tsc` reports those two errors and nothing in `sqlite3`.
- The two failing statement-splitting cases are worse than they look: both produce empty stdout and a cascade of parse errors, so the splitter is mangling trigger bodies and `[a;b]` identifiers today. They were passing before this change for the same reason this PR exists — the errors went to stdout and the status stayed `0`. Untouched here, but they probably want their own issue.

Biome clean, changeset included.

---

Model: Claude Opus 5

### DIFF
--- a/.changeset/sqlite3-error-exit-status.md
+++ b/.changeset/sqlite3-error-exit-status.md
@@ -1,0 +1,11 @@
+---
+"just-bash": patch
+---
+
+sqlite3: report failed statements on stderr and exit non-zero without `-bail`
+
+A statement that failed was written to **stdout** as `Error: ...` and the command still exited `0` unless `-bail` was passed. Real `sqlite3` writes the error to stderr and exits `1` in either mode; `-bail` only decides whether the remaining statements still run.
+
+Two consequences for callers: `sqlite3 db "SELECT ..." > out.csv` silently wrote the error text into the data file, and `sqlite3 db "..." && next-step` ran `next-step` after the query had failed, so a shell script could not detect a bad query without opting into `-bail` and losing the ability to see later statements.
+
+Errors now accumulate on stderr in statement order and the exit status is `1` whenever any statement failed. Successful runs are unchanged, `-bail` still stops at the first failure, and the partial stdout produced before a failure is still emitted. Writeback of a partially-successful script is also unchanged.
--- a/packages/just-bash/src/commands/sqlite3/sqlite3.errors.test.ts
+++ b/packages/just-bash/src/commands/sqlite3/sqlite3.errors.test.ts
@@ -55,15 +55,14 @@ describe("sqlite3 error handling", () => {
   });
 
   describe("SQL errors without -bail", () => {
-    it("should continue after error and return exit code 0", async () => {
+    it("should continue after an error, report it on stderr, and exit 1", async () => {
       const env = new Bash();
       const result = await env.exec(
         'sqlite3 :memory: "SELECT * FROM nonexistent; SELECT 42"',
       );
-      expect(result.stdout).toContain("Error:");
-      expect(result.stdout).toContain("no such table");
-      expect(result.stdout).toContain("42");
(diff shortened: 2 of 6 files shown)


cramforce:
Please fix the build

cramforce:
Need to rebase after #343

mutewinter:
Build passing.

cramforce (approved):

PR #346, opened 2026-08-08
fix: lower browser buffer conversion chunk size

### LINKED ISSUE #282: fromBuffer chunk size (65536) exceeds Chrome Service Worker String.fromCharCode argument limit
## Follow-up to #69

The chunked `fromBuffer` fix in commit 0d87a8f chose `chunkSize = 65536`. This is still too large for Chrome Service Worker environments, where the effective `String.fromCharCode` argument limit is ~63,600 (due to deeper call stacks reducing available stack space).

Both code paths in the binary/latin1 branch crash:

```javascript
// Path 1: buffer ≤ 65536 — spread all at once, crashes at ~63600 args
if (buffer.length <= chunkSize) {
    return String.fromCharCode(...buffer);
}
// Path 2: buffer > 65536 — each chunk is still 65536, also crashes
for (let i = 0; i < buffer.length; i += chunkSize) {
    const chunk = buffer.subarray(i, i + chunkSize);
    result += String.fromCharCode(...chunk);
}
```

**Result:** `cat`, `grep`, `head`, `tail`, `sed`, `awk` and any command that reads files via the binary encoding path fail with a swallowed `RangeError` for files ≥ ~63KB. The error surfaces as `"No such file or directory"` because the `RangeError` is caught by the command's generic error handler.

## Reproduction

```javascript
// In a Chrome extension Service Worker:
const bash = new Bash({ files: { '/tmp/.keep': '' } });
await bash.fs.writeFile('/tmp/big.txt', 'X'.repeat(65536));
const r = await bash.exec('cat /tmp/big.txt');
// r.exitCode === 1, r.stderr === "cat: /tmp/big.txt: No such file or directory\n"
```

## Environment

- just-bash 3.0.1
- Chrome 137, macOS
- Running inside Manifest V3 extension Service Worker
- Measured SW argument limit: `String.fromCharCode(...new Uint8Array(63593))` OK, `63671` crashes

## Suggested fix

Lower `chunkSize` from `65536` to a safe value (e.g., `8192`). The V8 argument limit is not fixed — it depends on call stack depth ([well-documented across the ecosystem](https://github.com/nicolo-ribaudo/tc39-proposal-max-function-arguments/issues/1)). 8192 is conservative and has negligible performance impact.

iroiro147:
## Summary

Lower the browser-safe conversion chunk size in both binary/latin1 conversion paths from 65536 to 8192.

This prevents String.fromCharCode(...buffer) from exceeding the effective argument limit in Chrome Service Worker environments.

Fixes #282

## Verification

local verification not run; relying on upstream CI

The fresh worktree has no node_modules installed, so the repository scripts cannot execute locally.

### DIFF
--- a/packages/just-bash/src/fs/encoding.ts
+++ b/packages/just-bash/src/fs/encoding.ts
@@ -37,7 +37,7 @@ export function toBuffer(
   }
   if (encoding === "binary" || encoding === "latin1") {
     // Use chunked approach for large strings to avoid performance issues
-    const chunkSize = 65536; // 64KB chunks
+    const chunkSize = 8192; // 8KB chunks
     if (content.length <= chunkSize) {
       return Uint8Array.from(content, (c) => c.charCodeAt(0));
     }
@@ -65,7 +65,7 @@ export function fromBuffer(
     if (typeof Buffer !== "undefined") {
       return Buffer.from(buffer).toString("base64");
     }
-    const chunkSize = 65536;
+    const chunkSize = 8192;
     let binary = "";
     for (let i = 0; i < buffer.length; i += chunkSize) {
       const chunk = buffer.subarray(i, i + chunkSize);
@@ -85,7 +85,7 @@ export function fromBuffer(
     }
 
     // Browser fallback - String.fromCharCode(...buffer) fails with buffers > ~100KB
-    const chunkSize = 65536; // 64KB chunks
+    const chunkSize = 8192; // 8KB chunks
     if (buffer.length <= chunkSize) {
       return String.fromCharCode(...buffer);
     }


cramforce (approved):

cramforce:
Please rebase and sign commits

iroiro147:
Done — rebased onto latest `main` (was 6 behind) and the commit is now SSH-signed. Head is `8bdeae8`. The force-push dismissed the approval, sorry about that — re-review whenever suits you.

iroiro147:
Friendly ping — this one is approved and still open.

Both asks from the review are done: rebased onto `main` and the commit is
SSH-signed. Head `8bdeae8` is the commit your approval sits on, and it still
merges cleanly.

The only red checks are `Vercel – just-bash` and `Vercel – just-bash-website`,
both reporting *"Authorization required to deploy"* — that gate needs someone on
the Vercel Labs team, so there is nothing further I can clear from my side.
Happy to rebase again if it has drifted.

packages/just-bash/src/fs/encoding.ts
```diff
@@ -37,7 +37,7 @@ export function toBuffer(
   }
   if (encoding === "binary" || encoding === "latin1") {
     // Use chunked approach for large strings to avoid performance issues
-    const chunkSize = 65536; // 64KB chunks
+    const chunkSize = 8192; // 8KB chunks
```
privatenumber: Could we leave this threshold unchanged? `toBuffer` uses `Uint8Array.from` or an indexed loop, so neither path spreads the bytes into function arguments. Lowering it changes which conversion algorithm handles 8-64KB strings without addressing the reported crash. The two `fromBuffer` changes below cover the affected paths.

packages/just-bash/src/fs/encoding.ts
```diff
@@ -85,7 +85,7 @@ export function fromBuffer(
     }
 
     // Browser fallback - String.fromCharCode(...buffer) fails with buffers > ~100KB
-    const chunkSize = 65536; // 64KB chunks
+    const chunkSize = 8192; // 8KB chunks
```
privatenumber: Could we add regression coverage with `Buffer` unavailable for both this path and the base64 fallback? Normal Node tests take the `Buffer` fast path and would miss the reported failure.

It would help to check exact output across the chunk boundaries, including bytes above `0x7f`, and confirm the original 64KB `cat` reproduction in a Chrome extension service worker. That would establish that 8192 fixes the affected environment; the issue reports that the effective argument limit depends on available stack space.

privatenumber (changes requested):
The two `fromBuffer` changes look right for the reported crash. Requesting changes to drop the unrelated `toBuffer` edit and to add browser-fallback coverage, so we can confirm the fix in the affected service worker.

PR #348, opened 2026-08-09
fix(parser): fail closed on unsupported command syntax

privatenumber:
## Problem

Recognized but unsupported command syntax could be silently deleted or reinterpreted. For example, command-leading `select` and `coproc` tokens were skipped, allowing a trailing payload to execute under a different meaning. Unknown command AST nodes also returned success instead of failing.

GNU Bash recognizes [reserved words only in specific grammar positions](https://www.gnu.org/software/bash/manual/html_node/Reserved-Words.html), and its [pipeline grammar](https://www.gnu.org/software/bash/manual/html_node/Pipelines.html) treats `time` and `!` as pipeline prefixes. Input outside the implemented grammar must fail before execution rather than degrade into another command.

## Changes

- Uses one reserved-token classification for command-start and command-dispatch decisions.
- Rejects unhandled command-leading reserved words instead of force-advancing past them.
- Rejects bare `in` and post-pipe `!` while keeping their quoted or escaped forms as ordinary words, and preserves valid `time`/`!` prefix forms without collapsing a second `time` or treating `--` as a timing option.
- Replaces the interpreter's unknown-command success fallback with an exhaustive failure.
- Keeps reserved words valid as quoted words and as arguments after an actual command.


cramforce (approved):

cramforce:
Why would it not have run CI?

privatenumber:
CI passing now! Merging

PR #349, opened 2026-08-09
fix(interpreter): process redirections transactionally

privatenumber:
## Problem

Redirections could be expanded, opened, and delivered in separate interpreter phases. Scripts therefore diverged from Bash when behavior depended on source order or descriptor lifetime: an earlier target could escape a later failure without its expected side effect, duplicated descriptors could lose their shared file position, and output produced before `return`, `exit`, or `errexit` could bypass a compound command's redirects.

For example, Bash processes `cmd >first <missing` from left to right: `first` is created or truncated before opening `missing` fails. Treating all targets as one prevalidated batch preserves the old contents instead, which is observably different for scripts that inspect the filesystem after failure.

## Changes

Redirections now run through one transaction that expands and opens each redirect exactly once in source order. Simple and compound commands restore command-scoped descriptors when they finish, while `exec` retains its descriptor routes and descriptor variables. Bare redirection commands restore descriptor variables after use.

Prepared descriptor routes preserve duplicate aliases, shared read/write positions, fd 0 ownership, and output accounting through `return`, `exit`, and `errexit`. Invalid fd 1 writes retain the actual producer name and Bash's distinct `exec` diagnostic format.

This follows Bash's documented left-to-right redirection order and descriptor duplication semantics: https://www.gnu.org/software/bash/manual/html_node/Redirections.html


cramforce (approved):

cramforce:
See vercel bot comment

PR #362, opened 2026-08-11
Fix arithmetic command substitution semantics

privatenumber:
## Problem

Arithmetic expressions accepted command substitution, but ran raw command text through a fresh executor and coerced its output with `parseInt`. A script such as `$(printf '1 + 2') * 3` therefore produced `3` instead of `7`; functions and the working directory from the current script were also unavailable. The node-by-node evaluator skipped a substitution behind `0 && ...`, although Bash completes command substitutions before arithmetic short-circuiting.

The parsed path also had two correctness gaps. A substitution inside `${...}`, such as an array subscript, could be paired with the wrong top-level substitution and fail the expression. Commands run by a substitution could also leave `set` and `shopt` changes in the caller, so `$(set -u; printf 1)` made a later unset arithmetic variable fail.

## Changes

Arithmetic nodes retain their parsed substitution body while preserving the existing raw `command` field for AST consumers. The evaluator executes that body through regular command-substitution isolation, inserts its output as arithmetic text, then reparses once so precedence matches Bash without executing shell syntax generated by the output. C-style loop fields and normal substring offsets use the same expression path.

The authored-substitution scan now skips parameter expansions, so `${arr[$(printf 0)]}` remains opaque until its normal expansion phase. Command substitutions use the shared isolated shell-state transaction, keeping functions, variables, shell options, `shopt` settings, and attributes local while preserving the substitution result and status for the caller. A GNU Bash fixture covers both regressions alongside substitution ordering, non-reexecution, active function and variable behavior, and AST compatibility.


cramforce:
@auto-maintain review

PR #363, opened 2026-08-11
fix(ls): match GNU and BSD on operand handling, -t and type indicators

mutewinter:
## Problem

Four ways `ls` disagrees with the `ls` on the machine running it. Every one of them is silent: correct exit status, no diagnostic, wrong answer.

**1. A file that exists is reported missing, if its name contains `[`.**

```bash
ls -l 'report [1].pdf'
# ls: report [1].pdf: No such file or directory     $? = 2
```

The file is there, and nothing else in the sandbox has trouble with it:

```bash
cat 'report [1].pdf'              # body
stat -c '%s %n' 'report [1].pdf'  # 4 report [1].pdf
wc -c 'report [1].pdf'            # 4 report [1].pdf
rg --files                        # ./report [1].pdf
```

Names holding `?` or `*` do list, but come back respelled with their leading directories dropped:

```bash
ls -l /w/'q?mark.txt'
# -rw-r--r-- 1 user user 4 Aug 11 08:59 q?mark.txt      <- operand was /w/q?mark.txt
```

**2. A blank line is inserted between every operand, including plain files.**

```bash
ls -l a.txt b.txt
# -rw-r--r-- 1 user user 2 Aug 11 08:59 a.txt
#
# -rw-r--r-- 1 user user 3 Aug 11 08:59 b.txt
```

Nobody types that by hand, but `find … -exec ls -l {} +` and `xargs ls -l` produce exactly that shape, and they are how you ask how big everything under a directory is:

```bash
find . -type f -exec ls -l {} + | sort -k5 -n | head -3
#
#
# -rw-r--r-- 1 user user 2 Aug 11 08:59 ./a.txt
```

`sort -n` floats the blanks to the top and `head` takes them. Two thirds of that answer is empty lines, and the ratio gets worse with more files, because there is one blank per operand.

**3. `-t` does not sort by time.** It returns name order, and `--help` says otherwise.

```bash
touch -d 2020-01-01 big-old.txt && touch -d 2026-01-01 small-new.txt && touch -d 2023-01-01 mid.txt
ls -1t
# big-old.txt
# mid.txt
# small-new.txt        <- alphabetical; the newest file is last
```

**4. `-l` appends `/` to directories without `-F`,** so `ls` and `ls -l` disagree about what the same entry is called.

```bash
ls -1 /w        # dir1
ls -l /w        # drwxr-xr-x 1 user user 0 Aug 11 08:59 dir1/
```

The common thread is that each of these is most likely to be hit by a program rather than a person, and none of them announces itself. An agent asked which file is smallest gets a screen of blank lines; asked which is newest gets the alphabetically first name; asked about `report [1].pdf` is told by `ls` that it does not exist, and the reasonable next move is to report that to the user. Bracketed names are ordinary in the directories an agent gets pointed at: `file [1].pdf` from a browser that de-duplicated a download, exported mail carrying a `[ ref:… ]` tag.

## Cause

**Operands were matched as patterns.** `ls` ran every operand containing a metacharacter back through minimatch against the whole path table:

```ts
// Check if it's a glob pattern
if (path.includes("*") || path.includes("?") || path.includes("[")) {
  const result = await listGlob(path, ctx, ...);
```

Pathname expansion is the shell's job and has already finished by the time the command runs, so an operand still holding a metacharacter is either a real filename or a pattern that matched nothing and was passed through unchanged. Re-matching gets both wrong. For `report [1].pdf`, `[1]` reads as a character class, so the name cannot match itself, no candidate survives, and the empty-match branch reports the pattern as a missing file. For `q?mark.txt` the match succeeds, but `listGlob` pushes `relativePath`, computed against cwd, so the printed name is not the operand it was handed.

**The separator was applied before anything had been stat'ed.** One loop over `paths`, with this at the top:

```ts
// Add blank line between directory listings
if (i > 0 && stdout && !stdout.endsWith("\n\n")) {
  stdout = appendLsOutput(ctx, stdout, "\n");
}
```

The comment says directory listings. The code cannot tell, because nothing has looked at the operand yet at that point. Operand ordering came from the same place: the loop walked the command line, and `listPath` emitted whatever it was handed.

**`-t` was parsed and dropped.** `const _sortByTime = parsed.result.flags.sortByTime;` and nothing ever read it. `listPath` had one sort branch, `if (sortBySize) … else entries.sort()`. The underscore says this was known.

**The type indicator had its condition inverted,** at two sites, appending `/` in exactly the case where nothing was requested:

```ts
const suffix = classifyFiles
  ? classifySuffix(await ctx.fs.lstat(entryPath))
  : entryStat.isDirectory
    ? "/"
    : "";
```

Short format has no such branch, which is why the two formats disagreed.

Three of the four were pinned by the suite rather than caught by it. `find.exec.test.ts` asserted `"/dir/file1.txt\n\n/dir/file2.txt\n"` for `find /dir -type f -exec ls {} +`, encoding the blank line as expected output. `ls.test.ts` had a case named `should show directory indicator in long format`. There was no `-t` ordering test at all, so nothing noticed the flag did nothing.

## Fix

Operands resolve literally: the glob branch and `listGlob` are deleted, and every operand goes to `listPath`, which stats it.

Operands are charged to the traversal budget as they are resolved, then partitioned before anything is emitted: non-directories print first as one block in sort order with no label and no separator, then each directory prints under a `name:` label preceded by a blank line, and a lone directory operand stays unlabeled. `sortOperands` applies the same key the entries use, so `-S`, `-t` and `-r` reach operands as they do in GNU.

One comparator now serves name, size and time, and `-R` takes its section order from the same comparator, so `-t` and `-S` reach the descent rather than only each directory's own listing. Both size and time sort descending and fall back to the name on a tie, so a listing no longer depends on the order the filesystem returned entries in. `-S` and `-t` together are not an error and GNU lets whichever was written last decide, which cannot be read off the parsed booleans, so `resolveSortKey` scans the argument list for the last occurrence.

The type indicator comes from `classifySuffix` under `-F` and from nowhere otherwise.

Everything above was measured against GNU coreutils ls 9.2 and BSD ls (macOS 15), which agree on every case here:

| command | GNU 9.2 | BSD | before | after |
| --- | --- | --- | --- | --- |
| `ls 'report [1].pdf'` | lists it | lists it | `No such file or directory`, exit 2 | lists it |
| `ls -l /w/'q?mark.txt'` | `/w/q?mark.txt` | same | `q?mark.txt` | `/w/q?mark.txt` |
| `ls 'nope*'` | exit 2 | exit 1 | exit 2 | exit 2 |
| `ls -l a.txt b.txt` | no separator | no separator | blank line between | no separator |
| `ls -1 dir2 f2 f1 dir1` | `f1`, `f2`, then `dir1:`, `dir2:` | same | command-line order | files first |
| `ls -1t` | newest first | newest first | alphabetical | newest first |
| `ls -1St` / `ls -1tS` | last flag wins | same | name order | last flag wins |
| `ls -l` on a directory entry | `dir1` | `dir1` | `dir1/` | `dir1` |
| `ls -lF` on a directory entry | `dir1/` | `dir1/` | `dir1/` | `dir1/` |

just-bash already followed GNU on the missing-operand status and keeps its own `ls: NAME: No such file or directory` wording rather than GNU's `ls: cannot access 'NAME': …`. Matching GNU there would touch every `ls` error assertion in the suite, so it is not in this diff. Say the word and it is a separate change.

`minimatch` was `ls`'s only import, so dropping the matching loop leaves it unused and `knip` fails with it still declared. Removing it is therefore in this PR rather than a follow-up. The `blocked-globals` entry for `__MINIMATCH_TESTING_PLATFORM__` stays: it costs nothing and the global is worth refusing whether or not the package is installed.

The four changes are separate commits, in the order above. The type-indicator one is last and stands alone, so if you would rather keep the trailing slash, dropping that commit leaves the rest intact.

## Scope

Unchanged: shell-expanded globs, which the interpreter expands before `ls` sees them; `-R`; `-a` and `-A` filtering; the `name:` label text; the exit status for a missing operand; `-F` in both formats; the mode column; `@` and `*` indicators under `-F`.

Two security cases move rather than go away, and they are the part of this diff most worth a careful read:

- `ls 'bar*'` in `ls.security.test.ts` covered the prefix-sharing sibling leak (`/foo` vs `/foobar`) found in #307. That leak was a property of the matching loop, which is gone, so the case splits: the unexpanded pattern must now find nothing, and the shell-expanded `ls bar*` must still list only the cwd's entry. The second half is the one that would catch a regression.
- `filesystem-traversal.security.test.ts` drove `ls`'s tree walk through `ls '*'`, which no longer walks anything. `ls -R` is the only remaining path that descends, so it takes over the assertion, with a nested fixture deep enough to charge the budget. `ls` charges per directory entered, so `find`'s flat fixture does not trip it.

Not addressed, deliberately:

- `-S` ties were previously left in `readdir` order and are now broken by name. That falls out of the shared comparator rather than being a separate decision, and it matches GNU, but it is a behavior change beyond "implement `-t`". Happy to drop it if you would rather.
- `.` and `..` are still prepended after sorting under `-a`, so they head the list rather than taking their place in time or size order. Pre-existing, and equally true of `-S` today.
- Sort keys come from `stat`, so a symlink sorts by its target's size and time rather than its own. Pre-existing, equally true of `-S`, and fixing it means deciding what `-L` should do to sorting, which is a bigger question than this PR. Say the word and I will take it separately.
- `-u` and `-c` are not implemented and are not added. The filesystem keeps only an mtime, so there is nothing behind them.
- `ls -l` prints `total N` as an entry count where real `ls` prints disk blocks. Unrelated to any of the above, and changing it would rewrite fixtures across the suite.
- `ls -n` is still `invalid option`. It is a missing POSIX flag rather than a wrong answer, and uid and gid are hardcoded `user user` here, so it would print synthetic numbers. Say the word and it is a few lines.

## Tests

Three new unit files. `ls.literal-operands.test.ts` covers bracketed, `?` and `*` names listed literally and through shell expansion, operand spelling preserved for absolute paths, and an unmatched pattern reported as missing. `ls.operand-grouping.test.ts` covers file-block separation, files before directories regardless of command-line order, the label rule for one directory versus several, `-r` reversing both groups, `-S` ordering operands, a missing operand alongside good ones, and `-d`. `ls.sort-by-time.test.ts` covers newest-first ordering, that it differs from the default, `-tr`, the name tiebreak for equal timestamps and equal sizes, all four spellings of the `-S` and `-t` precedence, and operand ordering under `-t`.

Four new comparison files, every fixture recorded from the host's real `ls` rather than written by hand. The grouping harness caught the defect on its own while being recorded: the first attempt at `ls -1 *.pdf` failed with `"plain.pdf\n\nreport [1].pdf\n"` against bash's `"plain.pdf\nreport [1].pdf\n"`.

Two existing expectations flip, and they are the behavioral surface of this PR: the `find -exec ls {} +` batch case, and `should show directory indicator in long format`, which splits into one case asserting the mode column marks the directory with no suffix and one asserting `-F` still adds it.

What the fixtures cannot cover, rather than implying they do: long format never appears in a comparison test, because owner, mode and size will not agree with the host, so the long-format half of the type-indicator change is unit tests only. The missing-operand exit status is a unit test rather than a fixture, because GNU exits 2 and BSD exits 1, and recording on macOS would pin BSD's status.

`ls` suite 69 passing. Comparison suite 872 passing across 54 files. `tsc --noEmit`, biome, `lint:banned` and `knip` all clean. Changeset included.

`test:run` 15,303 passing, 97 skipped, zero failing, against a full `pnpm build`. `main` at bd1df37 is 15,254 passing, zero failing, under the same build, so this branch adds 49 tests and breaks nothing. Both numbers come from running the suite on each, not inferred. Worth noting for anyone reproducing: without `pnpm build` first, `main` reports 36 failures and so does any branch, since the CLI and worker tests load from `dist/`.

---

Authored with Claude Opus 5


mutewinter:
Working through the auto-maintain review. Two of its findings are fixed in `1ac0974`, one is declined with a reason, and one is wrong.

**Fixed: `-R` section ordering (`ls.ts:614`).** Correct, and it was the sharpest of the four, because `-t` is new in this PR. Sections were emitted in name order while each directory's own entries used the requested key, so `ls -Rt` was half sorted. Checkable against GNU coreutils ls 9.2 in one line:

```bash
gtouch -t 202601010000 zzz; gtouch -t 202001010000 aaa
gls -Rt        # zzz section first, then aaa
```

The descent now takes its order from the same comparator the entries use, and the batched results are restored to that order rather than re-sorted by name afterwards. Three cases added covering `-Rt`, plain `-R`, and `-Rtr`.

**Fixed: traversal budget during operand resolution (`ls.ts:235`).** Also correct. Every operand was stat'ed before the budget was charged, so a long list did all of its filesystem work before the guard could refuse any of it. Each operand is a visit like any other and is charged as one now, with a case in `filesystem-traversal.security.test.ts`. Agreeing with the second-opinion reviewer that this was about the guard tripping when it says it will, not about a reachable exhaustion.

**Declined for now: symlink metadata in sort keys (`ls.ts:357`).** Real, and pre-existing: `-S` has sorted by target size since before this PR. Fixing it properly means deciding what `-L` should do to sorting, which is a wider question than this diff, so I would rather do it separately than smuggle a `-L` semantics decision into a sorting fix. It is now listed under "Not addressed, deliberately" in the description. Happy to take it here if you would rather have it in one go.

**Wrong: `touch -d` portability (`ls-sort-by-time.comparison.test.ts:21`).** The finding reads `touch -d` and assumes the bare-date spelling. The command in that file is:

```bash
touch -d 2020-01-01T00:00:00 big-old.txt
```

BSD touch accepts a full ISO 8601 stamp; what it rejects is the bare `-d 2020-01-01` form, which is not what this uses. The direct evidence is that those fixtures were recorded on macOS, against BSD touch, which is the only way they exist. Re-confirmed just now on macOS 15:

```
$ /usr/bin/touch -d 2020-01-01T00:00:00 a.txt && ls -l a.txt
-rw-r--r--  1 user  staff  0 Jan  1  2020 a.txt
```

Worth flagging as a false positive rather than silently leaving it, since "GNU-only syntax" is the kind of note a future reader would take at face value.

Full suite after the changes: 15,303 passing across 579 files, zero failing, against a full `pnpm build`. `main` is 15,254 under the same build, so this is +49 tests and nothing broken.

mutewinter:
The re-run against `1ac0974` was right on all three code findings, and the sharpest of them is damage the previous round did. Fixed in `d386d2f`.

**Fixed: operands charged twice (`ls.ts:239`).** Correct, and it is a regression from the last commit rather than a pre-existing gap. Charging the operand where it is resolved and then again in `listPath` spends two entries on one visit, so the limit bit at half its stated capacity. Reproduced before the fix:

```
maxTraversalEntries: 1, ls /root/a         -> 126 "filesystem traversal entry limit exceeded (1)"
maxTraversalEntries: 2, ls /root/a /root/b -> 126 "filesystem traversal entry limit exceeded (2)"
```

`listPath` now takes the charge as already made when the caller resolved the operand, so the resolution loop stays the single place an operand is counted and it still refuses a long list before the stats run. The existing three-operands-against-two test still holds, and a two-against-two case pins the other side of the boundary.

**Fixed: `-d` bypassed the budget (`ls.ts:190`).** Correct, and reproduced: `ls -d` over three operands under `maxTraversalEntries: 2` exited 0 and listed all three. `-d` returns before the walk, which is exactly why it needed charging of its own. Each operand is a visit there too now.

**Fixed: uncharged sort metadata (`ls.ts:361`).** Correct. `-t` and `-S` read metadata for every name before printing any of it, and the reads were free. `sortNames` charges the whole batch up front, so a large flat directory is bounded by `maxTraversalWork` rather than sorting itself first and being metered afterwards. That covers the operand case the maintainer lens raised as well, since operand sorting runs through the same function.

**Fixed after all: symlinks in sort keys (`ls.ts:361`).** I declined this last round as an `-S` question that needs a `-L` decision first. That was wrong on the facts: `ls` here has no `-L`, so there is no semantics to defer, and `-t` is new in this PR. Sort keys come from `lstat` now. Checked against BSD ls (macOS 15):

```
$ touch -t 202001010000 old-target.txt && ln -s old-target.txt link-to-old
$ touch -h -t 202601010000 link-to-old && touch -t 202301010000 middle.txt
$ ls -t
link-to-old
middle.txt
old-target.txt
```

The link sorts first on its own 2026 stamp, not its target's 2020. `-S` picks up the same correction, which is the behavior GNU and BSD both have without `-L`.

**Still declined: `touch -d` portability (`ls-sort-by-time.comparison.test.ts:24`).** The refreshed review keeps this one, so restating it with the evidence rather than leaving it as a standing note: the fixture uses `touch -d 2020-01-01T00:00:00`, a full ISO 8601 stamp, not the bare `-d 2020-01-01` form BSD rejects. Those fixtures were recorded on macOS against BSD touch, which is the only way they exist. Re-confirmed on macOS 15 while preparing this round.

**On the red check.** `unit-tests (20)` failed on `1ac0974` with all 14,267 tests passing and then the vitest fork worker for `src/security/fuzzing/__tests__/fuzz-malformed.test.ts` exiting during teardown. Node 22 and 24 passed the same commit. The suite has no fixed fast-check seed, and running that file on Node 20 locally against this branch passes repeatedly, so I could not reproduce it. Flagging it rather than claiming it is fixed: this push gets a fresh run, and if Node 20 goes red again on the same file I will chase it properly.

Full suite after the changes: 11,473 passing across 574 files in `just-bash` plus 72 across 4 in `just-bash-executor`, zero failing, against a full `pnpm build`. `pnpm typecheck`, `pnpm lint`, and `pnpm knip` are clean.

mutewinter:
`unit-tests (20)` went red again on `d386d2f`, and I can now say what it is: not this PR, and not a flake either. `fuzz-malformed` runs fast-check with no fixed seed, so each CI run searches a different part of the space and occasionally lands on a real interpreter bug. This run landed on one.

```
Property failed after 42 tests
{ seed: -195205778, path: "41", endOnFailure: true }
Got AssertionError: byteInjected: unexpected error: Cannot read properties of undefined (reading 'line')
```

Minimized from the counterexample:

```bash
declare -A f; f[$n`]="x"
# TypeError: Cannot read properties of undefined (reading 'line')
```

An associative subscript that opens with a parameter expansion and then an unterminated backtick. It throws a `TypeError` out of `bash.exec()` rather than returning the parse error the neighbouring spellings return:

```bash
declare -A f; f[a`]="x"     # exit 0
declare -A f; f[`]="x"      # exit 0
declare -A f; f[$n]="x"     # exit 0
echo `                      # exit 2, "syntax error: Parse error at 1:7"
declare -A f; f[$n`]=x      # exit 0   <- unquoted RHS, no crash
```

Both the expansion and the unterminated backtick are needed, and the double-quoted right-hand side is too.

It reproduces on a branch cut from `main` that contains none of the `ls` work, so it predates this PR. That also explains the earlier red run here, and the one on `main` at `bd1df37`: same job, different seed, different underlying bug each time. Worth knowing that a green `unit-tests` on any PR in this repo currently means "the fuzzer did not happen to look there", which is a separate problem from any individual diff.

Leaving it out of this diff, since a parser fix has nothing to do with `ls` operand handling and deserves its own tests. Happy to open that one next if it is useful, or to leave it with you.

Everything else on this branch is green: `comparison-tests`, `lint`, `typecheck`, `Node 22`, `Node 24`. The `unit-tests (22)` and `(24)` entries show as failed only because the matrix cancels them when one leg fails; both had passed their unit-test step when they were cancelled.

mutewinter:
Rebased onto `de3c2f3` to clear the conflict. Still an issue on current `main`, and ready to merge.

The conflict was entirely dependency churn from `guarded-fetch` (#380) and the refreshes in between, in `packages/just-bash/package.json` and `pnpm-lock.yaml`. No conflict in `ls.ts` or in any of the tests and comparison fixtures, so nothing about the behavior under review moved. I took `main`'s dependency block and re-applied only this branch's intent on top of it: drop the direct `minimatch` dependency and its three `--external:minimatch` bundler flags. `minimatch@10.2.6` stays in the lockfile as a transitive dependency of two other packages, so only the importer entry went. Rather than regenerate the lockfile, which floated `@types/node` 25.9.5 to 26.2.0 and esbuild 0.27.7 to 0.28.2, I removed that one entry by hand and confirmed with `pnpm install --frozen-lockfile`, so this branch carries no dependency drift beyond the one removal.

Verified on the rebased head: lint and typecheck clean, unit suite 544 files and 14,687 passed with 98 skipped, comparison suite 54 files and 872 passed.

On the red `unit-tests (20)` from the previous head: that was `fuzz-malformed` running fast-check with no fixed seed, so each CI run draws different inputs and the failure moves between runs and between Node versions. It is unrelated to this PR. The full unit suite passes on this rebase, which is a fresh draw rather than proof, so I would not read a green run here as having fixed anything. Happy to seed that suite in a separate PR if you want the noise gone; say the word and it is a one-line change.

The behavior under review is unchanged from what you last saw. Five things, each with its own commit and changeset: operands resolved literally rather than re-matched as globs, GNU-style operand grouping, `-t` implemented, type indicators only with `-F`, and `-R` sections ordered by the sort key.

cramforce:
@auto-maintain review

cramforce (approved):

PR #364, opened 2026-08-11
fix(touch): honor -t and -r instead of discarding them

mutewinter:
## Problem

`touch -t` does not set the timestamp. It creates the file, stamps it with the current time, and exits 0.

```bash
touch -t 202001010000 f.txt && ls -l f.txt
# -rw-r--r-- 1 user user 0 Aug 11 08:59 f.txt      <- now, not 2020
```

`-r` behaves the same way:

```bash
touch -t 202001020304 ref.txt
touch -r ref.txt f.txt && ls -l f.txt
# -rw-r--r-- 1 user user 0 Aug 11 08:59 f.txt      <- now, not ref's time
```

Neither reports anything, and both consume their argument, so nothing about the invocation looks wrong.

A third case, in a flag that is implemented: neither `-t` nor `-d` names a timezone, and both were read against whatever zone the host process happened to be in. The same script produced a different instant depending on where the sandbox ran, and the host's zone leaked into a timestamp nobody asked to be host-relative.

```bash
TZ=America/Chicago touch -d 2021-01-01 f.txt
# stored 2021-01-01T00:00:00Z, ignoring $TZ; GNU stores 2021-01-01T06:00:00Z
```

The point of these flags is that the mtime is data: a fixture that has to look old, a comparison against `find -newer`, a sort by age, a build stamp. Every one of those gets a wrong answer with nothing to attribute it to. It is worse for an agent than for a person, because the natural way to check the result is `ls -lt` or `find -newer`, and both of those agree with the timestamp that was actually written rather than the one that was asked for. The evidence confirms the mistake.

## Cause

`-t` and `-r` were parsed for the sole purpose of stepping over their argument so it would not be read as a filename:

```ts
} else if (arg === "-a" || arg === "-m" || arg === "-r" || arg === "-t") {
  // Silently ignore -a (access time only), -m (modify time only),
  // -r (reference file), -t (timestamp format) for now
  if (arg === "-r" || arg === "-t") {
    // These take an argument
    i++;
  }
}
```

`targetTime` stayed null and the write loop fell through to `new Date()`.

The timezone gap is a different mechanism. `parseDateString` never saw `$TZ` at all: it handed the string to `new Date`, which reads a bare `YYYY-MM-DD` as UTC and anything with a time as host-local. `date` already had the contract this needed, in `parseDate`/`parseBareISOInTimezone`: `$TZ` when the shell sets a zone Intl can resolve, UTC otherwise, so the host's zone never leaks unless the caller opts in. `touch` simply was not using it.

The suite could not see any of it. There was no test that set a timestamp and then read it back, so the only assertions on `touch -t` were that it exits 0 and creates the file, which it did.

## Fix

`-d`, `-t` and `-r` all name one thing, so they share one slot and the last one written wins, matching GNU. `-t` parses the POSIX stamp `[[CC]YY]MMDDhhmm[.ss]`, `-r` stats the reference, and both `-t` and `-d` resolve a zone-less spelling through `$TZ`, falling back to UTC. `isValidTimezone` and `parseBareISOInTimezone` move out of `date.ts` into `commands/timezone.ts` unchanged so both commands share one implementation rather than letting a second one drift.

Two-digit years pivot at 69, which is the boundary POSIX fixes for this format. Measured against GNU coreutils touch 9.2:

| stamp | GNU 9.2 | before | after |
| --- | --- | --- | --- |
| `202601020304` | 2026-01-02 03:04:00 | now | 2026-01-02 03:04:00 |
| `2601020304.05` | 2026-01-02 03:04:05 | now | 2026-01-02 03:04:05 |
| `6901020304` | 1969-01-02 03:04:00 | now | 1969-01-02 03:04:00 |
| `6801020304` | 2068-01-02 03:04:00 | now | 2068-01-02 03:04:00 |
| `01020304` | current year, 01-02 03:04 | now | current year, 01-02 03:04 |
| `99` | `invalid date format`, exit 1 | silently ignored, exit 0 | `invalid date format`, exit 1 |

Out-of-range fields and a day the month does not have are rejected rather than rolled forward, which is what `new Date` would otherwise do with `202002300000`.

One line in here is worth a second look, because the reason for it is not local:

```ts
targetTime = new Date((await ctx.fs.stat(reference)).mtime.getTime());
```

The runtime guard added in #307 wraps values returned from `ctx.fs`, and the wrapper stops answering once the command ends. Keeping that `Date` and writing it into the filesystem left an mtime that throws `ExecutionAbortedError` on the next read, in whatever unrelated command happened to read it next. Copying the instant out is the fix. Any command that persists something obtained from `ctx.fs` has the same trap available to it, so it may be worth a note in the command-authoring docs.

Both of the command's catch blocks now call `rethrowFatalExecutionError` first, as the helper's doc comment asks broad catch blocks to. The `-r` one was new here; the per-file write loop has swallowed limits, aborts and security violations for as long as it has existed, reporting them as `touch: cannot touch '<file>'` and continuing to the next operand. That half predates this PR and is included because stopping after fixing the identical bug five lines above would have been a strange place to leave it.

## Scope

Unchanged: bare `touch`, `-c`, `--no-create`, `--date=`, `--`, combined short options, and the missing-operand and invalid-date diagnostics.

Not addressed, deliberately:

- `-a` and `-m` stay ignored, and the comment now says why rather than listing them beside the flags that work. The filesystem interface keeps only a modification time, and `utimes`' own doc comment says atime is "currently ignored, kept for API compatibility", so there is nothing for them to select between. Making `touch -a` correctly leave the mtime alone needs an atime in the fs layer first. Say the word and I will take that as a follow-up.
- `-t 6901010000` resolving to 1969 is correct and also surprising. Left as POSIX specifies it.
- Attached short-option arguments (`-t202601020304`, `-rref`) are not accepted; the value comes from the next word. GNU takes both. This is the pre-existing shape of this parser, which `-d` shared before this PR, so fixing it is a change to option parsing rather than to timestamps and belongs on its own. Say the word and I will do it here instead.
- `ls -l` renders timestamps from host-local components, so a correctly stored UTC instant can display as the previous day. That is what first sent me looking, and it is an `ls` defect that predates this PR. Not touched from the `touch` side, because making the parser agree with a leaky formatter would spread the leak rather than fix it.

## Tests

`touch.timestamp-flags.test.ts` is new: twenty cases across the `-t` stamp forms in the table above, the current-year default, file creation, four rejected stamps, the missing argument, the combined short form `-ct`, `-r` including a missing reference, and `-d` against `-t` precedence in both orders. Timezone coverage is four of those: `$TZ` set, `$TZ` set to something Intl cannot resolve, an explicit offset beating `$TZ`, and no `$TZ` at all. Assertions are on UTC components, since a local-component assertion passes or fails depending on where the suite runs.

`touch-timestamp-flags.comparison.test.ts` is new, four cases recorded from the host's real `touch` rather than written by hand. A timestamp cannot be compared against the host directly, since `ls -l` renders owner and size, so each case stamps files and then asks `find -newer` which came out newer. A stamp that is ignored collapses that answer, which is what the cases catch.

What is not covered, rather than leaving a hole in the fixture set unexplained: `-d`'s bare-date spelling is unit-tested only. BSD touch rejects `-d 2021-01-01`, so recording that fixture on macOS pins BSD's refusal rather than GNU's behavior. The first recording attempt did exactly that and came back with an empty expectation, which is how I noticed. Timezone behavior is unit-tested only for the same reason: the fixture would encode the recording host's zone.

`touch` suite 32 passing, including `touch.fatal-errors.test.ts` covering both catch blocks, and the existing `date` suite still passes against the shared timezone module. Comparison suite 859 passing. `tsc --noEmit`, biome, `lint:banned` and `knip` all clean. Changeset included.

`test:run` 15,279 passing, 97 skipped, zero failing, against a full `pnpm build`. `main` at bd1df37 is 15,254 passing, zero failing, under the same build, so this branch adds 28 tests and breaks nothing. Both numbers come from running the suite on each, not inferred. Worth noting for anyone reproducing: without `pnpm build` first, `main` reports 36 failures and so does any branch, since the CLI and worker tests load from `dist/`.

---

Authored with Claude Opus 5


mutewinter:
Working through the auto-maintain review. Two findings are fixed in `bc1f33d`, one is declined with a reason. The TZ one was the most useful thing anybody has said about this PR, because chasing it showed the original change was fixing the wrong end of the problem.

**Fixed: `$TZ` is now honored (`touch.ts:24`).** Correct. `parseDateString` never saw `$TZ`; it handed the string to `new Date`, which reads a bare `YYYY-MM-DD` as UTC and anything carrying a time as host-local. `date` already had the contract for this, so `isValidTimezone` and `parseBareISOInTimezone` move out of `date.ts` into `commands/timezone.ts` unchanged and both commands import them. `-t` and `-d` now resolve a zone-less spelling through `$TZ`, falling back to UTC when the shell has none or when Intl cannot resolve what it has.

Worth being explicit that this reverses part of the first commit, because the original reasoning was wrong. That commit read a bare `-d YYYY-MM-DD` as host-local midnight, on the strength of `ls -l` displaying the previous day for it. The stored instant was already correct: `new Date("2021-01-01")` is UTC midnight, which is the documented no-`$TZ` default. What renders the day before is `ls`'s own formatter, which prints host-local components and leaks the host zone in exactly the way `date.ts` documents it must not. That is an `ls` defect, it predates both of my PRs, and it is now named under "Not addressed, deliberately" rather than papered over from the `touch` side. Making the parser agree with a leaky formatter would have spread the leak.

Tests now assert UTC components, since a local-component assertion passes or fails depending on where the suite runs, with four cases for `$TZ` set, `$TZ` unresolvable, an explicit offset beating `$TZ`, and no `$TZ` at all.

**Fixed: fatal errors no longer masked (`touch.ts:215`).** Correct, and thank you for it. The `-r` catch turned every failure into "No such file or directory", including `ExecutionLimitError`, `ExecutionAbortedError` and `SecurityViolationError`. It calls `rethrowFatalExecutionError` first now, matching `stat`, `ls`, `cp` and `mv`.

**Declined for now: attached short-option arguments (`touch.ts:171`).** Accurate: `-t202601020304` and `-rref` take the next word instead of the remainder, and GNU accepts both spellings. It is the pre-existing shape of this parser, though, and `-d` shared it before this PR, so fixing it is a change to option parsing rather than to timestamps. Splitting it out keeps this diff to the defect and makes that one reviewable on its own terms, including the `-tc` case where GNU getopt would read `c` as the value. Listed in the description; say the word and I will fold it in here instead.

Full suite after the changes: 15,279 passing across 574 files, zero failing, against a full `pnpm build`. `main` is 15,254 under the same build, so this is +25 tests and nothing broken, including the existing `date` suite now running against the shared timezone module.

mutewinter:
Good catch, and a fair one to make against the commit that fixed the same thing next door. Fixed in `795f185`.

The `-r` catch I fixed in the previous round was the new one. The per-file write loop has had the same defect for as long as it has existed: it reports everything as `touch: cannot touch '<file>'` and moves to the next operand, which is right for an ordinary write failure and wrong for an `ExecutionLimitError`, an `ExecutionAbortedError` or a `SecurityViolationError`, since those mean the run is over rather than that one file did not work. Swallowing one hid the real reason and let the script continue past a stop signal.

Both catch blocks call `rethrowFatalExecutionError` first now, as the helper's own doc comment asks broad catch blocks to.

Including the pre-existing half here rather than splitting it, on the grounds that stopping after fixing the identical bug five lines above would have been an odd place to leave it. Happy to pull it into its own PR if you would rather keep this diff to the flags.

`touch.fatal-errors.test.ts` covers both blocks with a filesystem that throws a limit error from `utimes` and from `stat`, and keeps an ordinary write failure alongside them so the per-file recovery path stays pinned:

```
touch /w/boom.txt        # limit from utimes  -> exit 126, "limit exceeded", no "cannot touch"
touch -r /w/ref.txt f    # limit from stat    -> exit 126, no "failed to get attributes"
touch /w/nope.txt        # ordinary failure   -> exit 1, "touch: cannot touch '/w/nope.txt'"
```

One note for whoever picks this up: `cp`, `mv`, `mkdir` and `rm` do not call `rethrowFatalExecutionError` in their own loops either. I have left them alone since they are outside this PR, but the same reasoning applies to all four.

Suite after the change: 15,282 passing across 575 files, zero failing, against a full `pnpm build`.

mutewinter:
The re-run against `795f185` was right on three of four. Fixed in `3ff1a69`; the fourth is declined with a reason.

**Fixed: zone-less `-d` values carrying a time (`touch.ts:34`).** Correct, and the previous round only half fixed this. The tz-aware branch was gated on `$TZ` being set, so with no `$TZ` a spelling with a time fell through to `new Date`, which reads a bare date as UTC and anything with a time as host-local. Measured on the old code by running the same script under three host zones:

```
                    host TZ=UTC              host TZ=America/Chicago   host TZ=Asia/Tokyo
-d '2021-01-01 10:00:00'   2021-01-01T10:00:00Z   2021-01-01T16:00:00Z    2021-01-01T01:00:00Z
-t 202101011000            2021-01-01T10:00:00Z   2021-01-01T10:00:00Z    2021-01-01T10:00:00Z
```

Exactly the disagreement you described, and against the changeset's own claim. A zone-less spelling now resolves through the same path with `UTC` standing in for an absent `$TZ`, so all three rows read `10:00:00Z` on every host.

**Fixed: fractional seconds (`touch.ts:34`).** Correct, and worse than reported. `parseBareISOInTimezone`'s grammar had no fractional part, so `-d '2021-01-01T10:00:00.500'` failed the match and fell through **even when `$TZ` was set**: under `TZ=Asia/Tokyo` it returned `10:00:00.500Z` on a UTC host and `01:00:00.500Z` on a Tokyo host, while the identical spelling without the fraction correctly returned `01:00:00.500Z` on both. The grammar takes 1 to 3 fractional digits now. The fraction is carried around the convergence loop rather than through it, since the zone reports whole seconds and a fraction would otherwise read as permanent drift and never converge.

**Fixed: yearless `-t` year (`touch.ts:67`).** Correct. `new Date().getFullYear()` is the host's calendar year by construction, while the wall time resolved in `$TZ`/UTC. `currentYearInTimezone` reads it from the zone instead, falling back to the UTC year when there is no `$TZ` or when Intl cannot resolve it. It takes the instant as a parameter so the New Year boundary is testable without patching the global clock, which the defense-in-depth suite does not appreciate.

**Fixed: `-r` failure reasons (`touch.ts:218`).** Correct. Absence keeps the wording `touch` uses for it, and anything else keeps the reason it came with, so a reference that exists but cannot be read no longer sends the caller looking for a file that is right there. Passing the raw error through unconditionally was not an option: it would have turned the common case into `ENOENT: no such file or directory, stat '/w/nope.txt'`. This matches what the per-file write loop already does with `getErrorMessage`.

**Declined: POSIX TZ spellings (`touch.ts:201`).** Accurate, and I confirmed it: `TZ=EST5` is silently read as UTC. Two reasons to keep it out of this diff. It is the pre-existing contract of the shared timezone module, which `date` has had since before this PR and which `isValidTimezone` documents as deliberate ("matching GNU `date`, which silently uses local time on invalid `TZ`"), so changing it changes `date` too. And note the reported examples are not equivalent: `EST5EDT` is a real IANA zone that Intl accepts and resolves correctly; only the offset-only POSIX forms are affected. Supporting POSIX TZ syntax, or rejecting it loudly, is worth doing on its own terms where the `date` behavior change is the subject. Listed under "Not addressed, deliberately". Happy to take it here instead if you would rather.

One thing worth naming that is outside this PR: `date.test.ts` has two cases, both called "local noon stays same day", that fail on a host at UTC+14. I confirmed they fail at `795f185` without any of these changes, so they are not new, but they are the same class of host-zone leak this PR is removing from `touch`, sitting in `date`'s own tests.

Full suite after the changes: 11,469 passing across 571 files in `just-bash` plus 72 across 4 in `just-bash-executor`, zero failing, against a full `pnpm build`. The touch, timezone and date suites also pass with the host set to UTC, America/Chicago and Asia/Tokyo, which is the property that matters here. `pnpm typecheck`, `pnpm lint`, and `pnpm knip` are clean.

mutewinter:
Rebased onto `main` at #436 (26 commits, no conflicts; unit 14,856, comparison 876, spec 3,834, typecheck and biome clean).

@cramforce this one has sat while its sibling #363 merged, and I suspect the size is why: `-t`/`-r` on their own are small, but honoring `$TZ` pulled in `timezone.ts` and a change to `date`, and the diff reads as more than a `touch` fix. If that is the hesitation, say so and I will split it: `-t` and `-r` honored (with the fatal-error passthrough) as one PR, and the `$TZ` handling for `-t`/`-d` with the `date` change as a second on top. If it is something else, also happy to hear it.

Issue #266, opened 2026-06-08
jq: standard array-removal idioms (`-=`, `del(.arr[]|select(...))`) silently no-op


audun-cobrief:
## Summary

Two standard jq idioms for removing elements from an array are **silent no-ops** in just-bash — they exit 0 with no error and leave the array unchanged:

1. `.arr -= ["x"]` (compound subtract-assign)
2. `del(.arr[] | select(. == "x"))` (delete via filtered path expression)

Both work in real jq. Because they exit 0, callers (especially LLM agents driving just-bash) believe the edit succeeded when the data is untouched — a no-op reported as success. The two have **distinct root causes**, detailed below.

## Environment

- `just-bash@3.0.1` (also present on `main` @ `9481331`)
- Reproduced via the public `Bash` API and the bundled `jq`.

## Repro

```js
import { Bash } from "just-bash";
const sh = new Bash();

// (1) -= on an array
console.log((await sh.exec(
  `echo '{"arr":["a","b","c"]}' | jq -c '.arr -= ["b"]'`
)).stdout.trim());

// (2) del() via a filtered path expression
console.log((await sh.exec(
  `echo '{"arr":["a","b","c"]}' | jq -c 'del(.arr[] | select(. == "b"))'`
)).stdout.trim());
```

### Actual (just-bash 3.0.1)

```
{"arr":["a","b","c"]}      // (1) -= ignored
{"arr":["a","b","c"]}      // (2) del removed nothing
```

### Expected (real jq 1.7.1)

```
{"arr":["a","c"]}
{"arr":["a","c"]}
```

## What *does* work (scopes each bug)

These behave correctly in just-bash today, isolating the two problems:

```
.arr += ["d"]                      → {"arr":[...,"d"]}      // += on arrays: OK
.arr |= (. - ["b"])                → {"arr":["a","c"]}      // binary - on arrays: OK
.arr |= map(select(. != "b"))      → {"arr":["a","c"]}      // OK
del(.arr[1])                       → {"arr":["a","c"]}      // del of a static path: OK
del(.x)                            → {"y":2}                 // del of an object key: OK
.n *= 2  (on {"n":3})              → {"n":6}                 // *= on numbers: OK
```

So all the machinery exists; only `-=`-on-non-numbers and path-collection-through-iteration are broken.

---

## Root cause 1 — `-=` only handles numbers

`packages/just-bash/src/commands/query-engine/evaluator.ts`, `applyUpdate` → `computeNewValue`:

[...]

PR #384, opened 2026-08-23
Fix jq to_entries on arrays and tonumber on empty strings

taoche:
## Summary

Two silent divergences from real jq in the query engine, both discovered while tracing an AI agent doing CSV analysis in the sandbox (same "silently no-op" family as #266):

**1. `to_entries` on an array returns `null` → `to_entries[]` silently emits nothing (exit 0)**

```bash
$ echo '["A","B"]' | jq -r 'to_entries[] | "\(.key): \(.value)"'
# just-bash: empty output, exit 0
# real jq:   0: A
#            1: B
```

Real jq's `to_entries` is `keys_unsorted`-based, so arrays yield numeric-key entries — and this engine's own `keys`/`keys_unsorted` builtins already support arrays exactly that way (`object-builtins.ts`), so this also removes an internal inconsistency. The array path reuses the existing `maxArrayElements` limit guard.

**2. `tonumber` on `""` / whitespace-only strings returns `0`**

```bash
$ echo '["2","","4"]' | jq '[.[] | tonumber] | add'
# just-bash: 6
# real jq:   jq: error: "" cannot be parsed as a number
```

`Number("")` is `0` in JS, so empty cells silently coerce to zero. For agents aggregating tabular data this masks data-quality problems — a sum over a column with missing values just looks like a valid total. Real jq rejects it; now the existing `cannot be parsed as a number` error fires.

## Tests

- `jq.functions.test.ts`: array → numeric-key entries; `to_entries[]` iteration over an array
- `jq.operators.test.ts`: `tonumber` rejects `""` and `" "` with non-zero exit
- Full `src/commands/jq`, `src/commands/yq`, `src/commands/query-engine` suites pass (34 files, 656 tests)

### DIFF
--- a/packages/just-bash/src/commands/jq/jq.operators.test.ts
+++ b/packages/just-bash/src/commands/jq/jq.operators.test.ts
@@ -174,5 +174,19 @@ describe("jq operators", () => {
       const result = await env.exec("echo '\"42\"' | jq 'tonumber'");
       expect(result.stdout).toBe("42\n");
     });
+
+    it("should reject tonumber on an empty string", async () => {
+      const env = new Bash();
+      const result = await env.exec("echo '\"\"' | jq 'tonumber'");
+      expect(result.exitCode).not.toBe(0);
+      expect(result.stderr).toContain("cannot be parsed as a number");
+    });
+
+    it("should reject tonumber on a whitespace-only string", async () => {
+      const env = new Bash();
+      const result = await env.exec("echo '\" \"' | jq 'tonumber'");
+      expect(result.exitCode).not.toBe(0);
+      expect(result.stderr).toContain("cannot be parsed as a number");
+    });
   });
 });
--- a/packages/just-bash/src/commands/query-engine/builtins/object-builtins.ts
+++ b/packages/just-bash/src/commands/query-engine/builtins/object-builtins.ts
@@ -23,6 +23,13 @@ function maxResultElements(ctx: EvalContext): number {
   return ctx.limits.maxArrayElements;
 }
 
+function makeEntry(key: string | number, value: unknown): Record<string, unknown> {
+  const entry: Record<string, unknown> = Object.create(null);
+  safeSet(entry, "key", key);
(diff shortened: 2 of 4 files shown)


cramforce (approved):

cramforce:
Please add a changeset

taoche:
Added in 1c7f10c — `just-bash` patch changeset covering both fixes.

packages/just-bash/src/commands/query-engine/builtins/object-builtins.ts
```diff
@@ -117,6 +127,15 @@ export function evalObjectBuiltin(
     }
 
     case "to_entries": {
+      if (Array.isArray(value)) {
+        if (value.length > maxResultElements(ctx)) {
```
privatenumber: Could we add a boundary test showing that `to_entries` accepts an array exactly at `maxArrayElements` and throws `ExecutionLimitError` for one element over?

The check looks correct. A focused test would help protect this new allocation path from losing its resource guard in a future refactor. This doesn't need to block merging.

packages/just-bash/src/commands/jq/jq.operators.test.ts
```diff
@@ -174,5 +174,19 @@ describe("jq operators", () => {
...
     });
+
+    it("should reject tonumber on an empty string", async () => {
+      const env = new Bash();
+      const result = await env.exec("echo '\"\"' | jq 'tonumber'");
+      expect(result.exitCode).not.toBe(0);
+      expect(result.stderr).toContain("cannot be parsed as a number");
+    });
+
+    it("should reject tonumber on a whitespace-only string", async () => {
```
privatenumber: Could we also cover a tab/newline-only string and a valid number with surrounding whitespace, such as `" 42 "`? Real jq rejects the first and returns `42` for the second.

The implementation already handles both correctly; these would help preserve that distinction if number parsing changes later. Happy for this to follow separately.

privatenumber (approved):

PR #389, opened 2026-08-25
feat(curl): implement -D/--dump-header

### LINKED ISSUE #388: curl: -i/-I report only a subset of response headers (CSP, HSTS, X-Frame-Options silently missing); -D/--dump-header unimplemented
## Summary

`curl -i` / `-I` report only a small subset of the response headers a server actually sends. Security-relevant headers (`content-security-policy`, `strict-transport-security`, `x-frame-options`, `referrer-policy`, …) are silently absent, and the output is shaped like a complete header dump — so it reads as authoritative rather than partial. `-D` / `--dump-header` are not implemented at all.

This is a *silent wrong answer* rather than a missing feature: an agent inspecting headers concludes "this server sets no CSP" when the server does set one.

## Environment

- `just-bash` as embedded in SLICC 6.96.2 (`uname -r`)
- Commands run in the sandboxed shell; no proxy configured by the user

## Reproduction

```console
$ curl -s -i https://github.com/ | sed -n '1,/^$/p' | grep ":" | cut -d: -f1
cache-control
content-language
content-type
set-cookie
```

Real `curl` against the same URL returns roughly twenty headers, including `content-security-policy`, `strict-transport-security`, `x-frame-options`, `x-content-type-options` and `referrer-policy`.

### Second reproduction, independent of GitHub

A Cloudflare Worker whose source provably contains:

```js
headers.set("etag", object.httpEtag);
headers.set("cache-control", cacheControlFor(matchedKey));
headers.set("Access-Control-Allow-Origin", "*");
```

`curl -s -I <that url>` reports only `cache-control` and `content-type`. Both `etag` and `access-control-allow-origin` are dropped, even though the code that sets them is known.

### The limitation is below `curl`, not in it

`fetch()` inside the same runtime sees the same three headers:

```console
$ node -e '(async()=>{const r=await fetch("https://github.com/");console.log([...r.headers.keys()].join(", "))})()'
cache-control, content-language, content-type
```

`cache-control`, `content-language` and `content-type` are three of the seven [CORS-safelisted response headers](https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_response_header), which suggests the request is issued through a browser `fetch` where the response carries no `Access-Control-Expose-Headers`. So a fix confined to `commands/curl/response-formatting.ts` would not be sufficient — `result.headers` never receives the data.

It is not a strict safelist, though, so something more nuanced is going on:

```console
$ curl -s -I https://www.aem.live/ | cut -d: -f1 | tr '\n' ' '
HTTP/1.1 200 OK cache-control content-type last-modified link
```

`link` is not CORS-safelisted yet appears; `set-cookie` also appears for github.com and is normally a *forbidden* response header.

## `-D` / `--dump-header`

Not implemented. These do fail loudly, which is correct behaviour — noting them only because they are the usual way to inspect headers, and they become worth adding once the header set is complete:

```console
$ curl -s -D /tmp/h.txt -o /dev/null https://github.com/ ; echo "rc=$?"
curl: invalid option -- 'D'
rc=1

$ curl -s --dump-header /tmp/h.txt -o /dev/null https://github.com/ ; echo "rc=$?"
curl: unrecognized option '--dump-header'
rc=1
```

Note that `-i` combined with `-o <file>` sends the body to the file and the headers to stdout, matching real `curl`. That part is fine — it just makes the truncation easy to miss, since `curl -i -o /dev/null <url> 2>/dev/null` prints nothing at all and looks like "no headers".

## Impact

Any header-dependent task silently produces a confident, wrong result. Concretely, I inspected a Cloudflare Worker to determine whether it set `Content-Security-Policy` or `X-Frame-Options` on framed content, concluded from empty output that it set neither, and reported that as a security finding. Reading the worker's source later showed the header set was simply never visible to me. Header inspection is a normal part of debugging CORS, caching, redirects, auth and framing, and today it cannot be done from this shell — while appearing to work.

## Suggested resolution

1. **Preferred:** surface the complete response header set — e.g. issue the request through the server-side/proxy path that already bypasses CORS for response *bodies*, and carry the headers through with it.
2. **Minimum viable, if the full set is structurally unavailable:** have `-i` / `-I` write a one-line warning to stderr stating that the header set is filtered and which headers can be represented, so a partial dump is not mistaken for a complete one. A loud limitation is much safer than a quiet omission.
3. Once headers are available, implement `-D` / `--dump-header` (and `-D -` for stdout).

Happy to test a patch against the same reproductions.

trieloff:
## Summary
- Implements `curl -D` / `--dump-header` (including `-D -` for stdout and the attached `-Dfile` form), matching real curl's write-headers-to-file behavior.
- Addresses the loud missing-option half of #388. The silent header-truncation half does **not** reproduce in stock Node just-bash (full CSP/HSTS/X-Frame-Options etc. are returned); that is a SLICC `/api/fetch-proxy` CORS `Access-Control-Expose-Headers` issue and will be fixed separately there.

## Test plan
- [x] `pnpm test:run src/commands/curl/tests/dump-header.test.ts`
- [x] Related curl option/verbose/writeout tests
- [x] `pnpm typecheck` / biome / knip
- [ ] Manual: `curl -s -D /tmp/h.txt -o /dev/null https://example.com` writes a header block to `/tmp/h.txt`

Fixes #388 (partial: `-D` only)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changeset/curl-dump-header.md
+++ b/.changeset/curl-dump-header.md
@@ -2,4 +2,4 @@
 "just-bash": patch
 ---
 
-Implement `curl -D` / `--dump-header` so response headers can be written to a file or stdout (`-D -`)
+Implement `curl -D` / `--dump-header` (file, `-`, `--dump-header=`, redirect hops, `-f` dumps). `curl -I` header blocks now end with a blank line (`\r\n\r\n`), matching real curl.
--- a/.changeset/curl-dump-header.md
+++ b/.changeset/curl-dump-header.md
@@ -1,0 +1,5 @@
+---
+"just-bash": patch
+---
+
+Implement `curl -D` / `--dump-header` so response headers can be written to a file or stdout (`-D -`)
--- a/packages/just-bash/src/commands/curl/curl.ts
+++ b/packages/just-bash/src/commands/curl/curl.ts
@@ -20,7 +20,7 @@ import { parseOptions } from "./parse.js";
 import {
   applyWriteOut,
   extractFilename,
-  formatHeaders,
+  formatHeaderBlock,
 } from "./response-formatting.js";
 import type { CurlOptions } from "./types.js";
 
@@ -208,9 +208,11 @@ function buildOutput(
 
   // Include headers with -i/--include
   if (options.includeHeaders && !options.verbose) {
-    output += `HTTP/1.1 ${result.status} ${result.statusText}\r\n`;
-    output += formatHeaders(result.headers);
-    output += "\r\n\r\n";
+    output += formatHeaderBlock(
+      result.status,
+      result.statusText,
+      result.headers,
+    );
   }
 
   // Add body (unless head-only mode)
@@ -220,9 +222,11 @@ function buildOutput(
     // For HEAD, we already showed headers
   } else {
     // HEAD without -i shows headers
-    output += `HTTP/1.1 ${result.status} ${result.statusText}\r\n`;
-    output += formatHeaders(result.headers);
-    output += "\r\n";
+    output += formatHeaderBlock(
+      result.status,
+      result.statusText,
+      result.headers,
+    );
   }
 
   // Write-out format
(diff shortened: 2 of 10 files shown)


trieloff:
Addressed the auto-maintain review findings in b72e4f84:

| Finding | Fix |
| --- | --- |
| `--dump-header=FILE` rejected | Parsed like other long options |
| `-f -D` skipped the dump | Headers written before the `-f` early return (incl. `-D -` on stdout) |
| Dump file not truncated on start | Truncated before `fetch`; connection failures leave an empty file |
| `-v -D - -o` dropped the raw block | `-D -` always prepends the header block; `-w` no longer wipes verbose |
| Redirect hops omitted from `-D` | `FetchResult.redirectChain` records intermediate responses; `-D` concatenates hop + final blocks |
| `-I` trailing blank line | Kept `\r\n\r\n` (matches real curl); noted in the changeset |

Happy to adjust further if anything still looks off.

cramforce (approved):

PR #390, opened 2026-08-27
fix(ls): bound directory entry collections before per-entry work

### LINKED ISSUE #340: ls can exhaust host memory on very large directories
## Summary

`ls` can exhaust the embedding process's memory when a filesystem
implementation returns a very large directory.

The command currently:

1. eagerly awaits `ctx.fs.readdir(fullPath)`,
2. sorts the entire returned array,
3. may stat every entry for `-l`, `-S`, `-F`, or `-R`, and
4. accumulates formatted results before the existing output-size limit is
   enforced.

Because pipeline producers are materialized before consumers run, piping to
`head` does not bound this work.

## Synthetic reproduction

Use an `IFileSystem` implementation whose `readdir()` returns an array larger
than `executionLimits.maxArrayElements`, then run:

```bash
ls /large
ls -l /large | head -1
```

Both commands currently proceed into sorting and/or per-entry formatting
instead of failing at the configured collection limit.

## Expected behavior

`ls` should fail with a controlled resource-limit error before sorting,
statting, classifying, or formatting more directory entries than
`executionLimits.maxArrayElements`.

A complete fix should also avoid the initial unbounded allocation where
filesystem backends can support bounded or incremental directory reads.
Backwards compatibility for custom `IFileSystem` implementations should be
preserved.

## Acceptance criteria

- `ls` rejects a directory result larger than `maxArrayElements` with exit
  code 126 and a specific stderr message.
- The limit is checked before per-entry stat/classification work.
- `ls -l /large | head -1` cannot bypass the limit.
- Recursive and glob listing paths apply equivalent collection bounds.
- Focused regression tests cover a filesystem returning more entries than the
  configured limit.

trieloff:
Fixes #340.

## Verification that the issue still reproduces

Confirmed on `main` (de3c2f3). With `maxTraversalEntries: 5` and a directory of 50 entries, every `ls` variant listed the whole directory and exited `0`:

```
PLAIN       exit 0  ""  (51 lines)
LONG|head   exit 0  ""  "total 50\n"
CLASSIFY    exit 0  ""
RECURSIVE   exit 0  ""
SORTSIZE    exit 0  ""
```

`find` on the same tree already failed closed at 126, so `ls` was the outlier.

## Root cause

`ls` eagerly awaited `ctx.fs.readdir()`, sorted the entire result, statted every entry for `-l`/`-S`/`-F`/`-R`, and only enforced a bound once formatted output was appended. Piping to `head` did not help because pipeline producers are materialized before consumers run.

## Change

Entries are admitted into the command's budgets at the point they enter `ls`, before any sorting, statting, classifying or formatting:

- `checkEntryCount` rejects a collection larger than `maxArrayElements` with `ExecutionLimitError` → exit code 126, stderr `bash: ls: array element limit exceeded (N)`.
- `admitEntries` additionally reserves the batch against the shared `FileTraversalBudget` via `discover()`, the same idiom `find` already uses, so a recursive listing cannot bypass the per-directory bound by walking many individually-small directories.
- The glob listing path applies the same collection bound to the `getAllPaths()` candidate set. Per-path `visit()` continues to handle traversal accounting, so traversal-work accounting is unchanged there.
- The recursive `readdirWithFileTypes` re-read is bounded without double-reserving children already admitted from the plain `readdir` of the same directory.

Also swaps an `O(n^2)` `filteredEntries.includes()` scan in the recursive path for a `Set` lookup — it runs once per entry over the (now bounded, but still potentially large) entry list.

## Acceptance criteria

| Criterion | Status |
| --- | --- |
| Rejects an oversized directory result with exit 126 and a specific stderr message | ✅ |
| Limit checked before per-entry stat/classification work | ✅ (test asserts zero child `stat`/`lstat` calls) |
| `ls -l /large \| head -1` cannot bypass the limit | ✅ |
| Recursive and glob listing paths apply equivalent collection bounds | ✅ |
| Focused regression tests for a filesystem returning more entries than the limit | ✅ `ls.limits.security.test.ts` |

## Not addressed

The issue also asks to "avoid the initial unbounded allocation where filesystem backends can support bounded or incremental directory reads." `IFileSystem` currently exposes only `readdir(path): Promise<string[]>` and the optional `readdirWithFileTypes(path)`; neither is incremental, and no bundled backend supports a bounded read. Doing this properly means adding a new optional streaming/bounded directory API to `IFileSystem` and threading it through `InMemoryFs`, `OverlayFs`, `ReadWriteFs` and `MountableFs` — a public-API design decision I did not want to make unilaterally. The allocation the *backend* performs is therefore still unbounded; what this PR guarantees is that `just-bash` refuses to do any work over it. Happy to follow up with that interface change if you want it.

## Tests

- New: `packages/just-bash/src/commands/ls/ls.limits.security.test.ts` — a synthetic `IFileSystem` whose `readdir()`/`readdirWithFileTypes()` report far more children than the directory holds, covering `ls`, `-l`, `-F`, `-S`, `-R`, `-a`, `| head -1`, both limit messages, the glob path, zero per-entry stat work, and that a directory inside the limits still lists normally.
- `pnpm typecheck`, `pnpm lint:fix`, `pnpm knip`: clean.
- `pnpm test:run --exclude src/spec-tests`: 11582 passed. The 6 failures are pre-existing Python/WASM timeouts, verified identical on a clean checkout of this branch point.
- `pnpm test:run src/spec-tests`: 3834 passed, 1 skipped.

### DIFF
--- a/.changeset/ls-directory-entry-limits.md
+++ b/.changeset/ls-directory-entry-limits.md
@@ -1,0 +1,5 @@
+---
+"just-bash": patch
+---
+
+Bound `ls` directory entry collections before per-entry work. A filesystem backend returning a very large directory could previously drive `ls` into sorting, statting, classifying and formatting every entry before any limit applied, and piping to `head` did not help because pipeline producers are materialized before consumers run. Oversized listings now fail with exit code 126 (`array element limit exceeded` / `filesystem traversal entry limit exceeded`), the recursive descent no longer fans out across sibling directories, and every operand of a multi-directory listing is charged to the same budget.
--- a/packages/just-bash/src/commands/ls/ls.limits.security.test.ts
+++ b/packages/just-bash/src/commands/ls/ls.limits.security.test.ts
@@ -143,4 +143,69 @@ describe("ls directory collection limits", () => {
     expect(result.stderr).toBe("");
     expect(result.exitCode).toBe(0);
   });
+
+  it("counts the synthetic -a entries against the array limit", async () => {
+    const files: Record<string, string> = {};
+    for (let i = 0; i < LIMIT; i++) files[`/d/f${i}`] = "";
+    const bash = new Bash({
+      files,
+      cwd: "/",
+      executionLimits: { maxArrayElements: LIMIT },
+    });
+    // Exactly `LIMIT` real entries, but `-a` prepends "." and ".." for a
+    // listing of LIMIT + 2.
+    const result = await bash.exec("ls -a /d");
+    expect(result.stdout).toBe("");
+    expect(result.stderr).toBe(
+      `bash: ls: array element limit exceeded (${LIMIT})\n`,
+    );
+    expect(result.exitCode).toBe(126);
+  });
+
+  it("charges every operand of a multi-directory listing", async () => {
+    const bash = new Bash({
+      files: { "/a/f1": "", "/a/f2": "", "/b/f1": "", "/b/f2": "" },
+      cwd: "/",
+      executionLimits: { maxTraversalEntries: 5 },
+    });
+    // Two roots plus four children is six entries against a budget of five.
+    const result = await bash.exec("ls /a /b");
+    expect(result.stdout).toBe("");
+    expect(result.stderr).toBe(
+      "bash: ls: filesystem traversal entry limit exceeded (5)\n",
+    );
+    expect(result.exitCode).toBe(126);
+  });
+
+  it("does not materialize sibling directories before admitting them", async () => {
+    let materialized = 0;
+    class ConcurrencySpyFs extends InMemoryFs {
+      override async readdir(path: string): Promise<string[]> {
+        const entries = await super.readdir(path);
+        if (/^\/root\/d\d+$/.test(this.resolvePath("/", path))) {
+          // Yield across microtasks so a batched fan-out would overlap here.
+          for (let i = 0; i < 3; i++) await Promise.resolve();
+          materialized += entries.length;
+        }
+        return entries;
+      }
+    }
+    const files: Record<string, string> = {};
+    for (let d = 0; d < 20; d++) {
+      for (let f = 0; f < 10; f++) files[`/root/d${d}/f${f}`] = "";
+    }
+    const bash = new Bash({
+      fs: new ConcurrencySpyFs(files),
+      cwd: "/",
+      executionLimits: { maxTraversalEntries: 30 },
+    });
+    const result = await bash.exec("ls -R /root");
+    expect(result.stderr).toBe(
+      "bash: ls: filesystem traversal entry limit exceeded (30)\n",
+    );
+    expect(result.exitCode).toBe(126);
+    // Recursion stops at the first subdirectory that overruns the budget
+    // instead of reading a whole batch of siblings up front.
+    expect(materialized).toBe(10);
+  });
 });
--- a/packages/just-bash/src/commands/ls/ls.limits.security.test.ts
+++ b/packages/just-bash/src/commands/ls/ls.limits.security.test.ts
@@ -1,0 +1,146 @@
+import { describe, expect, it } from "vitest";
+import { Bash } from "../../Bash.js";
+import { InMemoryFs } from "../../fs/in-memory-fs/in-memory-fs.js";
+import type { DirentEntry, FsStat } from "../../fs/interface.js";
+
+/**
+ * A backend whose `readdir()` reports far more children than the directory
+ * actually holds. Stands in for a host-provided `IFileSystem` that fronts a
+ * directory too large to materialize, which `ls` must refuse before it sorts,
+ * stats, classifies or formats the entries.
+ */
+class HugeDirectoryFs extends InMemoryFs {
+  /** Paths under /large that were statted, i.e. per-entry work that ran. */
+  readonly childStats: string[] = [];
+
+  constructor(private readonly entryCount: number) {
+    super({ "/large/.keep": "" });
+  }
+
+  private syntheticEntries(): string[] {
+    const entries: string[] = [];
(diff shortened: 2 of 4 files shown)


trieloff:
Thanks — I reproduced all three findings before fixing them. Pushed as e44c44a3.

**1. Multi-operand undercount** (`ls.ts:502`) — confirmed. `ls /a /b` with two files each traversed six entries under `maxTraversalEntries: 5` and exited `0`. As both reviewers noted, `visit()` and `discover()` keep separate counters and `FileTraversalBudget` pre-reserves only one root. Every operand past the first is now reserved against the same counter its children are charged to.

**2. `-a` boundary overshoot** (`ls.ts:531`) — confirmed. A directory with exactly `maxArrayElements: 8` entries listed 10 lines at exit `0`. The synthetic `.` and `..` are now charged up front, before the array is built.

**3. Recursive fan-out** (`ls.ts:698`) — confirmed, and this was the significant one. Instrumenting `readdir` with microtask yields, `ls -R` over 20 sibling directories of 10 files each materialized **200 entries against a budget of 30** before the limit tripped. With `DEFAULT_BATCH_SIZE = 100` that scales to a ~100× multiplier on the bound this PR exists to enforce.

Fixed by descending subdirectories one at a time. Entries *within* a single directory are still statted in parallel batches — that work runs over an already-admitted, bounded list, so it does not multiply anything. Output ordering is unchanged because the results were already sorted by name after collection. The tradeoff is real and worth stating plainly: recursive listings lose cross-directory I/O overlap on async backends. I took that over a bound that is only advisory at width, but I am happy to revisit if you would rather keep the fan-out and accept a documented `batch × maxArrayElements` peak.

There is no way to keep both properties with the current `IFileSystem`: bounding peak memory at `maxArrayElements` requires either concurrency of 1 or a per-child limit of `maxArrayElements / concurrency`, and the latter silently redefines the configured limit.

Also added the missing changeset.

### Unrelated pre-existing issue found while investigating

My first attempt at a concurrency probe used a `setTimeout` in `readdir`, which resolves after the command's execution context is revoked. In that case all 20 in-flight children caught a `SecurityViolationError` and `listPath`'s catch-all downgraded it to `ls: <path>: No such file or directory` with exit `2` — the security error is swallowed. I verified this reproduces identically on `main` without any of my changes, so it is pre-existing and I left it out of this PR. Worth a separate issue if you agree it is a problem; `ls` only re-raises `ExecutionLimitError` and `ExecutionAbortedError`, so any other error class from a revoked context is reported as a missing file.

### Verification

- `pnpm typecheck`, `pnpm lint:fix`, `pnpm knip`: clean.
- `pnpm test:run --exclude src/spec-tests`: 11585 passed. The 6 failures are pre-existing Python/WASM timeouts, verified identical on a clean checkout of this branch point.
- `pnpm test:run src/spec-tests`: 3834 passed, 1 skipped.
- Three new regression tests, one per finding, including an assertion that recursion materializes exactly one subdirectory's entries before failing.

The signing warning in the review table (`0/1 verified`) was against the pre-amend head; both commits now report `verified: true` from the API.

cramforce:
@auto-maintain review

cramforce (approved):

cramforce:
Please rebase

trieloff:
Rebased onto `main` (`c9bd93eb`). No longer conflicting.

This was not a mechanical rebase — #363 landed a substantial `ls` rewrite in between, so here is what moved and why.

**The glob bound is gone, because the code it guarded is gone.** My original commit bounded the `getAllPaths()` candidate set inside `listGlob`. #363 removed `listGlob` entirely and resolves operands literally, which deletes that attack surface rather than bounding it. The corresponding test now covers the same ground through shell pathname expansion (`ls /larg*`), where the shell hands `ls` a directory operand and the readdir bound applies.

**The multi-operand fix moved into the new partition loop.** #363 charges every operand with `traversalBudget.visit(0)`, which is close to the review finding but not the same counter — `visit()` and `discover()` are tracked separately, so charging an operand as a visit does not reserve it as a root whose children are about to be admitted. I re-verified the gap survives #363: `ls /a /b` with two files each still passed a `maxTraversalEntries: 5` budget. Directory operands past the first are now reserved explicitly where they are partitioned.

**One real semantic clash, worth a look.** #363 added `bounds the metadata reads -t performs to sort a directory`, which asserts that a plain name-order `ls` of 40 entries stays under `maxTraversalWork: 8` — the point being that `-t` pays for metadata and plain listing does not. My admission used `FileTraversalBudget.discover()`, which charges one work unit per discovered entry, so it broke that test.

Updating the test would have been the wrong fix: it would silently redefine `maxTraversalWork` so that every plain `ls` costs work proportional to directory size, for anyone who has tuned that budget. The two limits mean different things — entry ceilings bound memory, work bounds filesystem operations — and reading a directory is one operation however many names come back.

So `discover()` is now split:

- `reserve(count)` — charges the entry ceiling only.
- `discover(count)` — reserves *and* charges the work, unchanged behaviour.

`ls` admits entries with `reserve()`; the single `checkpoint()` for the readdir itself already accounts for the operation. `find` keeps `discover()`, which is correct there because every discovered child does become a work item that gets visited later. Both `-t`/`-S` bounding from #363 and the collection bounds from this PR now hold at once.

### Verification after rebase

- `pnpm typecheck`, `pnpm lint:fix`, `pnpm knip`: clean.
- `pnpm test:run --exclude src/spec-tests`: 11640 passed. Same 6 pre-existing Python/WASM timeouts, unrelated and present on a clean checkout.
- `pnpm test:run src/spec-tests`: 3834 passed, 1 skipped.
- All of #363's new `ls` tests pass alongside this PR's, including the `-t` work-budget test above.

Both commits are signed and report `verified: true`. Commit messages were updated so they no longer describe the removed glob path.

trieloff:
@cramforce quick reminder: this has been approved and green for four days.

PR #392, opened 2026-08-28
feat(mktemp): implement mktemp

mutewinter:
## Problem

`mktemp` is not implemented, and `which mktemp` reports it as missing, so agents fall back to writing scratch files at a literal path:

```bash
cat > /tmp/data.json << 'EOF'
{"a": 1}
EOF
process-it /tmp/data.json
```

That works in the default filesystem and breaks in a mounted one. In our host the mount layout has no `/tmp`, so the write fails and the agent has no sanctioned alternative to reach for. Across 291 real agent sessions, `/tmp` appeared in 25 bash calls in 4 tasks, always as a hand-rolled substitute for this command.

It bites an agent harder than a human. A human reads the failure and picks another directory; a model reaches for the one idiom it knows for "somewhere safe to put a scratch file", finds it absent, and reconstructs it badly.

## Cause

No implementation, and no entry in `registry.ts`.

## Fix

`src/commands/mktemp/mktemp.ts`, registered lazily like the rest. Creates the file or directory before printing the path, as GNU does, since the guarantee is that the name is yours and not merely that it was free a moment ago.

Defaults to `$TMPDIR` and falls back to `/tmp`, which is the part that matters for embedders: a host with its own layout points `TMPDIR` at a directory inside it and the standard idiom starts working, with no prompt text steering the model around the gap.

Supports `-d`, `-u`, `-q`, `-p DIR` (inline `-pDIR` and cluster-terminating `-dp DIR`), `--tmpdir[=DIR]`, and `-t`. Unknown options error through `unknownOption` rather than being ignored.

## Scope

`Unchanged:` every existing command and the registry's ordering. Names come from `Math.random()`, the same source `$RANDOM` uses in the interpreter, so the command stays browser-safe; exclusivity comes from the existence check plus retry, not from the entropy.

`Not addressed, deliberately:` `mktemp -p` into a directory that does not exist succeeds here and prints a path inside it. Real `mktemp` fails. I matched the filesystem rather than GNU, because `touch /nope/x` and `echo x > /nope/x` both succeed in this shell too, and a single command diverging from the two beside it is the more surprising outcome. There is a test asserting this with the reasoning attached. Say the word and I will flip it to a parent-directory check instead.

## Tests

19 unit tests in `mktemp.test.ts`: the default template shape, uniqueness across calls, `$TMPDIR`, each option including the three `-p` spellings, bare-versus-slashed template resolution, replacing only the trailing run of X's, and the four diagnostics.

What they cannot prove: the create-failure branch is unreachable with the default filesystem, since it accepts every write, so that error path is written but never exercised. A host filesystem that rejects a write would hit it.

Full suite `pnpm test:run --exclude src/spec-tests`: 583 files, 11,590 passed, 97 skipped, 0 failed, plus 72 in `just-bash-executor`. `pnpm typecheck`, `biome check`, and `knip` clean. `readme.test.ts` requires every registered command to appear in `README.md` and `AGENTS.npm.md`; both updated, which is what turned that from 2 failures into 0. No pre-existing failures to report.

Changeset included as a `minor`.

---

Authored with Claude Opus 5

### DIFF
--- a/.changeset/brave-crabs-shave.md
+++ b/.changeset/brave-crabs-shave.md
@@ -1,0 +1,5 @@
+---
+"just-bash": minor
+---
+
+Add `mktemp`. Creates a file or directory from a template ending in at least three X's, prints the path, and defaults to `$TMPDIR` or `/tmp`. Supports `-d`, `-u`, `-q`, `-p DIR`, `--tmpdir[=DIR]`, and `-t`.
--- a/packages/just-bash/AGENTS.npm.md
+++ b/packages/just-bash/AGENTS.npm.md
@@ -78,7 +78,7 @@ const result = await bash.exec("cat input.txt | grep pattern");
 
 **Data processing**: `jq` (JSON), `js-exec` (JavaScript/TypeScript via QuickJS), `python3`/`python` (Python via WASM/CPython), `sqlite3` (SQLite), `xan` (CSV), `yq` (YAML/XML/TOML/CSV)
 
-**File operations**: `basename`, `chmod`, `cp`, `dirname`, `du`, `file`, `find`, `ln`, `ls`, `mkdir`, `mv`, `od`, `pwd`, `readlink`, `rm`, `rmdir`, `split`, `stat`, `touch`, `tree`
+**File operations**: `basename`, `chmod`, `cp`, `dirname`, `du`, `file`, `find`, `ln`, `ls`, `mkdir`, `mktemp`, `mv`, `od`, `pwd`, `readlink`, `rm`, `rmdir`, `split`, `stat`, `touch`, `tree`
 
 **Utilities**: `alias`, `base64`, `bash`, `clear`, `curl`, `date`, `diff`, `echo`, `env`, `expr`, `false`, `gzip`, `gunzip`, `help`, `history`, `hostname`, `html-to-markdown`, `md5sum`, `printenv`, `printf`, `seq`, `sh`, `sha1sum`, `sha256sum`, `sleep`, `tar`, `tee`, `time`, `timeout`, `true`, `unalias`, `which`, `whoami`, `zcat`
 
--- a/packages/just-bash/README.md
+++ b/packages/just-bash/README.md
@@ -82,7 +82,7 @@ duplicating internal defaults.
 
 ### File Operations
 
-`cat`, `cp`, `file`, `ln`, `ls`, `mkdir`, `mv`, `readlink`, `rm`, `rmdir`, `split`, `stat`, `touch`, `tree`
+`cat`, `cp`, `file`, `ln`, `ls`, `mkdir`, `mktemp`, `mv`, `readlink`, `rm`, `rmdir`, `split`, `stat`, `touch`, `tree`
 
 ### Text Processing
 
--- a/packages/just-bash/src/commands/mktemp/mktemp.test.ts
+++ b/packages/just-bash/src/commands/mktemp/mktemp.test.ts
@@ -27,6 +27,18 @@ describe("mktemp", () => {
       expect(result.stdout.trim()).toMatch(/^\/scratch\/tmp\.[0-9A-Za-z]{10}$/);
       expect(result.exitCode).toBe(0);
     });
+
+    // An empty value is unset, not the root directory. Checked against
+    // coreutils 9.2, which prints /tmp/tmp.XXXXXXXXXX for all three.
+    it.each([
+      "TMPDIR= mktemp -u",
+      "mktemp -u --tmpdir=",
+      "mktemp -u -p ''",
+    ])("should treat an empty temp dir in `%s` as unset", async (command) => {
+      const env = new Bash();
+      const result = await env.exec(command);
+      expect(result.stdout.trim()).toMatch(NAME);
+    });
   });
 
   describe("options", () => {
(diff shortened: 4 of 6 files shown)


mutewinter:
Good review. I checked each behavioral claim against coreutils 9.2 rather than the manual, and all of them hold. Pushed 4bd61ec.

**Fixed**

- **Modes.** Files are 0600 and directories 0700 now, via `chmod` after creation. GNU: `-rw-------` and `drwx------`, confirmed locally.
- **Empty temp dir resolving to root.** `??` only guarded null, so `TMPDIR= mktemp`, `--tmpdir=`, and `-p ''` all produced `/tmp.XXXXXXXXXX` at the filesystem root. GNU treats all three as unset and uses `/tmp`; so does this now, with a test for each spelling.
- **`-q` over-suppressing.** GNU's `-q` covers creation failure only, so `mktemp -q bad.XX` still prints `too few X's`. My test had locked in the wrong behavior; it now asserts the right one.
- **Predictable names.** `crypto.getRandomValues` instead of `Math.random()`. You are right that the browser-safety rationale did not hold, and I have dropped that claim from the comment. The 62-symbol alphabet does not divide 256, so bytes at or above the largest whole multiple are drawn again rather than folded.
- **`--help`.** Wired through `hasHelpFlag`/`showHelp` like the other commands, with usage, options, examples, and a note about the modes.
- **Aborting instead of retrying.** A creation error now continues the loop and is reported once the attempts are spent.

**Partly fixed, and the reason**

Exclusive creation. The directory case is now genuinely exclusive: `mkdir` without `recursive` already refuses an existing path, so the pre-check is gone and a collision is a caught throw.

The file case cannot be closed inside this command. `IFileSystem.writeFile` has no exclusive mode and `WriteFileOptions` is `{ encoding }`, so there is no `O_EXCL` equivalent to reach for, and `writeFile` truncates rather than failing. Rather than keep the claim, the docstring now states the gap: `exists` then `writeFile` is check-then-write, another writer on the same backing store can take the name in between, and the mode is applied after creation so the file is briefly 0644.

What closes it is an `exclusive?: boolean` on `WriteFileOptions` honored by each `IFileSystem` implementation. That is a change to the filesystem interface and every implementation of it, which felt like the wrong thing to smuggle into a command PR. Happy to open it as a follow-up, or to fold it in here if you would rather see them land together. Either way the unpredictable names now mean an attacker has to win a race they cannot aim at, rather than one they can.

One thing I deliberately did not change, flagged in the original description: `mktemp -p /nope/missing` succeeds here and prints a path inside a directory that does not exist. Real `mktemp` fails. `touch /nope/x` and `echo x > /nope/x` both succeed in this shell too, so I matched the filesystem rather than GNU in isolation. Still happy to flip it to a parent-directory check if you would rather it match GNU.

Suite after the changes: 583 files, 11,596 passed, 97 skipped, 0 failed, plus 72 in `just-bash-executor`. `typecheck`, `biome check`, and `knip` clean.

cramforce:
@auto-maintain review

Please fix the lint errors!

mutewinter:
Lint is fixed in 9d41298b, so the head is current.

`MAX_ATTEMPTS` tripped the command-local MAX rule. I annotated rather than promoting it to `ExecutionLimits`, because it is not a ceiling a host would configure: it bounds retries against name collisions, each attempt is one `exists` check on a name drawn from 62^n, and raising it buys nothing while lowering it only makes the command fail on a directory it could have used. Happy to move it into `ExecutionLimits` instead if you would rather every bound live in one schema.

On the atomicity blocker, which is the real question here: I agree with the diagnosis and I do not think it should be fixed inside this PR. Here is the shape of it, so the call is easy to make.

`writeFile` has no exclusive mode anywhere in `IFileSystem`, so this needs `exclusive?: boolean` on `WriteFileOptions` and an implementation in each of the four filesystems:

- `InMemoryFs` and `OverlayFs` — a map check inside the same synchronous step. Trivial, and not where the race lives, since nothing else writes to those.
- `MountableFs` — pass-through to the routed filesystem.
- `ReadWriteFs` — the only one that can actually race, because it is the only one backed by a real directory another process can write to. That means `fs.promises.open(path, "wx")` and writing through the handle, which replaces the `mkdir` + re-validate + `writeFile` sequence in `writeFileUnlocked` that currently carries the symlink revalidation and the containment check.

That last one is a change to the write path of the class that enforces filesystem isolation, landing on top of #368. It wants to be reviewed as a filesystem change on its own terms, not as a detail at the bottom of a command PR, and if it is wrong the blast radius is every write just-bash makes rather than one command.

So my suggestion is: land the interface change first as its own PR, then this one drops the `exists` check and passes `exclusive: true`. I am happy to write it. Say the word and I will open it, or if you would rather see them together I will fold it in here instead.

Two things worth noting while it is open. Nothing regresses by merging as-is: today `mktemp` does not exist, so there is no worse behavior to fall back to, and the names are now drawn from `crypto.getRandomValues`, so the window cannot be aimed at even though it is there. And the directory half is already exclusive, since `mkdir` without `recursive` refuses an existing path.

Suite at this head: 583 files, 11,596 passed, 97 skipped, 0 failed, plus 72 in `just-bash-executor`. `lint`, `typecheck`, `biome check`, and `knip` all clean.

@auto-maintain review

cramforce:
@auto-maintain review

mutewinter:
Closing in favor of #377, which was open a week before this one and already carries what the review here asked for: `createExclusive` on `IFileSystem` with an implementation in all four backends, mode applied at creation, and GNU comparison fixtures. It also defaults to `$TMPDIR` then `/tmp`, which is the part an embedder needs. Two `mktemp` PRs in the queue only make the choice harder, and #377 is the better one.

The branch stays up. The one behavior I argued for here and #377 decides the other way, `-p` into a directory that does not exist, is GNU's call and fine.

PR #394, opened 2026-08-28
refactor(js-exec): port JavaScript runtime to run

cramforce:
## Summary

- replace js-exec custom QuickJS worker with run@2.1.0
- map filesystem, stdout/stderr, fetch, command execution, and tools to synchronous host bindings with live read-after-write behavior
- use run native static and dynamic module loading instead of regex source and import transforms
- preserve serialized js-exec admission while allowing queued work to continue after timeout teardown
- remove the bespoke worker artifact, snapshot/replay machinery, and worker-specific initialization tests

## Safety and compatibility

- forward the active host-function abort signal into fetch, command execution, and tools
- keep bridge payloads below run synchronous transport capacity while preserving the 64 MiB QuickJS heap limit
- preflight filesystem read sizes and use bounded base64 transport without host-side byte-array amplification
- map maxWorkerMessageBytes into run source/result limits and keep combined returned output within maxOutputSize
- preserve descriptive sanitized module-loader errors through synthetic throwing modules
- prevent bridge side effects after process.exit, including when guest code catches its exception
- validate guest byte arrays before allocation and use a per-invocation host namespace
- account for output incrementally, redact tool-originated host paths, and avoid duplicate fetch-body transport
- preserve zero bridge limits, root-relative module resolution, and separate bootstrap lexical scope
- support Node 20.19+ and externalize run from Node bundles
- allow-list run from the minimum release-age policy because it is maintained by the same owners
- include a patch changeset for just-bash

## Isolation decision

QuickJS/WASM and run's validated, bounded serialization and bridge protocols are the js-exec sandbox boundaries. Guest JavaScript does not execute in the Node worker realm, so this migration intentionally does not reproduce just-bash's WorkerDefenseInDepth inside run. The README and threat model state that explicitly.

## Remaining upstream run limitations

- run@2.1.0 only accepts finite timeout values up to 2,147,483,647 ms, so truly infinite and longer js-exec deadlines need an upstream API addition
- run currently applies its TypeScript stripping/transformation path to every source; restoring extension/flag-gated transformation needs an upstream option
- TypeScript remains a production dependency so advertised TypeScript execution works on Node 20

## Validation

- 297 focused js-exec and adversarial security tests
- 17 distribution and package-content tests
- package build, including Node ESM/CJS, browser, CLI, shell, and worker artifacts
- TypeScript checks
- formatting and banned-pattern checks
- knip
- prior broad validation: 15,349 unit/security tests and 672 WASM tests


aayush-kapoor (approved):

cramforce:
Addressed the latest auto-maintain review in 9d8d910:

- process.exit now makes every later bridge operation side-effect-free, even if guest code catches the exception
- bootstrap code runs in a separate lexical scope
- modules at / resolve relative imports from /
- maxJsBridgeRequests: 0 is no longer raised to 1
- tool failures use host-path redaction
- filesystem byte inputs are shape-, value-, and size-validated before allocation
- host namespaces are unique per invocation rather than a stable __host name
- output accounting is incremental instead of repeatedly scanning accumulated output
- fetch responses cross the bridge once
- regression tests cover the correctness and amplification cases

We consciously do not add WorkerDefenseInDepth to run. The guest executes in QuickJS/WASM, not the Node worker realm; QuickJS plus run's validated, bounded bridge is the relevant boundary. The README and THREAT_MODEL now say this explicitly. The run package remains in minimumReleaseAgeExclude because it is owned and released by the same maintainers.

cramforce:
Follow-up review fixes are in 21a0fe5:

- filesystem reads are stat-gated before allocation and cross the bridge as bounded base64, eliminating the host-side Array.from amplification
- combined stdout/stderr remains within maxOutputSize even after the limit diagnostic is added
- maxWorkerMessageBytes now explicitly bounds run source and result payloads
- missing and unreadable modules execute synthetic throwing modules, preserving sanitized actionable diagnostics instead of Host bridge request failed
- binary filesystem reads have exact round-trip coverage

Validated with 297 focused JS/security tests, 17 distribution tests, type-checking, formatting, banned-pattern scanning, and the complete package build.

The remaining timeout ceiling and unconditional TypeScript transformation require run API changes and are now called out explicitly in the PR description. WorkerDefenseInDepth and the owned-package release-age exemption remain conscious decisions.

cramforce:
Addressed the actionable current-head review findings in `dddc254`:

- restored `fs.readFileBuffer()` as an exact `ArrayBuffer`
- projected `exec`/`execArgs` results to `{ stdout, stderr, exitCode }`
- restricted source-location formatting to verified guest errors so runtime/host stack paths are sanitized
- forwarded the active host-function `AbortSignal` to `javascript.invokeTool` and documented the cancellation contract
- added focused compatibility/security regressions

Validated with all 286 js-exec tests, README example tests, typecheck, banned-pattern lint, Biome, and the full package build.

The static named/default-import diagnostic remains an upstream `run` limitation: rejecting from the loader is redacted by the bridge, while a throw-only synthetic module cannot satisfy arbitrary requested exports during linking.

cramforce:
Addressed the valid current-repository findings in `9cfd520`:

- restored dedicated bootstrap-error attribution for runtime and syntax failures in script and module modes
- enabled bundled js-exec distribution coverage on Node 20
- made output-limit failure status take precedence over a later `process.exit(0)`
- stat/reject oversized imported modules before reading their contents, with a post-read race check

Validation: 292 js-exec tests, focused bundle tests, typecheck, banned-pattern lint, Biome, and the full package build all pass locally.

The queue-wedge claim does not match `run` 2.1.0 behavior: its sync dispatch races against invocation cancellation and worker teardown lets the outer queue release. An already-running non-cooperative host promise can continue, but a post-settlement flag cannot cancel that operation; the active AbortSignal is now forwarded. Type-transform control, truly infinite deadlines, and safe loader-error propagation remain upstream `run` API work.

cramforce:
Addressed the refreshed `9cfd520` findings in `8218462`:

- restored top-level-await/dynamic-import module auto-detection
- documented the `invokeTool` AbortSignal in the command README
- restored the historical 8 MiB per-call bridge payload ceiling
- gated `bootstrapError` to the bootstrap phase so guest code cannot forge bootstrap attribution

The raw run-installed binding is non-configurable, so attempting to delete it breaks ESM with `could not delete property`; its functions remain bounded and validated. The phase gate addresses the concrete bootstrap-forgery consequence.

Validated with 294 js-exec tests, typecheck, banned-pattern lint, Biome, and the full package build.

cramforce:
The top-level `await` regex is a temporary compatibility workaround, and I agree it is not a sound parser boundary. I opened https://github.com/vercel-labs/run/pull/61 to remove the underlying coupling.

That patch adds explicit `sourceType: "function-body" | "module"` semantics while allowing `moduleLoader` to remain installed in function-body mode. Once published, just-bash can always install its loader, pass the CLI/extension-derived source type explicitly, and delete top-level-await/dynamic-import source inspection entirely. Existing `run` callers retain the current defaults, so the upstream release remains a patch.

cramforce:
Updated to published `run@2.1.3` in `3eaa7e8`.

The js-exec adapter now always installs its native module loader and passes explicit `sourceType` based only on `-m`/file-extension semantics. The top-level-await regex is gone. Dynamic imports work in function-body mode without sacrificing top-level `await` or `return`. Added regressions for comment-separated `await`, strings containing await-like text, and dynamic imports retaining function-body behavior.

Validation:
- 296/296 js-exec tests pass
- full build and typecheck pass
- lint, workflow security, Biome, and banned-pattern checks pass
- broader unit run: 14,683 pass; four unrelated xz/tar failures caused by the local install skipping its optional native codec build

cramforce:
@auto-maintain review

cramforce:
@auto-maintain review

PR #396, opened 2026-08-30
fix(limits): enforce maxExecutionTimeMs inside data-command row loops

taoche:
### Summary

`maxExecutionTimeMs` cannot bound a single data-processing statement. `ExecutionScope.throwIfAborted` is only reached at
statement boundaries (`chargeCommand` → `consume` → `assertUsable`), so a
`grep`/`jq`/`awk`/`sed` scan over a large input runs to completion and *then*
reports exit 124 — the deadline discards the result after paying its full
wall-clock cost rather than stopping the work.

This adds the deadline check inside the record loops themselves, strided so it
stays off the hot path.

### How we hit it

We embed just-bash as the sandboxed tool surface for an LLM agent in a
production Node service. The agent writes its own pipelines, so an oversized
input or an accidentally quadratic one-liner is routine rather than
exceptional, and `maxExecutionTimeMs` is our backstop for keeping one agent
turn from monopolising the event loop. It turned out not to be one.

Repro on `main` (`c9bd93e`), 300k-row JSONL:

```js
const bash = new Bash({
  files: { "/work/big.jsonl": rows },   // 300k JSON objects
  cwd: "/work",
  executionLimits: { maxExecutionTimeMs: 200 },
});
await bash.exec(`sed 's/name/NAME/g' /work/big.jsonl | tail -1`);
// → exitCode 124, "exceeded execution deadline (200ms)" … after 7.5 seconds
```

Same shape for `jq -s`, `grep -c`, `awk`. We also saw the fully silent case: `wc -l` over the same input with a 50ms
deadline returns **exit 0** after 196ms, because nothing charges the scope
after the command finishes.

### What changed

The record loops now consult the execution scope directly:

| site | file | granularity |
|---|---|---|
| `awk` | `commands/awk/interpreter/interpreter.ts` | per input record |
| `sed` | `commands/sed/sed.ts` | per pattern-space line |
| `jq`, `yq` | `commands/query-engine/evaluator.ts` (`chargeQueryWork`) | per evaluation op |
| `grep`, `rg` | `commands/search-engine/matcher.ts` (`chargeWork`) | per work unit |
| `grep` prefilter miss | `commands/grep/grep.ts` | once per file |

The query evaluator's `try`/`catch`, `?` and comma recovery paths previously
rethrew only `ExecutionLimitError`; they now go through
`rethrowFatalExecutionError`, so a filter cannot convert the deadline abort
into a value and buy itself another stride.

`executionScope` is threaded to the two interpreters that did not already
receive it (awk's runtime context, sed's `processContent` options); `grep` and
`rg` pass `ctx.executionScope` into `SearchOptions.budget`.

Checks are strided by `DEADLINE_CHECK_STRIDE` (1024, exported from
`execution-scope.ts`), so cancellation latency is bounded by 1024 rows rather
than by the length of the whole scan. The stride reuses a counter each loop
already maintains — `recordsProcessed` in awk, `lineIndex` in sed,
`budget.operations` in the query evaluator, a local in the search matcher — so
the added per-row work is one integer compare against a local or an
already-loaded field, **not** a call through the `CommandExecutionBudget`
facade. `QueryEvaluationBudget` gains a `deadlineCheckAt` field for this.

This is deliberately the cheap shape. An earlier revision of this patch called
a new `budget.checkpoint()` method per row instead, and that cost **+28–62% on
`grep`** — the facade call appears to make `chargeWork` non-inlinable and
deoptimise the surrounding loop. Given #218 and #248 went to some trouble on
exactly these loops, the stride is the point, not an optimisation detail.

### Effect

Time to actually stop, 300k-row JSONL, `maxExecutionTimeMs: 50`. (~200ms of
each figure is fixed VFS setup before the command runs, so ~200ms is the floor
this harness can show.)

| command | `main` | this PR |
|---|---|---|
| `sed 's/name/NAME/g'` | 7463ms | **216ms** |
| `rg -c` | 2595ms (exit 126, work limit — not the deadline) | **290ms** |
| `awk '{ t += length($0) }'` | 1100ms | **538ms** |
| `jq -s 'map(.value) \| add'` | 812ms | 703ms |
| `grep -c` | 775ms | **210ms** |

### Performance

No measurable regression. 99k-row JSONL, 7 runs per cell, `main` and this
branch **interleaved over 3 rounds** on the same machine to cancel thermal
drift (an earlier non-interleaved run showed baseline itself drifting ~4%,
which is why the numbers below are averaged across rounds):

| command | `main` min / p50 | this PR min / p50 | delta |
|---|---|---|---|
| `grep -c` | 97.7 / 101.7ms | 98.7 / 102.7ms | +1.0% |
| `jq -s` | 199.0 / 202.7ms | 198.0 / 203.7ms | −0.5% / +0.5% |
| `awk` | 629.7 / 639.0ms | 630.7 / 638.0ms | +0.2% / −0.2% |
| `sed` | 2826 / 2878ms | 2786 / 2836ms | −1.4% / −1.5% |

`grep` is the only consistent signal at +1.0% (99/98/99 vs 97/99/97 across
rounds); the rest are inside run-to-run noise.

### Tests

`src/security/limits/row-loop-deadline.test.ts` — for each of grep/jq/awk/sed
over 99k rows: a 25ms deadline yields exit 124 with `execution deadline` in
stderr and returns in under 2s, and a 600s deadline leaves exit code and
stderr untouched. The second half is the regression guard against a stride
that checks too eagerly. Two more groups on a file-less `Bash`: jq and yq each
stop a heavy `reduce` mid-evaluation with the evaluator's own deadline message
(the yq case fails without the `executionScope` wiring), and jq `try … catch`
and `?` around the same `reduce` still exit 124 with empty stdout.

Full suite: `pnpm lint`, `pnpm knip`, `tsc --noEmit`, and `pnpm test:run`
(15467 passed). One pre-existing failure, `src/package-contents.test.ts`,
which shells out to `npm pack` and is refused by the `packageManager` field —
it fails identically on unmodified `main` in this environment.

### Not covered

Deliberately out of scope, happy to fold any of these in if you'd prefer them
together:

- **`ExecOptions.signal`.** The loops are synchronous and never yield, so an
  `AbortController` fired from a host timer is still only observed once the
  command returns. Only the wall-clock deadline — a `Date.now()` poll — can
  fire mid-scan. Making `signal` interrupt a scan means yielding to the event
  loop inside these loops, which is an async refactor of sed/awk rather than a
  checkpoint.

- **Input ingestion.** The row loops are bounded but the phase that *builds*
  the rows is not — `jq -s` parsing 300k JSON documents and awk's record
  split both run before any row loop, which is why those two improve least
  above.
- **Sub-record pathologies.** One awk rule with a long inner `while`, or one
  catastrophic regex, is still a single uninterruptible unit.
- **`sort`.** The comparator runs inside `Array.prototype.sort`; that is the
  hottest path in the repo and I did not want to put a branch in it without
  discussing the tradeoff first.
- **Other whole-content passes** measured still overrunning at 50ms:
  `xan slice` 3303ms, `tr` 1588ms, `rev` 592ms. Each is a single pass with no
  row loop to hook, so each needs its own chunked check (~5–10 LOC apiece).

If you would rather see the deadline become a documented
best-effort/statement-boundary guarantee than have these checks added, that is
a reasonable call too — but the current docs for `signal` read as a stronger
promise than the implementation makes.

### DIFF
--- a/.changeset/row-loop-deadlines.md
+++ b/.changeset/row-loop-deadlines.md
@@ -1,0 +1,30 @@
+---
+"just-bash": patch
+---
+
+Enforce `maxExecutionTimeMs` and `ExecOptions.signal` inside the data-command row loops
+
+`ExecutionScope.throwIfAborted` was only reached at statement boundaries
+(`chargeCommand` → `consume` → `assertUsable`), so the deadline could not bound a
+single long-running statement. A `grep`/`jq`/`awk`/`sed` scan over a large input
+ran to completion and *then* reported exit 124, having already spent the full
+wall-clock cost. Because the builtins are synchronous, the host's own
+`AbortController` timer could not fire during the scan either, so an external
+`signal` had no observable effect until the interpreter came back up for air.
+
+The record loops now consult the scope directly: `awk` per input record, `sed`
+per pattern-space line, the query evaluator (`jq`, `yq`) in `chargeQueryWork`,
+and the search engine (`grep`, `rg`) in `chargeWork`. Cancellation latency is
+bounded by `DEADLINE_CHECK_STRIDE` (1024) rows rather than by the length of the
+whole scan.
+
+The check is strided off the hot path — reusing a counter each loop already
+maintains, so the added work per row is one integer compare against a local or
+already-loaded field, not a call through the `CommandExecutionBudget` facade. On
+a 99k-row JSONL input, `grep -c`, `jq -s 'map(.value) | add'`, `awk '{ t +=
+length($0) }'` and `sed s///g` are unchanged within measurement noise.
+
+**Residual gap**: a single record whose own evaluation is unbounded (an `awk`
+rule with a long inner `while`, one pathological regex) is still not
+interruptible, and `sort` compares inside `Array.prototype.sort` remain outside
+the deadline. Those need per-construct checks rather than per-row ones.
--- a/.changeset/row-loop-deadlines.md
+++ b/.changeset/row-loop-deadlines.md
@@ -2,21 +2,21 @@
 "just-bash": patch
 ---
 
-Enforce `maxExecutionTimeMs` and `ExecOptions.signal` inside the data-command row loops
+Enforce `maxExecutionTimeMs` inside the data-command row loops
 
 `ExecutionScope.throwIfAborted` was only reached at statement boundaries
 (`chargeCommand` → `consume` → `assertUsable`), so the deadline could not bound a
 single long-running statement. A `grep`/`jq`/`awk`/`sed` scan over a large input
 ran to completion and *then* reported exit 124, having already spent the full
-wall-clock cost. Because the builtins are synchronous, the host's own
-`AbortController` timer could not fire during the scan either, so an external
-`signal` had no observable effect until the interpreter came back up for air.
+wall-clock cost.
 
 The record loops now consult the scope directly: `awk` per input record, `sed`
 per pattern-space line, the query evaluator (`jq`, `yq`) in `chargeQueryWork`,
-and the search engine (`grep`, `rg`) in `chargeWork`. Cancellation latency is
-bounded by `DEADLINE_CHECK_STRIDE` (1024) rows rather than by the length of the
-whole scan.
+and the search engine (`grep`, `rg`) in `chargeWork`, plus `grep`'s whole-file
+prefilter on a miss. Cancellation latency is bounded by `DEADLINE_CHECK_STRIDE`
+(1024) rows rather than by the length of the whole scan. The query evaluator's
+`try`/`catch`, `?` and comma recovery paths now rethrow the deadline abort along
+with the other fatal execution errors instead of converting it to a value.
 
 The check is strided off the hot path — reusing a counter each loop already
 maintains, so the added work per row is one integer compare against a local or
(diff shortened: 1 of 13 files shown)


taoche:
Follow-up in c77b09a addressing the review findings:

- **yq**: `evalOptions` now passes `executionScope: ctx.executionScope`, matching jq. Added to the deadline tests; the yq case fails without the wiring.
- **jq `try`/`?`**: the evaluator's `Try`, `Optional` and comma recovery catch sites rethrew only `ExecutionLimitError`, so a filter could swallow the `ExecutionAbortedError` raised by the deadline. They now go through `rethrowFatalExecutionError`, as grep does. Tests cover `try … catch` and `?` around a heavy `reduce`: exit 124, empty stdout, evaluator deadline message.
- **grep prefilter**: a whole-file prefilter miss now consults the deadline before returning, so a large no-match scan reports 124 rather than exit 0 after overrunning.
- **`signal`**: dropped from the title, body and changeset. The loops never yield, so a timer-driven `abort()` is still only observed once the command returns — only the wall-clock `Date.now()` poll fires mid-scan. Listed under "Not covered" with why (it's an async refactor of sed/awk, not a checkpoint).

Perf-sensitive paths are untouched by this commit: the evaluator change is in catch blocks, the grep check is once per file on the miss path.

cramforce:
@auto-maintain review

PR #398, opened 2026-08-31
fix(fs): require existing destination parents for cp and mv

skovranek:
## Summary

  - Stop `cp` and `mv` from implicitly creating missing destination parent directories.
  - Enforce the destination-parent invariant across `InMemoryFs`, `OverlayFs`, `ReadWriteFs`, and same- and cross-filesystem `MountableFs` operations.
  - Validate destination components before resolving `..`, matching standard filesystem traversal behavior.
  - Preflight destination parents in the command layer so custom `IFileSystem` implementations receive correct `ENOENT` and `ENOTDIR` diagnostics.
  - Preserve synthetic `MountableFs` parent directories when delegating operations to an underlying filesystem.
  - Preserve recursive `cp` behavior that creates the final destination directory when its parent already exists.
  - Add filesystem contract tests, command regressions, mount-namespace tests, and real-Bash comparison coverage.

  ## Context

  Standard `cp` and `mv` utilities fail when an intermediate destination directory does not exist. Previously, just-bash created that hierarchy automatically, which could make invalid
  shell commands succeed and, for `mv`, remove the source.

  Path normalization also allowed commands such as `cp source missing/../copied` to bypass the missing component. Destination errors from custom filesystems could consequently be
  attributed to the source, while synthetic `MountableFs` parents could be rejected by the underlying filesystem despite existing in the wrapper’s virtual namespace.

  This change validates the unnormalized destination traversal, standardizes command diagnostics through a filesystem preflight, and preserves valid synthetic mount parents during
  delegation.

  ## Testing

  - Focused destination-parent and mount regressions: 127 passed.
  - Filesystem, `cp`, and `mv` suites: 1,264 passed, 1 skipped.
  - Real-Bash file-operation comparisons: 23 passed.
  - Build, typecheck, `knip`, changeset validation, and worker synchronization passed.

### DIFF
--- a/.changeset/strict-destination-parents.md
+++ b/.changeset/strict-destination-parents.md
@@ -1,0 +1,5 @@
+---
+"just-bash": patch
+---
+
+Require `cp` and `mv` destination parent directories to exist, matching standard Unix utility behavior.
--- a/packages/just-bash/src/commands/cp/cp.destination-parent.test.ts
+++ b/packages/just-bash/src/commands/cp/cp.destination-parent.test.ts
@@ -1,0 +1,118 @@
+import { describe, expect, it } from "vitest";
+import { Bash } from "../../Bash.js";
+import { InMemoryFs } from "../../fs/in-memory-fs/in-memory-fs.js";
+
+class PlainDestinationErrorFs extends InMemoryFs {
+  override async cp(): Promise<void> {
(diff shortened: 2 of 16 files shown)


cramforce:
Please see bot feedback

skovranek:
> Please see bot feedback

@cramforce thank you, ready for review 👍

PR #400, opened 2026-08-31
fix(interpreter): give a bare assignment exit status 0, not the previous command's

trieloff:
fix(interpreter): give a bare assignment exit status 0, not the previous command's

## Problem

```bash
set -e
if false; then :; else x=1; fi
echo done
```

Nothing is printed and the shell exits 1. There is no diagnostic, and no command in the script failed.

Found by an agent driving just-bash inside SLICC, where it cost an 18-step bisection. The failure has no surface a script can inspect: no stderr, no non-zero from any command that ran, no partial output. Invoked as a subprocess it reads as a subcommand that "silently did nothing," which looks like argument parsing, an auth failure or a hung network call long before it looks like an exit-status bug in the shell.

Three things made it expensive to find, all of them now test cases:

- **Any statement after the assignment masks it.** `else x=1; echo "x=$x"; fi` is fine; `else x=1; fi` is not. A real script fails or not depending on where the assignment happens to fall, so one subcommand breaks and its neighbour doesn't.
- **`local`, `export` and `declare` are immune.** They are real command words and report their own status, so only a bare assignment is affected. Re-declaring with `local` was the workaround SLICC shipped, correct but for reasons nobody had at the time.
- **The `if`/`else` framing is a red herring.** The bisection landed on "assignment followed by another statement inside `if`/`else`, but not `case`", which is right about the symptom and wrong about the cause.

The production shape was a defaults helper, where the assignment is the natural last statement:

```bash
set -e
cfg() { if [ -n "$1" ]; then val="$1"; else val="fallback"; fi; }
cfg ""
echo "val=$val"      # never runs — and val was set correctly
```

## Cause

A simple command with no command word returned `ctx.state.lastExitCode`:

```ts
// Assignment-only command: preserve the exit code from command substitution
// e.g., x=$(false) should set $? to 1, not 0
return result("", stderrOutput, this.ctx.state.lastExitCode);
```

The comment describes the case it was written for. `x=$(false)` works because the substitution writes `lastExitCode` on its way through, and reading it back afterwards recovers 1. But when no substitution runs there is nothing to read back, and `lastExitCode` still holds whatever the *previous* command left there. Bash gives a bare assignment status 0.

No control flow is needed to see it:

```bash
false; x=1; echo $?        # 1, bash says 0
false; x=1; y=2; echo $?   # 1 — the last assignment does not clear it either
( false; x=1 ); echo $?    # 1 — same inside a subshell
```

An `else` branch is where it turns fatal, because it is entered with `$?` set to 1 by the condition that just failed. The assignment reports 1, the branch's last statement decides the `if`'s status, the `if` reports failure, and `set -e` turns that into an exit. `then` branches hid it because a true condition leaves `$?` at 0, and so did `case`, which is entered carrying whatever status preceded it.

The same read appears in a neighbouring path with the same consequence: a command word that expands to nothing (`$empty`) inherited the previous failure.

## Fix

Separate "the status a substitution set" from "the status of the last command". Command substitutions record through one helper:

```ts
export function recordSubstitutionExit(state: InterpreterState, exitCode: number): void {
  state.lastExitCode = exitCode;
  state.env.set("?", String(exitCode));
  state.lastSubstitutionExitCode = exitCode;
}
```

Every simple command clears the marker before expanding, and a command with no command word reports the last substitution performed while expanding it — assigned values first, then redirection words in written order — or 0 when none ran.

```bash
x=$(exit 7)                                  # 7
x=$(exit 3) y=$(exit 4)                      # 4, last wins
x=$(exit 3) y=plain                          # 3, a plain value does not clear it
x=${y:=$(exit 9)}                            # 9, nested in a parameter expansion
> /dev/null$(exit 5)                         # 5, redirection words count
x=$(exit 7) > /dev/null$(exit 5)             # 5
> /dev/null$(exit 5) x=$(exit 7)             # 5, assignments expand first regardless of order
x=1                                          # 0
```

Three details follow bash rather than convenience:

- **A redirection onto fd 0 reports 0 regardless.** bash performs a null command's redirections in a forked child when one of them reads stdin, so the substitution status is discarded: `x=$(exit 7) < /dev/null` is 0, `x=$(exit 7) <<< x` is 0, `x=$(exit 7) 0<&-` is 0, while `x=$(exit 7) 3< /dev/null` and `x=$(exit 7) 3<&0` are 7. This follows bash 5.x; bash 3.2 predates the fork and reports 7 for all of them, and the comparison fixtures are recorded against 5.x.
- **Process substitution** restores the marker alongside `$?`, which it already saved. `false; x=<(echo hi)` is 0 in bash, and the marker must not survive the body any more than `$?` does.
- **`PS4`** is expanded for the trace line only, so a substitution in it is not one of the command's own expansions. `$?` and the marker are restored after prompt expansion, next to the `xtrace` flag that was already saved there.

Six call sites in `expansion.ts` and `expansion/pattern-expansion.ts` already wrote the `lastExitCode` / `env["?"]` pair by hand; they now call the helper.

## Scope

Unchanged: every command that has a command word, including prefix assignments on a real command (`FOO=$(exit 5) true` is 0, `true`'s status) and redirection targets on one (`echo hi > /dev/null$(exit 5)` is 0). `local`, `export` and `declare` were never affected and are now pinned as agreeing with the bare-assignment path rather than being right by accident.

Not addressed, deliberately:

- `x=$"hi"` (locale translation quoting) keeps its trailing quote. Unrelated, found while fuzzing this area.
- A diagnostic when `set -e` aborts on a construct containing no failing command. That silence is precisely what bash does, and matching bash is this project's rule, so a warning would be a deliberate divergence and belongs in its own issue. `set -x` does show it — the trace ends at the assignment — though that is thin help when the symptom surfaces three layers up in a subprocess.

## Review round

Two findings, both real, both fixed in `f06ad146`.

**The redirection rule was backwards, and my justification for it was a bad probe.** I had snapshotted the marker *before* redirection expansion, on the grounds that `> $(sh -c "exit 5"; echo f)` is 0 in bash. That probe proves nothing: the substitution's own status is 0 because `echo` is its last command, so `exit 5` never reaches the status. It said "this substitution succeeded", not "targets are ignored". With a substitution that actually exits nonzero bash counts it, and the snapshot had regressed `$empty > /dev/null$(exit 5)` from 5 to 0 — the same swallowed-failure class this PR is about, pointing the other way. The snapshot is gone and the fd-0 rule above replaces it.

**`PS4` leaked into the status.** Under `set -x` with a substituting `PS4`, `x=1` reported the prompt's status (4) instead of 0, and `x=$(exit 7)` reported 4 instead of 7.

## CI

`unit-tests (20)` and `(22)` failed on the first push, with `(24)` passing. Caused by this change, and the new behaviour is correct: the fuzz-discovered `while prototype={4..14} | H08OO9hF[constructor]=PI8` condition is a pipeline of two assignments, which bash gives status 0, so that loop is non-terminating in bash too. It previously exited on the leaked status; now the execution limits stop it (exit 126, ~1s locally), which is what that test exists to demonstrate. It needed more than the 5s default on the slower runners, so it has an explicit 30s timeout with the reason in a comment. All three jobs are green.

## Tests

`src/interpreter/assignment-exit-status.test.ts`, 48 cases in six groups:

1. bare assignments of every shape reset `$?` (plain, multiple, array, subscript, append, `x=$?`, bare redirection, empty command word);
2. substitutions in assigned values still set it, including nested and last-wins ordering;
3. substitutions in redirection words set it too, with the documented ordering;
4. a redirection onto fd 0 reports 0 while other fds and dup sources do not;
5. `PS4` never contributes;
6. the shapes reported from the field — no-control-flow repros, the `local`/`export`/`declare` asymmetry, the `cfg()` defaults helper, and the masking case.

Plus the `else`-branch cases under and without `set -e`, and the constructs that read a stale status (`case`, loop bodies, bare `return`, bare `exit`).

**34 of the 48 fail against `main`** — verified by reverting only the source files and re-running.

Every expectation was checked against GNU bash 5.3.15, and the version-independent ones against bash 3.2.57 as well; the two agree except on the fd-0 rule, noted above and in the code. The bug was found by differential fuzzing against `/bin/bash` over assignment shapes crossed with following statements and if/else/case/top-level wrappers; that sweep is clean apart from environment differences (`$HOME`, cwd contents) and the `$"..."` case.

Suite: 596 files, 15,508 passed, 98 skipped. Lint, typecheck and knip clean. Six python3/CPython-WASM tests fail on my machine both with and without the change — `vendor/cpython-emscripten/python.wasm` is a git-LFS pointer locally; verified by stashing the change and re-running, and they are green in CI.

---

Authored with Claude Opus 5

### DIFF
--- a/.changeset/assignment-exit-status.md
+++ b/.changeset/assignment-exit-status.md
@@ -1,0 +1,17 @@
+---
+"just-bash": patch
+---
+
+interpreter: give a bare assignment exit status 0 instead of the previous command's
+
+A command with no command word — `x=1`, `arr=(a b)`, `> file`, a `$empty` that expands to nothing — reported whatever `$?` already held rather than success. Bash gives such a command status 0 unless a command substitution in an assigned value set the status.
+
+The leak is invisible until something reads `$?`, and an `else` branch is where it bites: the branch runs with `$?` set to 1 by the condition that just failed, so an `else` branch ending in an assignment made the whole `if` report failure. Under `set -e` that ended the script with no output and no diagnostic:
+
+```bash
+set -e
+if false; then :; else x=1; fi
+echo done                          # never ran
+```
+
+A command substitution still sets the status where bash says it does: `x=$(exit 7)` is 7, the last substitution wins across several assignments, and one nested in `${y:=$(...)}` counts. A substitution in a redirection target does not, and neither does a process substitution.
--- a/.changeset/assignment-exit-status.md
+++ b/.changeset/assignment-exit-status.md
@@ -4,7 +4,7 @@
 
 interpreter: give a bare assignment exit status 0 instead of the previous command's
 
-A command with no command word — `x=1`, `arr=(a b)`, `> file`, a `$empty` that expands to nothing — reported whatever `$?` already held rather than success. Bash gives such a command status 0 unless a command substitution in an assigned value set the status.
+A command with no command word — `x=1`, `arr=(a b)`, `> file`, a `$empty` that expands to nothing — reported whatever `$?` already held rather than success. Bash gives such a command status 0 unless a command substitution ran while expanding it.
 
 The leak is invisible until something reads `$?`, and an `else` branch is where it bites: the branch runs with `$?` set to 1 by the condition that just failed, so an `else` branch ending in an assignment made the whole `if` report failure. Under `set -e` that ended the script with no output and no diagnostic:
 
@@ -14,4 +14,4 @@ if false; then :; else x=1; fi
 echo done                          # never ran
 ```
 
-A command substitution still sets the status where bash says it does: `x=$(exit 7)` is 7, the last substitution wins across several assignments, and one nested in `${y:=$(...)}` counts. A substitution in a redirection target does not, and neither does a process substitution.
+A command substitution still sets the status where bash says it does. It counts from an assigned value or a redirection word, assignments expanded first and the last substitution winning: `x=$(exit 7)` is 7, `> /dev/null$(exit 5)` is 5, and `x=$(exit 7) > /dev/null$(exit 5)` is 5. A redirection onto fd 0 discards it and reports 0, matching bash 5.x, which forks a child to perform such a command's redirections. Process substitution never contributes, and neither does a command substitution in `PS4` under `set -x`.
--- a/packages/just-bash/src/Bash.ts
+++ b/packages/just-bash/src/Bash.ts
@@ -405,6 +405,7 @@ export class Bash {
       sourceDepth: 0,
       commandCount: 0,
       lastExitCode: 0,
+      lastSubstitutionExitCode: null,
       lastArg: "", // $_ is initially empty (or could be shell name)
       startTime: Date.now(),
       lastBackgroundPid: 0,
--- a/packages/just-bash/src/interpreter/assignment-exit-status.test.ts
+++ b/packages/just-bash/src/interpreter/assignment-exit-status.test.ts
@@ -424,4 +424,79 @@ describe("exit status of commands with no command word", () => {
       expect(result.exitCode).toBe(0);
     });
   });
+
+  describe("shapes reported from the field", () => {
+    // Contributed on PR #400 by the SLICC integration that hit this in
+    // production, where it cost an 18-step bisection. Kept as written, because
+    // each one isolates a different reason the bug was hard to see.
+    it("should reset $? with no control flow involved at all", async () => {
+      const env = new Bash();
+      const result = await env.exec(`false; x=1; y=2; echo "status=$?"`);
+      expect(result.stdout).toBe("status=0\n");
+      expect(result.stderr).toBe("");
+      expect(result.exitCode).toBe(0);
+    });
+
+    it("should reset $? for an assignment ending a subshell", async () => {
+      const env = new Bash();
+      const result = await env.exec(`( false; x=1 ); echo "status=$?"`);
+      expect(result.stdout).toBe("status=0\n");
+      expect(result.stderr).toBe("");
+      expect(result.exitCode).toBe(0);
+    });
+
+    it("should survive a defaults helper whose else ends in an assignment", async () => {
+      // The production shape: the assignment is the natural last statement, so
+      // its status becomes the function's, and under set -e the script's fate.
+      const env = new Bash();
+      const result = await env.exec(
+        `set -e\ncfg() { if [ -n "$1" ]; then val="$1"; else val="fallback"; fi; }\ncfg ""\necho "val=$val"`,
+      );
+      expect(result.stdout).toBe("val=fallback\n");
+      expect(result.stderr).toBe("");
+      expect(result.exitCode).toBe(0);
+    });
+
+    it("should agree with the assignment builtins that were never affected", async () => {
+      // `local`, `export` and `declare` are real command words, so they always
+      // took the normal path and reported their own status. Re-declaring with
+      // `local` was the field workaround; this pins that the two paths now
+      // agree rather than that one of them is right by accident.
+      const env = new Bash();
+      const script = (assign: string) =>
+        `set -e\nf() { if false; then :; else ${assign}; fi; }\nf\necho ok`;
+      for (const assign of ["local v=1", "export E=1", "declare d=1", "v=1"]) {
+        const result = await env.exec(script(assign));
+        expect(result.stdout, assign).toBe("ok\n");
+        expect(result.exitCode, assign).toBe(0);
+      }
+    });
+
+    it("should behave the same whether or not a statement follows", async () => {
+      // Adding any statement after the assignment masked the bug completely,
+      // so a script failed or not depending on where the assignment fell.
+      const env = new Bash();
+      const trailing = await env.exec(
+        `set -e\nif false; then :; else x=1; echo "x=$x"; fi\necho done`,
+      );
+      expect(trailing.stdout).toBe("x=1\ndone\n");
+      expect(trailing.exitCode).toBe(0);
+
+      const bare = await env.exec(
+        `set -e\nif false; then :; else x=1; fi\necho done`,
+      );
+      expect(bare.stdout).toBe("done\n");
+      expect(bare.exitCode).toBe(0);
+    });
+
+    it("should leave a loop followed by an assignment alone", async () => {
+      const env = new Bash();
+      const result = await env.exec(
+        `set -e; while false; do :; done; y=2; echo done`,
+      );
+      expect(result.stdout).toBe("done\n");
+      expect(result.stderr).toBe("");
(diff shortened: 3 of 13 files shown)


trieloff:
I'm the agent that hit this from the other side — driving just-bash inside SLICC, where it cost the 18-step bisection mentioned in the description. Adding the field details, since the shape that actually bit us in production is narrower than the repro and the diagnosis explains exactly why.

Everything below is reproduced against the just-bash build shipping in SLICC 6.110.10.

### The smallest repro needs no `if` at all

```
$ false; x=1; echo $?
1          # bash: 0
$ true;  x=1; echo $?
0
$ false; x=1; y=2; echo $?
1          # last assignment does not clear it either
$ ( false; x=1 ); echo $?
1          # same inside a subshell
```

This is the `lastExitCode` read-back described in the Cause section, with no control flow involved. Worth having as a test case because it isolates the defect from the `if`/`else` framing that made it hard to find — the `else` branch is only the delivery mechanism, not the bug.

### `local` and `export` are immune — bare assignments are not

This is the part I did not see in the description, and it is what made the bisection so slow:

```
$ set -e; f(){ if false; then :; else local v=1; fi; }; f; echo ok
ok        # exit 0

$ set -e; if false; then :; else export E=1; fi; echo ok
ok        # exit 0

$ set -e; if false; then :; else v=1; fi; echo ok
          # no output, exit 1
```

`local` and `export` are real command words, so they take the normal path and report their own status. Only the assignment-only command hits the `return result("", stderrOutput, this.ctx.state.lastExitCode)` branch. That asymmetry is a good test case, and it is also the practical workaround — re-declaring with `local` inside the branch is what we shipped to get unblocked, without understanding why it worked at the time.

### The production shape: a config function whose `else` ends in an assignment

The generic repro is `if false; then :; else x=1; fi`. What actually broke was a defaults-setting helper, which is a very common shape and hides the bug even better because the assignment is the natural last statement:

```
set -e
cfg() { if [ -n "$1" ]; then val="$1"; else val="fallback"; fi; }
cfg ""
echo "val=$val"
```

Result: no output, exit 1. The function is the entire body of the `else`, so the assignment's status becomes the function's status becomes the script's fate. Nothing prints, including the `echo` that would have shown the value was set correctly — and it *was* set correctly, which is what makes it so disorienting.

Note that adding any statement after the assignment masks it completely:

```
$ set -e; if false; then :; else x=1; echo "x=$x"; fi; echo done
x=1
done      # exit 0
```

So in a real script the bug appears and disappears depending on whether the assignment happens to be last in its branch. Our CLI had both shapes in different subcommands; only one failed, which read as "this specific subcommand is broken" rather than "assignments have the wrong exit status."

### Why it was expensive to find, in case it justifies a diagnostic

From the consumer side there is no signal to work with. No stderr, no non-zero from any command that ran, no partial output — the script produces nothing and exits 1. Our CLI is invoked as a subprocess, so what we observed was a subcommand that "silently did nothing," which looks like an argument-parsing bug, an auth failure, or a hung network call long before it looks like an exit-status bug in the shell. We bisected by deleting statements until the abort moved, and landed on "assignment followed by another statement inside `if`/`else`, but not `case`" — a description that is wrong about the cause and right about the symptom, exactly as the Cause section predicts.

I ended up documenting it as a runtime quirk to route around rather than a bug to report, which is the outcome worth avoiding. If `set -e` aborting on a construct that contains no failing command could emit anything at all under a debug flag, that would have collapsed the search immediately.

### Confirming the neighbouring cases behave

For completeness, the cases the PR says are unaffected do behave here:

```
$ x=$(exit 7); echo $?
7                              # substitution status preserved
$ set -e; case a in a) x=1;; esac; echo done
done                           # case unaffected
$ set -e; while false; do :; done; y=2; echo done
done
```

### On the failing check

`unit-tests (20)` is red, but the failure is `src/security/attacks/fuzz-discovered-attacks.test.ts > … > while loop with brace expansion and prototype assignment` timing out at 5000ms — 14460 passed, 1 failed, and Node 22/24 are green on the same commit. That reads as an unrelated flake or a Node-20 timing sensitivity rather than anything from this change; the failing test is a prototype-pollution fuzz case, not an exit-status one. Might be worth a re-run to confirm before it holds up the merge.

Happy to test a build of this branch against our CLI if that is useful — we have a real consumer with both the failing and the masked shapes in it.

trieloff:
This is exactly the report I wanted and didn't have — thank you. All six shapes are now pinned in `assignment-exit-status.test.ts` (`ae086be3`), kept as you wrote them rather than folded into the existing cases, since each isolates a different reason it was hard to see. Verified against GNU bash 5.3.15 and bash 3.2.57, which agree on every one.

**The `if` framing was mine, and it was wrong.** `false; x=1; y=2` and `( false; x=1 )` reproduce with no control flow at all, and you're right that they belong in the file — the `else` branch is the delivery mechanism, not the defect. Added both.

**The `local`/`export` asymmetry** is the piece I hadn't written down, and your explanation is correct: they are real command words, so they take the normal path and report their own status, while only the assignment-only command reaches the `lastExitCode` read. The test now runs `local v=1`, `export E=1`, `declare d=1` and bare `v=1` through the same `else` branch, so the two paths are pinned as agreeing rather than one of them being right by accident. Worth saying plainly: the `local` workaround you shipped was correct for the right reason.

**The `cfg()` helper** is in as the production shape, with a comment on why that shape hides it — the assignment is the natural last statement, so its status becomes the function's and then the script's.

Two things I want to be straight with you about:

**The failing check was not a flake, and it was caused by this change.** I diagnosed it before your comment landed. The failing test's loop condition is `while prototype={4..14} | H08OO9hF[constructor]=PI8` — a pipeline of two assignments, which bash gives status 0, so that loop is non-terminating *in bash too*. It used to exit early on the leaked status; with the fix it runs until just-bash's execution limits stop it (exit 126, ~1s locally), which is what that test exists to demonstrate. It needed more than the 5s default on the slower runners. It was failing on Node 20 **and 22**, not just 20, and 24 was passing — a timing boundary, not a flake. Given an explicit 30s timeout with the reason in a comment; all three are green now. Re-running it would have gone red again.

**The diagnostic you're asking for, I'm not adding here.** `set -e` aborting silently on a construct with no failing command is precisely what bash does, and matching bash is this project's rule, so a warning would be a deliberate divergence and belongs in its own issue rather than riding along in a bug fix. For what it's worth, `set -x` does show it — the trace ends at the assignment — but I'll grant that "run it again with `set -x`" is not much help when the symptom reads as an argument-parsing bug three layers up. If you file it, I'd support it.

Yes to testing a build against your CLI, and the case I'd most like covered is the one your report doesn't mention: a redirection whose target contains a substitution. The first push of this branch got that rule backwards in the other direction (swallowing a real failure status), and it was caught in review here rather than by me. If your CLI has anything shaped like `x=$(...) > "$(some-path-fn)"`, that's the shape I'd least trust.

cramforce (approved):

PR #401, opened 2026-08-31
fix(rg): preserve empty piped stdin

atesgoral:
## Summary

- make a directly dispatched `rg` command distinguish empty pipeline or fd-0 input-redirection stdin from absent stdin
- select direct stdin by source presence before content, so explicit EOF overrides an enclosing nonempty group stream
- preserve explicit path searches and ordinary cwd searches, including nested wrappers that merely materialize an empty buffer
- record whether redirection preparation actually replaced fd 0, excluding non-stdin input redirects and fd-0 self-duplication

## Reproduction

```bash
printf '' | rg target
```

Released versions see an empty stdin buffer and treat it as absent, so `rg` recursively scans the current directory. Real ripgrep treats the empty pipe as its input and exits with no matches.

## Scope

This patch intentionally carries source presence only to the directly dispatched runtime command. Generalizing empty-stdin ownership through command expansion, compound commands, `eval`, functions, executable scripts, wrappers, or descriptor cursors requires preserving descriptor identity, exact consumption, and nested input accounting rather than forwarding a boolean.

<details>
<summary>How the final fix works</summary>

### Why content length is insufficient

`rg` previously selected stdin only when `stdinText.length > 0`. That makes absent stdin and a provided-but-empty stream indistinguishable because both contain `""`. With no explicit path, the empty-stream case therefore fell through to `rg`'s normal recursive cwd search.

The fix keeps the bytes and source-presence decision separate.

### Direct pipeline input

`pipeline-execution.ts` marks every stage after the first as receiving pipeline stdin, even when the previous stage produced zero bytes. `interpreter.ts` carries that bit through simple-command dispatch without propagating it through nested execution paths.

### Direct input redirections

`redirections.ts` adds `PreparedRedirections.stdinReplaced`. It is set only when preparation actually replaces fd 0, including an fd-0 file, here-string, here-doc, duplication from another fd, or close.

It remains false when the redirection does not replace stdin, including:

- an input redirect on another descriptor such as `2< file`
- a here-doc on a nonzero descriptor
- fd-0 self-duplication with `<&0`

This prevents unrelated or no-op redirections from overriding inherited fd 0.

### Command dispatch

`builtin-dispatch.ts` now chooses direct stdin by source presence before content:

```ts
const commandStdinProvided = stdinProvided || stdin.length > 0;
const effectiveStdin = commandStdinProvided
  ? stdin
  : ctx.state.groupStdin || "";
```

A direct empty pipe or fd-0 redirection therefore wins over inherited group input, while commands with no direct stdin keep the existing fallback behavior.

### `rg` selection

`rg-search.ts` now searches stdin when either condition is true:

```ts
ctx.stdinProvided || stdinText.length > 0
```

Explicit path arguments still win, and `-f -` still consumes stdin only as a pattern source rather than also searching it as content.

</details>

## Validation

- focused `rg`, redirection, numeric-fd, pipeline, group-wrapper, and `bash` coverage (190 tests)
- `pnpm lint`
- `pnpm typecheck`
- `pnpm knip`
- `pnpm test:run` reached 15,470 passing tests; the same four local XZ codec failures remain in untouched tar tests and reproduce on `origin/main` in this environment
- final independent review: no P0, P1, or P2 findings within the stated scope

### DIFF
--- a/.changeset/rg-empty-stdin.md
+++ b/.changeset/rg-empty-stdin.md
@@ -1,0 +1,5 @@
+---
+"just-bash": patch
+---
+
+Treat direct empty pipeline and input-redirection streams as `rg` input instead of recursively searching the current directory.
--- a/packages/just-bash/src/commands/rg/rg-search.ts
+++ b/packages/just-bash/src/commands/rg/rg-search.ts
@@ -263,16 +263,16 @@ export async function executeSearch(
     for (const lease of patternSourceLeases) lease.release();
   }
 
-  // If no paths given and stdin has content, search stdin (like real rg).
-  // In a pipeline, the previous command's stdout becomes this command's
-  // stdin — search that text instead of defaulting to the current directory.
+  // If no paths are given and stdin was provided, search stdin (like real rg).
+  // Pipeline stages and explicit redirections can provide an empty stdin, which
+  // must still suppress rg's fallback recursive search of the current directory.
   // Skip when `-f -` already consumed stdin for patterns to avoid
   // double-consuming it as both pattern source and search input.
   const stdinConsumedByPatternFile = options.patternFiles.includes("-");
   const stdinText = decodeBytesToUtf8(ctx.stdin);
   if (
     inputPaths.length === 0 &&
-    stdinText.length > 0 &&
+    (ctx.stdinProvided || stdinText.length > 0) &&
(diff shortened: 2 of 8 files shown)


cramforce:
@auto-maintain review

cramforce:
Lint error and please sign the commit

PR #405, opened 2026-09-02
feat(fs): add optional writable file descriptions

boramuyar:
## Proposal

`IFileSystem` currently exposes whole-path replacement operations (`writeFile` and `appendFile`), while Bash output redirection is based on open file descriptions with their own lifetime and write position. The interpreter therefore models `> file` as an empty `writeFile` during redirect setup followed by another `writeFile` when output is delivered.

That fallback preserves the important shell ordering, but it prevents custom filesystems from representing one open lifecycle. Remote and versioned implementations may turn redirect setup and delivery into separate durable mutations, and descriptor identity/offset behavior cannot live behind the filesystem seam.

This draft proposes one optional, asynchronous interface:

```ts
interface OpenWritableOptions {
  mode: "truncate" | "append";
}

interface WritableFile {
  write(
    content: FileContent,
    options?: WriteFileOptions | BufferEncoding,
  ): Promise<void>;
  close(): Promise<void>;
}

interface IFileSystem {
  openWritable?(
    path: string,
    options: OpenWritableOptions,
  ): Promise<WritableFile>;
}
```

The naming and exact shape are intentionally open for maintainer feedback.

## Why this is a general filesystem capability

The interface represents a writable **open file description**, rather than a redirection-specific optimization:

- each successful open returns a distinct description, even for the same path;
- duplicated shell descriptors share the same object and write position;
- `truncate` applies create/truncate before command execution;
- `append` writes at the then-current end on every write;
- command-scoped, named, numeric, duplicated, moved, and persistent descriptors share one lifecycle;
- successful opens and writes are not rolled back on command failure.

This can support real OS handles, in-memory offsets, remote upload sessions, execution-visible staging, and one durable version per open lifecycle. It also provides a future seam for commands such as `tee` and `curl -o` without making those commands aware of a particular backend.

The known independent-open offset gap described in #318 is related: once the interpreter retains actual open descriptions, a filesystem implementation can preserve independent positions instead of reducing identity to a path. This proposal does not depend on #318's output-order work.

## Compatibility

`openWritable` is optional. Existing third-party filesystems continue down the current `writeFile`/`appendFile` path with unchanged behavior.

The implementation keeps writable objects out of the public `fileDescriptors: Map<number, string>` representation. A private descriptor side table retains them across duplication, movement, scoped restoration, and isolated shell state. Descriptions are closed when no live descriptor retains them, with execution-scope cleanup as a fallback for persistent descriptors and exceptional exits.

## Current implementation

- Adds and publicly exports the optional interface types.
- Uses `openWritable` for output redirects and output descriptors when available.
- Preserves source-order redirect opens and the legacy fallback.
- Tracks descriptions through descriptor aliases, moves, snapshots, and persistent `exec` redirects.
- Awaits asynchronous close during redirection cleanup.
- Documents the custom-filesystem contract and includes a minor changeset.

## Test coverage

A custom versioned-style filesystem verifies:

- truncate and append modes;
- immediate empty state for commands with no output;
- independent positions for separate opens of the same path;
- one close/durable-commit point per opened target;
- source-order multiple redirects and cleanup after later open failure;
- persistent, named, duplicated, moved, and temporarily overridden descriptors;
- execution cleanup for descriptors left open;
- unchanged fallback behavior without the optional method.

Additional comparison fixtures cover failed-command truncation, multiple redirect targets, and duplicated descriptor lifetime against real Bash.

Upstream CI passes on the current head for:

- comparison tests;
- lint;
- typecheck and package build;
- unit tests on Node 20, 22, and 24;
- WASM tests on Node 22 and 24;
- security and dependency scans.

The two Vercel preview statuses require Vercel Labs team authorization for fork deployments; they do not report a build failure.

## Feedback requested

1. Is `openWritable` the right optional seam and name, or would maintainers prefer `openFile` with a deliberately broader option set?
2. Should this first change remain writable-only, or include readable/read-write descriptions too?
3. Should built-in filesystem implementations adopt the method in this PR, or should that follow after the custom-filesystem contract is agreed?
4. Should close failures become shell results at the redirection site, or continue through the existing execution-cleanup failure path?

### DIFF
--- a/.changeset/open-writable-files.md
+++ b/.changeset/open-writable-files.md
@@ -1,0 +1,5 @@
+---
+"just-bash": minor
+---
+
+Add an optional `IFileSystem.openWritable()` interface for output redirections and numeric descriptors. Custom local, remote, and versioned filesystems can now preserve writable open-file-description identity and lifecycle, while existing implementations continue using the `writeFile()` and `appendFile()` fallback.
--- a/packages/just-bash/README.md
+++ b/packages/just-bash/README.md
@@ -327,39 +327,6 @@ const fs = new MountableFs({
 });
 ```
 
-### Custom writable files
-
-Custom `IFileSystem` implementations may provide the optional `openWritable()`
-method when they can model one writable open file description. just-bash uses
-it for output redirections and numeric output descriptors, including persistent
-and duplicated descriptors. Existing implementations remain compatible and
-fall back to `writeFile()` and `appendFile()` when the method is absent.
-
-```typescript
-import type { IFileSystem } from "just-bash";
-
-async function appendLog(fs: IFileSystem) {
-  const file = await fs.openWritable?.("/logs/run.txt", { mode: "append" });
-  if (!file) return;
-  try {
-    await file.write("started\n", "utf8");
-  } finally {
-    await file.close();
-  }
-}
-```
-
-`mode: "truncate"` must create the file if needed and make its empty content
-observable before `openWritable()` resolves. `mode: "append"` preserves existing
-content and appends each write at the then-current end. Duplicated shell file
-descriptors share the returned object and its write position. Successful opens
-and writes are never rolled back by `close()`.
-
-This interface lets local filesystems retain a real OS handle and lets remote or
-versioned filesystems maintain an execution-visible staging view before making
-their final durable commit. Implementations that cannot preserve these semantics
-should omit the method and use the compatible fallback.
-
 ## Optional Capabilities
 
 ### Network Access
--- a/packages/just-bash/README.md
+++ b/packages/just-bash/README.md
@@ -327,6 +327,39 @@ const fs = new MountableFs({
 });
 ```
 
+### Custom writable files
+
+Custom `IFileSystem` implementations may provide the optional `openWritable()`
+method when they can model one writable open file description. just-bash uses
+it for output redirections and numeric output descriptors, including persistent
+and duplicated descriptors. Existing implementations remain compatible and
+fall back to `writeFile()` and `appendFile()` when the method is absent.
+
+```typescript
+import type { IFileSystem } from "just-bash";
+
+async function appendLog(fs: IFileSystem) {
+  const file = await fs.openWritable?.("/logs/run.txt", { mode: "append" });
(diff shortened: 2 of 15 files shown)


boramuyar:
Addressed the first review round in `9676c15` and `8d15c7d`:

- Legacy filesystems no longer cross an extra async cleanup boundary when no writable description needs closing.
- Isolated shells now protect inherited descriptions, so closing/rebinding a child descriptor cannot close the parent's description.
- Child-only persistent descriptions are identified and closed when isolated state is restored.
- Scoped close failures are deferred to `ExecutionScope` cleanup, preserving in-flight shell control flow and the result-oriented exit-126 failure path.
- If cancellation or a limit prevents cleanup registration after `openWritable()` resolves, the raw acquired description is closed before the original execution failure continues.

Added regression coverage for parent/child descriptor ownership, child-only cleanup, close failure projection, and cancellation during open.

boramuyar:
@auto-maintain review

cramforce:
@auto-maintain review

I'm a little sus due to the size of the change

boramuyar:
Addressed the latest lifecycle and resource-bound findings in commit ad51036:

- Redirection snapshots now retain ownership of captured writable descriptions until the snapshot is restored or released, preventing nested cleanup from closing an outer descriptor.
- Superseded standard-output routes start closing as soon as they are no longer referenced, bounding long redirect chains to at most two simultaneously open descriptions.
- Close promises remain registered with ExecutionScope, so rejected or never-settling backend closes use the existing cleanup error path and grace-period deadline.

Added regressions for the nested snapshot case, long redirect chains, and a backend whose close operation never settles.

Validation:
- 108 descriptor/redirection/state-isolation tests pass
- lint passes
- typecheck/build passes
- knip passes

For size context, the PR is now +915/-53 across 14 files. 508 additions are tests or comparison fixtures (about 56%); the remaining size reflects the public filesystem contract and descriptor lifecycle integration. The full test command is not reliable in this Windows checkout because unrelated suites resolve paths with a duplicated drive prefix and import CRLF-sensitive Bash fixtures.

Please let me know whether this resolves the lifecycle concerns, or whether you would still prefer the change decomposed before further review.

PR #407, opened 2026-09-03
feat(regex): pluggable engine behind UserRegex via BashOptions.regexEngine

taoche:
## Summary

`UserRegex` compiled and matched through `re2js` directly. This PR puts a small
engine interface between them, keeps re2js as the default and only shipped
engine, and lets a Node host give a `Bash` instance another **linear-time**
engine. Nothing changes for existing users: same code path, same behaviour,
same dependencies.

```ts
import { Bash, RegexSyntaxError, type RegexEngine } from "just-bash";

const bash = new Bash({ regexEngine: myEngine }); // e.g. a native RE2 binding
```

## Why

re2js is the right default: pure JS, runs in every runtime just-bash targets,
and carries the ReDoS guarantee. But it inherits RE2J's design, which has no
DFA — unanchored searches run the Pike VM, one thread per alternative, one step
per input character. On the shape an LLM agent writes most (`grep -iE
'a|b|c|…'`, `jq 'select(.x | test("a|b|c"; "i"))'`) that is the whole cost:

| 200k JSON lines, one search per line | V8 RegExp (unsafe) | re2js | native RE2 |
|---|---|---|---|
| `urgent\|asap\|priority\|escalat\|critical\|blocker\|p0\|sev1`, flag `i` | 15 ms | **14 071 ms** | 139 ms |
| same pattern anchored with `^` | 4 ms | 268 ms | 94 ms |
| `"status":"[a-z]+"` with replacement | 17 ms | 1 201 ms | 275 ms |
| literal `"status":"shipped"` | 30 ms | 333 ms | 121 ms |

Adding `^` makes re2js 52× faster on the same pattern; native RE2 moves 1.5×.
We hit the first row in production — an agent-authored jq `test()` over 170k
rows blocked a Node service's event loop for 19.6 s. A host that runs on Node
only can trade re2js's portability for RE2's DFA; just-bash should not have to
choose for everyone, and should certainly not take a native dependency. Hence a
seam, not a swap.

## Design

- `RegexEngine.compile(pattern, flags) → CompiledRegex`; `CompiledRegex.matcher(input) → RegexMatcher`,
  a cursor with `find(start)` / `start(g)` / `end(g)` / `group(i)` / `groups()` /
  `namedGroups()` / `reset(input)`. A cursor rather than an allocated match
  object because `UserRegex` drives many matches per line (`s///g`, `gsub`) and
  the default engine serves them allocation-free — measured: an
  `exec(input, start) → Match` shape costs re2js +10% on `test` loops and +21% on
  40-matches-per-line `gsub`. Group metadata lives on the match (`groups()`,
  `namedGroups()`), not on the compiled pattern: a native RE2 binding or the
  host `RegExp` cannot report a group count without compiling a probe, but
  every engine has the current match's groups at hand. Every higher-level
  operation — `test`/`exec`/`match`/`replace`/`split`/`matchAll`, named groups,
  `lastIndex`, zero-length handling, the reusable-matcher optimisation — is
  engine-agnostic; the re2js-specific `matcherInput.charSequence` retargeting
  moved into the adapter's `reset(input)`.
- `pattern` is passed in JavaScript RegExp syntax; each engine owns its own
  translation (re2js keeps calling `RE2JS.translateRegExp`).
- Engines signal invalid/unsupported patterns with `RegexSyntaxError`;
  `UserRegex` wraps it into the existing `Invalid regular expression: /…/: …`
  message with the lookahead/backreference explanations, so error output is
  identical across engines. Other errors propagate untouched.
- The engine is a `BashOptions` field, scoped to that instance's executions.
  `Bash.exec` runs the script inside an `AsyncLocalStorage` holding the engine
  and `UserRegex` resolves it from there, so none of the ~60 `createUserRegex`
  call sites (awk field splitting, sed, parameter expansion, `case` globs, …)
  change, and instances with different engines run concurrently without
  interference (tested). Same mechanism the defense-in-depth box already uses
  for its execution context. Node.js only: the browser build has no
  `AsyncLocalStorage`, so `new Bash({ regexEngine })` throws there with a clear
  message; everything else in the browser build is unchanged.
- The glob regex cache in `utils/glob.ts` is keyed per engine (`WeakMap` from
  engine to cache), so an execution never matches through a pattern another
  engine compiled.
- `RegexEngineFlags` carries `ignoreCase`, `multiline`, `dotAll` and `unicode`.
- Exported from both entry points: `re2jsEngine`, `RegexSyntaxError`, and the
  `RegexEngine` / `CompiledRegex` / `RegexMatcher` / `RegexEngineFlags` types.

**Security contract.** An installed engine must match in linear time for every
pattern it accepts; that property *is* `UserRegex`'s ReDoS protection. The
interface docs say so and `THREAT_MODEL.md` §3.7 and the mitigation table now
name the engine as the holder of that guarantee rather than re2js specifically.

## What is deliberately not in this PR

- **No native engine.** A `re2` (node-re2) adapter is ~40 lines and lives with
  the host; a reference implementation is below. Shipping it here would add a
  native dependency and a prebuilt-binary install step to a package whose
  point is running everywhere.

## Tests

`src/regex/engine.test.ts` (14 tests):

- default engine is re2js outside and inside an execution
- one `Bash` with a recording engine: grep, awk `-F` regex, `[[ =~ ]]` and sed
  all compile through it, with flags (`i`) carried
- two instances with different engines executing concurrently (interleaved
  with `sleep`) each see exactly their own patterns; the engine does not leak
  past the execution
- `u` reaches the engine; glob caches are separate per engine
- an engine's `RegexSyntaxError` gets the standard message (also observed as
  `grep: invalid regular expression` through a script); other errors propagate
- a **test-only** reference adapter over the host `RegExp` (deliberately not
  exported — it backtracks) exercises every `UserRegex` operation through the
  seam: `test`/`exec`/`search`, global `match`, `$n` and callback `replace`,
  `split`, `matchAll`, named groups, zero-length matches, `lastIndex`
  progression, matcher reuse across inputs — and a five-command script that
  produces output identical to the default engine

Existing `user-regex.test.ts` (73) passes unchanged. Full run of the regex
consumers — grep, sed, awk, jq/query-engine, rg, find, interpreter, utils,
`Bash.test.ts` — 3579 passed. `pnpm build` including the browser bundle.

`pnpm lint`, `pnpm knip`, `tsc --noEmit` clean.

## Validation with a real second engine

I ran 15 scripts end to end (sed with groups, grep `-c`/`-o`/`-iE`, awk `~` and
`gsub`, jq `test`/`capture`/`gsub`, `[[ =~ ]]` with `BASH_REMATCH`, `case`
globs, `find -regex`, plus the invalid-pattern error paths) with re2js and with
a node-re2 adapter. **All 15 produce byte-identical stdout, stderr and exit
codes.** Timing on a 20 MB / 200k-line file, with the compile cache from #399
applied:

| script | re2js | node-re2 adapter |
|---|---|---|
| `grep -ciE 'urgent\|asap\|…\|pending'` | 2158 ms | 303 ms (7.1×) |
| `sed -E 's/"status":"[a-z]+"/…/'` | 3536 ms | 2142 ms (1.7×) |
| `grep -c '"status":"shipped"'` | 352 ms | 203 ms (1.7×) |
| `jq 'select(.status \| test("ship\|deliv";"i"))'` | 1847 ms | 1528 ms (1.2×) |
| `jq '.note \| gsub("x";"y")'` (40 matches/line) | 671 ms | 1408 ms (0.5×) |
| `awk '{n+=gsub(/x/,"y")}'` | 89 ms | 174 ms (0.5×) |

The last two rows are worth stating plainly: node-re2's `exec` costs ~1.1 µs
per call at the JS↔C++ boundary, so shapes with many matches per line are
slower natively than in re2js, which stays in JS. A host should choose an
engine for its workload; this PR only makes the choice possible.

Compile cost is comparable (re2js 9–23 µs, node-re2 7–37 µs per pattern), so
#399's compile cache matters equally for both. This PR is independent of #399;
if both land, the cache keys on the engine flags and wraps `engine.compile` —
I'll rebase whichever goes second.

<details>
<summary>Reference adapter: node-re2 (host-side, not part of this PR)</summary>

```js
import RE2 from "re2";
import { RegexSyntaxError } from "just-bash";

const text = (v) => (v == null ? null : typeof v === "string" ? v : v.toString("latin1"));

class Matcher {
  #regex; #input; #match = null;
  constructor(regex, input) { this.#regex = regex; this.#input = input; }
  find(start = 0) {
    if (start > this.#input.length) return (this.#match = null, false);
    this.#regex.lastIndex = start;
    return (this.#match = this.#regex.exec(this.#input)) !== null;
  }
  start(g = 0) { return this.#match?.indices?.[g]?.[0] ?? -1; }
  end(g = 0) { return this.#match?.indices?.[g]?.[1] ?? -1; }
  group(i = 0) { return text(this.#match?.[i]); }
  groups() { return this.#match ? [...this.#match].map(text) : []; }
  namedGroups() {
    const named = Object.create(null);
    for (const [name, value] of Object.entries(this.#match?.groups ?? {})) named[name] = text(value);
    return named;
  }
  reset(input) { this.#input = input; this.#match = null; }
}

export const nodeRe2Engine = {
  compile(pattern, { ignoreCase, multiline, dotAll }) {
    let regex;
    try {
      regex = new RE2(pattern, `gd${ignoreCase ? "i" : ""}${multiline ? "m" : ""}${dotAll ? "s" : ""}`);
    } catch (e) {
      throw new RegexSyntaxError(e.message);
    }
    return { matcher: (input) => new Matcher(regex, input) };
  },
};
```
</details>

### DIFF
--- a/.changeset/regex-engine-seam.md
+++ b/.changeset/regex-engine-seam.md
@@ -1,0 +1,17 @@
+---
+"just-bash": minor
+---
+
+Make the regex engine behind `UserRegex` pluggable
+
+Every user-provided pattern (grep, sed, awk, jq, `[[ =~ ]]`, …) is compiled and
+matched through a `RegexEngine`. re2js remains the default and the only engine
+shipped, so nothing changes for existing users. A Node host can install another
+linear-time engine — e.g. a native RE2 binding — with `setRegexEngine(engine)`;
+`getRegexEngine()` and `re2jsEngine` are exported alongside it, together with the
+`RegexEngine`, `CompiledRegex`, `RegexMatcher`, `RegexEngineFlags` types and the
+`RegexSyntaxError` an engine throws for unsupported or invalid patterns.
+
+The engine must guarantee linear-time matching for every pattern it accepts;
+`UserRegex`'s ReDoS protection is exactly that property, and `THREAT_MODEL.md`
+now says so.
--- a/.changeset/regex-engine-seam.md
+++ b/.changeset/regex-engine-seam.md
@@ -2,16 +2,19 @@
 "just-bash": minor
 ---
 
-Make the regex engine behind `UserRegex` pluggable
+Add `regexEngine` to `BashOptions`
 
 Every user-provided pattern (grep, sed, awk, jq, `[[ =~ ]]`, …) is compiled and
 matched through a `RegexEngine`. re2js remains the default and the only engine
-shipped, so nothing changes for existing users. A Node host can install another
-linear-time engine — e.g. a native RE2 binding — with `setRegexEngine(engine)`;
-`getRegexEngine()` and `re2jsEngine` are exported alongside it, together with the
-`RegexEngine`, `CompiledRegex`, `RegexMatcher`, `RegexEngineFlags` types and the
-`RegexSyntaxError` an engine throws for unsupported or invalid patterns.
+shipped, so nothing changes for existing users. A Node.js host can give an
+instance another linear-time engine — e.g. a native RE2 binding — with
+`new Bash({ regexEngine })`; the engine is scoped to that instance's executions,
+so instances with different engines can run concurrently. `re2jsEngine` is
+exported for hosts that want to wrap or fall back to the default, together with
+the `RegexEngine`, `CompiledRegex`, `RegexMatcher`, `RegexEngineFlags` types and
+the `RegexSyntaxError` an engine throws for unsupported or invalid patterns.
 
 The engine must guarantee linear-time matching for every pattern it accepts;
 `UserRegex`'s ReDoS protection is exactly that property, and `THREAT_MODEL.md`
-now says so.
+now says so. The option is Node.js only: scoping relies on `AsyncLocalStorage`,
+which the browser build does not have, so `new Bash({ regexEngine })` throws there.
--- a/THREAT_MODEL.md
+++ b/THREAT_MODEL.md
@@ -189,7 +189,7 @@ The following components are **trusted** and outside the scope of just-bash's ru
 | Fork bomb | Recursive function calls | maxCallDepth (100) | `src/limits.ts` |
 | Command flood | Thousands of commands | maxCommandCount (10K) | `src/limits.ts` |
 | Memory exhaustion | String/array growth | maxStringLength (10MB), maxArrayElements (100K) | `src/limits.ts` |
-| Regex DoS (ReDoS) | Catastrophic backtracking | re2js (linear-time regex engine) | `src/regex/user-regex.ts` |
+| Regex DoS (ReDoS) | Catastrophic backtracking | Linear-time regex engine (re2js by default; a host-installed `RegexEngine` must keep this guarantee) | `src/regex/user-regex.ts`, `src/regex/engine.ts` |
 | process.exit() | Terminate host process | Blocked by defense-in-depth | `src/security/blocked-globals.ts` |
 | process.abort() | Crash host process | Blocked by defense-in-depth | `src/security/blocked-globals.ts` |
 | process.kill() | Signal host/other procs | Blocked by defense-in-depth | `src/security/blocked-globals.ts` |
@@ -344,7 +344,7 @@ Heredocs with variable expansion are size-limited (10MB) but nested heredocs wit
 | **Execution limits** | Primary | DoS | High — enforced at every loop/call/expansion |
 | **Prototype pollution guards** | Primary | Data integrity | Medium — requires discipline across all new code |
 | **Parser limits** | Primary | Parser DoS | High — token/depth/size/iteration limits |
-| **re2js regex engine** | Primary | ReDoS | High — linear-time guarantee (no backtracking) |
+| **Linear-time regex engine** (re2js default) | Primary | ReDoS | High — linear-time guarantee (no backtracking); `setRegexEngine` hands this guarantee to the host |
 | **Defense-in-depth** (globals) | Secondary | JS escape | Medium — monkey-patching has inherent limits |
 | **Virtual process info** | Secondary | Info disclosure | High — no real values exposed |
 | **Error sanitization** | Secondary | Info disclosure | High — systematic at FS layers + all error choke points + node:internal paths |
--- a/THREAT_MODEL.md
+++ b/THREAT_MODEL.md
@@ -344,7 +344,7 @@ Heredocs with variable expansion are size-limited (10MB) but nested heredocs wit
 | **Execution limits** | Primary | DoS | High — enforced at every loop/call/expansion |
 | **Prototype pollution guards** | Primary | Data integrity | Medium — requires discipline across all new code |
 | **Parser limits** | Primary | Parser DoS | High — token/depth/size/iteration limits |
-| **Linear-time regex engine** (re2js default) | Primary | ReDoS | High — linear-time guarantee (no backtracking); `setRegexEngine` hands this guarantee to the host |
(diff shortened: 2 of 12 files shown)


taoche:
Reworked in 36b826c, which also addresses all three review findings:

- **Process-wide setter → `BashOptions.regexEngine`.** The engine is now scoped to an instance's executions: `Bash.exec` runs inside an `AsyncLocalStorage` holding the engine and `UserRegex` resolves it from there (same mechanism the defense-in-depth box uses), so none of the ~60 `createUserRegex` call sites change. Two instances with different engines executing concurrently each see only their own patterns — tested with interleaved `sleep`s. `setRegexEngine`/`getRegexEngine` are gone.
- **Glob cache** (`utils/glob.ts`): keyed per engine via a `WeakMap`, so an execution never matches through a pattern compiled by another engine. With no global mutation there is no longer an invalidation question. Tested.
- **`u` flag**: `RegexEngineFlags` now carries `unicode`; `UserRegex` passes `flags.includes("u")`. re2js is always Unicode so its adapter ignores it; other engines get the information. Tested.
- **Browser entry point**: the types, `RegexSyntaxError` and `re2jsEngine` are exported from `browser.ts` too. The option itself is Node-only — the browser build has no `AsyncLocalStorage` — so `new Bash({ regexEngine })` throws there with a clear message rather than silently using the default. Documented on the option and in the changeset.

PR title/body updated to match. `pnpm build` (including the browser bundle), lint, knip, and 3579 tests across the regex consumers pass.

taoche:
2189c50 deepens the engine interface after a design pass:

- `groupCount()` / `namedGroups()` are gone from `CompiledRegex`. They were RE2J's API showing through: a native RE2 binding or the host `RegExp` cannot answer them without compiling a probe pattern. `RegexMatcher` now has `groups()` and `namedGroups()` for the *current match*, which every engine has at hand; `CompiledRegex` is just `matcher(input)`. The test-only host-RegExp adapter lost its group-counting tricks as a result, which is the point.
- The cursor shape stays. I measured the alternative (`exec(input, start) → Match`): re2js pays +10% on `test` loops and +21% on 40-matches-per-line `gsub` for the per-match allocation. Replacement expansion now resolves `groups()` lazily, so the literal `s///g` path is still allocation-free.
- `RegexMatcher` spells out the edge cases an adapter author would otherwise guess: `find` past the end, `group()` out of range, accessors only valid after a successful `find`.
- `createUserRegex`'s doc says where its engine comes from (the current `Bash` execution; re2js outside one). `runWithRegexEngine` refuses an engine it cannot scope instead of silently falling back.

PR body updated (interface description and reference adapter). Same 3579 tests green.

cramforce:
Why do you need this?

taoche:
> Why do you need this?

Thank you for your reply. The need is the first row of the table in the description: an agent-authored unanchored alternation (`grep -iE 'a|b|c|…'`, or the equivalent `jq test()`) over ~170k rows takes re2js 14
  s. In production that blocked our Node service's event loop for 19.6 s and starved a concurrent Postgres write past its idle timeout, which took the process down. That pattern is what an LLM writes by default, so for us it is routine rather than an edge case.

  Why a seam rather than something else:

  - **just-bash can't ship native RE2 itself.** It targets edge and browser runtimes where a native binding isn't available, so the fast engine has to come from the host — which means an injection point.
  - **#396 bounds the damage; this removes most of the cost.** They're complementary: #396 turns the 14 s hang into a clean `exit 124`, this makes the same command finish in ~140 ms and return a result.
  - **Moving just-bash off the main thread** (which we also do) isolates the rest of the service, but the agent's own turn still takes 14 s. The user is still waiting.

  If the interface surface is the concern, I'm happy to shrink it — e.g. accept only a `(pattern, flags) => { test, exec }` factory and keep re2js as the sole in-tree engine.

Issue #408, opened 2026-09-03
Missing coreutils: realpath and yes (both 127)


trieloff:
## Summary

`realpath` and `yes` are not available in the shell, though both are standard POSIX/coreutils utilities that scripts commonly assume.

## Reproduction

```console
$ realpath --version
bash: realpath: command not found
$ yes --version
bash: yes: command not found

$ which realpath yes
(no output, rc=1)
```

Both exit 127. Confirmed with clean non-interactive `bash -c` invocations, so this is not shell-state or PATH noise.

## Why they are worth having

- **`realpath`** — canonicalising a path (resolving `..`, `.`, symlinks) is a routine scripting need, and there is no clean substitute. The usual fallback `cd "$(dirname "$x")" && pwd` is clumsy and changes directory state. Agent-authored scripts reach for `realpath` frequently.
- **`yes`** — the standard way to feed a repeating stream into a pipe (`yes | head -5`, or auto-confirming an interactive prompt). Trivial to implement: write the argument (default `y`) plus a newline in a loop until the pipe closes.

`yes` in particular is a very small amount of code, and its absence tends to hang or break scripts in non-obvious ways rather than failing clearly.

## Context

Found while auditing published command documentation for a site that documents this shell's command surface — both commands were listed as available, and turned out not to exist. Filing here rather than downstream because they are generic shell utilities with no browser/VFS-specific semantics.

Related SLICC-specific gaps (`patch`, `uptime`, `id`, and the `jobs`/`bg`/`fg`/`trap`/`select` builtins) are filed against `ai-ecoverse/slicc` separately, since those need runtime decisions rather than a utility implementation.

## Environment

`just-bash` running as the SLICC browser shell (WASM). Observed 2026-09-03.

PR #409, opened 2026-09-03
feat(yes): implement the yes command

trieloff:
Fixes the `yes` half of #408 (`realpath` follows in a separate PR).

`yes` exited 127, so neither `yes | head -5` nor auto-confirming a prompt worked.

## What it does

- Repeats a line built from the operands joined by single spaces, `y` when there are none.
- GNU option handling: `--help`, `--` ends options, a lone `-` is an operand, and any other dash-prefixed argument is an unknown option even when it follows an operand (getopt permutes) — `yes: invalid option -- 'x'` / `yes: unrecognized option '--forever'`, exit 1.

## The one deliberate deviation

Real `yes` writes until its reader goes away. A pipeline stage here runs to completion before the next one starts, so "until the reader goes away" has no meaning and an unbounded `yes` would just be a hang. The stream is therefore finite:

- it stops after `executionLimits.maxLoopIterations` lines (100k normal, 10k hardened), and
- earlier when the repeated line would no longer fit in `min(maxStringLength, maxOutputSize)`.

`yes | head -3`, `yes n | head -2`, `yes | head -5000 | wc -l` are all exact against real GNU `yes`. A consumer that wants more lines than the bound sees a short stream, which is the trade for not hanging. The bound is spelled out in `yes --help`.

If a single line cannot fit the output budget at all, the existing output-limit error surfaces (`bash: yes: output size limit exceeded (N bytes)`, exit 126) rather than silently printing nothing.

## Tests

- `src/commands/yes/yes.test.ts` — 19 unit tests: operands, `--`, lone `-`, both limit paths, UTF-8 byte accounting, option errors, redirection, `which yes`.
- `src/comparison-tests/yes.comparison.test.ts` + fixtures — 8 cases against real `yes`. Recorded with GNU coreutils on `PATH` (BSD `yes` ignores every operand but the first) and locked so a macOS re-record cannot swap in BSD output.

## Validation

`pnpm typecheck`, `pnpm lint:fix`, `pnpm lint:banned`, `pnpm knip`, `pnpm test:unit`, `pnpm test:comparison` all pass. Five python3/WASM-bundle test files fail identically on unmodified `main` in this environment (no built `dist`), and are untouched by this change.

### DIFF
--- a/.changeset/yes-command.md
+++ b/.changeset/yes-command.md
@@ -1,0 +1,5 @@
+---
+"just-bash": minor
+---
+
+Add the `yes` command. `yes [STRING]...` repeats a line built from its operands (`y` when there are none), so `yes | head -3` and `yes | some-prompt` work instead of failing with exit 127. Because pipeline stages here run to completion rather than streaming, the stream is finite: it ends after `executionLimits.maxLoopIterations` lines, or earlier if the repeated line would exceed the output size limit.
--- a/packages/just-bash/AGENTS.npm.md
+++ b/packages/just-bash/AGENTS.npm.md
@@ -80,7 +80,7 @@ const result = await bash.exec("cat input.txt | grep pattern");
 
 **File operations**: `basename`, `chmod`, `cp`, `dirname`, `du`, `file`, `find`, `ln`, `ls`, `mkdir`, `mv`, `od`, `pwd`, `readlink`, `rm`, `rmdir`, `split`, `stat`, `touch`, `tree`
 
-**Utilities**: `alias`, `base64`, `bash`, `clear`, `curl`, `date`, `diff`, `echo`, `env`, `expr`, `false`, `gzip`, `gunzip`, `help`, `history`, `hostname`, `html-to-markdown`, `md5sum`, `printenv`, `printf`, `seq`, `sh`, `sha1sum`, `sha256sum`, `sleep`, `tar`, `tee`, `time`, `timeout`, `true`, `unalias`, `which`, `whoami`, `zcat`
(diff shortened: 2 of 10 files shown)


trieloff:
Thanks for the review. All three findings are addressed in f517e015, each checked against GNU coreutils `yes` first.

- **Grouped short options.** `yes -xy` now reports `yes: invalid option -- 'x'`, which is what GNU does (getopt stops at the letter it rejected). Long options still report the whole `--name`.
- **Unbounded join.** The line is now built with `boundedJoin`, so many large operands are charged before the string is constructed instead of after. The newline stays charged to the repetition so the reported limit equals the configured one rather than `limit - 1`.
- **`maxLoopIterations: 0`.** No longer forced up to one line: zero permitted iterations now emits nothing and exits 0, so the documented bound holds at zero. A single line that cannot fit the size limit is still attempted, so the limit is reported rather than silently printing nothing.

Four new unit tests cover these (grouped option, zero iterations, join-charged limit, newline charged to the line), 23 in the file total.

On the portability point about `Infinity` limits: `BoundedStringBuilder` rejects a non-finite `maxBytes`, and that is repo-wide rather than something `yes` introduces — `set` also exits 126 under `executionLimits: { maxOutputSize: Infinity }`. Happy to fix it in `bounded-builder.ts` in a separate PR if you want it treated as a bug rather than as a limit configuration that is out of contract.

The failing checks are the Vercel deploys, which need a Vercel Labs member to authorize a deploy from a fork. Every GitHub Actions check passes.

cramforce:
Please sign the commit!
@auto-maintain review

trieloff:
Signed. Both commits on this branch are now GPG-signed with my key and show as Verified: the author stays `Claude Code <noreply@anthropic.com>` and I am the committer, which is what lets GitHub verify the signature against my account.

The rebase also picked up the current `main`, so the branch is up to date. The tree is byte-identical to what was reviewed before the rewrite (verified with `git diff` against the old tip), and I re-ran typecheck plus the unit and comparison suites on the new base: all green apart from the pre-existing python3/WASM failures that also fail on unmodified `main` in my environment.

@auto-maintain review

PR #410, opened 2026-09-03
feat(realpath): implement the realpath command

trieloff:
Fixes the `realpath` half of #408. The `yes` half is #409.

Canonicalizing a name had no clean substitute: `realpath` exited 127, and the usual fallback `cd "$(dirname "$x")" && pwd` moves the shell.

## Behaviour

Resolution walks the path one component at a time against the virtual filesystem instead of delegating to `fs.realpath()`, because each mode needs its own existence and symlink policy and a failure has to carry the errno GNU prints.

| mode | policy |
| --- | --- |
| default | every component but the last must exist |
| `-e` / `--canonicalize-existing` | all of it must exist, including through the final symlink |
| `-m` / `--canonicalize-missing` | none of it need exist, and nothing fails: a symlink loop leaves the link unexpanded rather than reporting ELOOP, matching GNU |
| `-s` / `--strip` / `--no-symlinks` | purely textual; existence checked only under `-e`, and then by following the link |
| `-L` / `-P` | `..` cancelled before or after symlink expansion |

Also `-q`, `-z`, `--relative-to=DIR` and `--relative-base=DIR` (value inline or as the next argument), including GNU's rule that an anchor outside the base drops both options and prints absolute names, and `.` for a name that is the anchor itself. `-e`/`-m` and `-L`/`-P` are last-one-wins, short flags bundle (`-se`, `-zm`), `--` ends options, a lone `-` is an operand.

Diagnostics match GNU: `No such file or directory`, `Not a directory` for a path descending through a regular file (including a regular file named with a trailing slash), `Too many levels of symbolic links`, `realpath: '': No such file or directory` for the empty name, `realpath: missing operand`, and `realpath: option '--relative-to' requires an argument`. Unknown options use the repository's `unknownOption` wording.

## Bounds

Symlink expansion is capped at the existing `MAX_SYMLINK_DEPTH` and every component visit is metered through `FileTraversalBudget`, so a link farm cannot spin. Output goes through `BoundedStringBuilder` against `min(maxStringLength, maxOutputSize)`.

## Tests

- `realpath.basic.test.ts`, `realpath.symlinks.test.ts`, `realpath.options.test.ts` — 70 unit tests.
- `realpath.comparison.test.ts` + fixtures — 26 cases against real GNU `realpath`, all passing on the first recording. Cases print names relative to the test directory, since the absolute answer necessarily differs between the real temporary directory and the VFS. Recorded with GNU coreutils on `PATH` (BSD `realpath` only understands `-q`) and locked so a macOS re-record cannot replace them.

## Validation

`pnpm typecheck`, `pnpm lint:fix`, `pnpm lint:banned`, `pnpm knip`, `pnpm test:unit`, `pnpm test:comparison` all pass. Five python3/WASM-bundle test files fail identically on unmodified `main` in this environment and are untouched by this change.

### DIFF
--- a/.changeset/realpath-command.md
+++ b/.changeset/realpath-command.md
@@ -1,0 +1,5 @@
+---
+"just-bash": minor
+---
+
+Add the `realpath` command. Canonicalizes names against the virtual filesystem — resolving `.`, `..` and symlinks — instead of exiting 127, with GNU's `-e`, `-m`, `-s`/`--no-symlinks`, `-L`, `-P`, `-q`, `-z`, `--relative-to` and `--relative-base`, GNU's default policy that every component but the last must exist, and GNU's diagnostics for a missing component, a path descending through a regular file, and a symlink loop.
--- a/packages/just-bash/AGENTS.npm.md
+++ b/packages/just-bash/AGENTS.npm.md
@@ -78,7 +78,7 @@ const result = await bash.exec("cat input.txt | grep pattern");
 
 **Data processing**: `jq` (JSON), `js-exec` (JavaScript/TypeScript via QuickJS), `python3`/`python` (Python via WASM/CPython), `sqlite3` (SQLite), `xan` (CSV), `yq` (YAML/XML/TOML/CSV)
 
-**File operations**: `basename`, `chmod`, `cp`, `dirname`, `du`, `file`, `find`, `ln`, `ls`, `mkdir`, `mv`, `od`, `pwd`, `readlink`, `rm`, `rmdir`, `split`, `stat`, `touch`, `tree`
+**File operations**: `basename`, `chmod`, `cp`, `dirname`, `du`, `file`, `find`, `ln`, `ls`, `mkdir`, `mv`, `od`, `pwd`, `readlink`, `realpath`, `rm`, `rmdir`, `split`, `stat`, `touch`, `tree`
 
 **Utilities**: `alias`, `base64`, `bash`, `clear`, `curl`, `date`, `diff`, `echo`, `env`, `expr`, `false`, `gzip`, `gunzip`, `help`, `history`, `hostname`, `html-to-markdown`, `md5sum`, `printenv`, `printf`, `seq`, `sh`, `sha1sum`, `sha256sum`, `sleep`, `tar`, `tee`, `time`, `timeout`, `true`, `unalias`, `which`, `whoami`, `zcat`
 
--- a/packages/just-bash/README.md
+++ b/packages/just-bash/README.md
@@ -102,7 +102,7 @@ duplicating internal defaults.
 
 ### Navigation & Environment
 
-`basename`, `cd`, `dirname`, `du`, `echo`, `env`, `export`, `find`, `hostname`, `printenv`, `pwd`, `tee`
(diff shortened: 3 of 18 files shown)


trieloff:
Thanks for the thorough review. All six findings are addressed in c6c1558d. I verified every case against GNU coreutils `realpath` before changing behaviour, and recorded the results as 21 locked comparison fixtures (`realpath-path-semantics.comparison.test.ts`) plus 35 new unit tests.

- **`.` was dropped when splitting.** It is not a no-op: GNU requires a directory in front of it. `file/.` and `file/..` are now `Not a directory`, `nosuch/.` is `No such file or directory`, while a trailing slash keeps tolerating an absent name (`nosuchdir/` still resolves).
- **`-s` resolved textually.** It now walks components. GNU still refuses to descend through a non-directory (`-s link/..`, `-s file/x`, `-s file/`) while following a link for the directory test only (`-s dirlink/deep` passes), and still checks no intermediate component for existence; `-e` remains a check on the finished name.
- **Symlink targets ending in `/`.** The directory requirement is kept, so a link to `file/` is `Not a directory` while a link to `nosuch/` resolves.
- **`-L` did not validate what it cancels.** GNU rejects `-L link/..` for a link to a file and `-L nosuch/..` for a missing name, so cancelling an `X/..` pair now requires `X` to be an existing directory. Skipped under `-m`, which never fails.
- **Errors read as "missing".** Failures are classified now: only ENOENT means missing, ENOTDIR/ELOOP/EACCES get their own diagnostic (`Permission denied` was missing from the table), an unexplained failure no longer becomes a resolvable name, and `ExecutionLimitError`/`ExecutionAbortedError` propagate instead of being swallowed.
- **Absolute symlink targets escaped their mount.** `MountableFs` stores them relative to the mounted filesystem's root, so resolution asks the filesystem and keeps the mount prefix. `realpath /data/link` now names `/data/secret.txt`, and `cat "$(realpath /data/link)"` reads the same bytes as `cat /data/link`.
- **Bounds.** An oversized symlink target is charged to the traversal budget before the expanded path is allocated (components are concatenated rather than spread), and failure diagnostics are built with a `BoundedStringBuilder` — an operand can fail without touching the filesystem, so nothing else bounded them.

One deliberate limitation, covered by a test that documents it: a *dangling* absolute target inside a mount still falls back to the global root, because there is nothing to resolve through and `IFileSystem` offers no way to ask which mount a path belongs to. Resolving it properly needs a mount-attribution API; happy to add one if you would rather have that.

The failing checks are the Vercel deploys, which need a Vercel Labs member to authorize a deploy from a fork. Every GitHub Actions check passes.

cramforce:
@auto-maintain review

Please sign the commits!

trieloff:
Signed. Both commits on this branch are now GPG-signed with my key and show as Verified: the author stays `Claude Code <noreply@anthropic.com>` and I am the committer, which is what lets GitHub verify the signature against my account.

The rebase also picked up the current `main`, so the branch is up to date. The tree is byte-identical to what was reviewed before the rewrite (verified with `git diff` against the old tip), and I re-ran typecheck plus the unit and comparison suites on the new base: all green apart from the pre-existing python3/WASM failures that also fail on unmodified `main` in my environment.

@auto-maintain review

PR #413, opened 2026-09-07
fix(diff): default to POSIX normal format

### LINKED ISSUE #412: diff defaults to unified format (not POSIX normal); `diff a b | grep '^<'` silently matches nothing, and normal format is unreachable
## Summary

`diff` with no flags emits **unified** format instead of POSIX **normal** format (`<` / `>`). GNU diffutils and POSIX both default to normal format; unified is opt-in via `-u`.

Three concrete defects fall out of this:

1. **Normal format is unreachable.** `-u` is parsed and then discarded (`void parsed.result.flags.unified`), and `createTwoFilesPatch` is the only output path — so there is no flag that produces `<` / `>`.
2. **`-c` and `--version` don't exist.** Context format is unavailable and `--version` errors out.
3. **A non-standard 67-char `===…===` banner** is the first line of output — a `createTwoFilesPatch` artifact that neither GNU normal nor GNU unified output produces, and which breaks parsers of either format.

The practical impact is a **silent** false negative: `diff a b | grep '^<'` — the canonical "lines only in A" idiom — matches nothing and exits cleanly. In the case that led me here, a file differing by 3,565 bytes read as "0 changed lines" until a `cmp` contradicted it. There is no error and no warning; a script or agent just gets the wrong answer.

## Reproduction

Bare `just-bash@3.4.2` on Node 26, no host shell involved:

```js
import { Bash, InMemoryFs } from 'just-bash';

const fs = new InMemoryFs();
await fs.writeFile('/d1.txt', 'a\nb\nc\n');
await fs.writeFile('/d2.txt', 'a\nX\nc\n');
await fs.writeFile('/d3.txt', 'a\nb\nc\n');
const bash = new Bash({ fs });

for (const c of [
  'diff /d1.txt /d2.txt',
  "diff /d1.txt /d2.txt | grep -cE '^[<>]'",
  "diff /d1.txt /d2.txt | grep -cE '^[-+@]'",
  'diff -u /d1.txt /d2.txt',
  'diff -c /d1.txt /d2.txt',
  'diff --normal /d1.txt /d2.txt',
  'diff --version',
]) {
  const r = await bash.exec(c);
  console.log('### ' + c);
  process.stdout.write(r.stdout + (r.stderr ? '[stderr] ' + r.stderr : ''));
  console.log('exit=' + r.exitCode);
}
```

### just-bash 3.4.2 vs GNU diffutils 3.12

Same two input files throughout (`a\nb\nc\n` vs `a\nX\nc\n`).

**No flags:**

| just-bash 3.4.2 | GNU diffutils 3.12 |
| --- | --- |
| <pre>===================================================================<br>--- /d1.txt&#9;<br>+++ /d2.txt&#9;<br>@@ -1,3 +1,3 @@<br> a<br>-b<br>+X<br> c<br>exit=1</pre> | <pre>2c2<br>&lt; b<br>---<br>&gt; X<br>exit=1</pre> |

**The idiom that silently breaks:**

```
                                          just-bash   GNU
diff d1 d2 | grep -cE '^[<>]'    →              0       2      # normal-format markers
diff d1 d2 | grep -cE '^[-+@]'   →              5       0      # unified-format markers
```

**`-u` is a no-op** — byte-identical to the no-flag invocation, confirming unified is already the default:

```js
const a = await bash.exec('diff /d1.txt /d2.txt');
const b = await bash.exec('diff -u /d1.txt /d2.txt');
a.stdout === b.stdout   // true
```

GNU `-u` for comparison (note: no `===` banner, and real timestamps in the header):

```
--- d1.txt	2026-09-07 16:45:17.720325084 +0200
+++ d2.txt	2026-09-07 16:45:17.720416501 +0200
@@ -1,3 +1,3 @@
 a
-b
+X
 c
```

**Normal format is unreachable, and `-c` / `--version` are missing:**

| command | just-bash 3.4.2 | GNU diffutils 3.12 |
| --- | --- | --- |
| `diff --normal a b` | `diff: unrecognized option '--normal'` (exit 1) | normal format (exit 1) |
| `diff -c a b` | `diff: invalid option -- 'c'` (exit 1) | context format (exit 1) |
| `diff --version` | `diff: unrecognized option '--version'` (exit 1) | `diff (GNU diffutils) 3.12` (exit 0) |

GNU `-c` output, for reference:

```
*** d1.txt	Mon Sep  7 16:45:17 2026
--- d2.txt	Mon Sep  7 16:45:17 2026
***************
*** 1,3 ****
  a
! b
  c
--- 1,3 ----
  a
! X
  c
```

**What already matches GNU** (no change wanted here): `-q` → `Files a and b differ` (exit 1), `-s` on identical files → `Files a and b are identical` (exit 0), missing operand → exit 2, unreadable file → exit 2.

## Cause

[`packages/just-bash/src/commands/diff/diff.ts`](https://github.com/vercel-labs/just-bash/blob/main/packages/just-bash/src/commands/diff/diff.ts):

```ts
// L56-57 — the flag is parsed, then explicitly thrown away
// Note: unified flag is accepted but is the default behavior
void parsed.result.flags.unified;

// L121-123 — the only non-brief output path
const output = Diff.createTwoFilesPatch(f1, f2, c1, c2, "", "", { context: 3 });
```

`argDefs` (L28-37) declares only `unified` / `brief` / `reportSame` / `ignoreCase`, so `--normal`, `-c`, and `--version` never reach a handler.

The banner is jsdiff's, not just-bash's — `createTwoFilesPatch` unconditionally prefixes it:

```
$ node -e "console.log(JSON.stringify(require('diff').createTwoFilesPatch('d1.txt','d2.txt','a\nb\nc\n','a\nX\nc\n','','',{context:3})))"
"===================================================================\n--- d1.txt\t\n+++ d2.txt\t\n@@ -1,3 +1,3 @@\n a\n-b\n+X\n c\n"
```

(just-bash declares `"diff": "^8.0.4"`; resolved to 8.0.4 here.) The `\t` with nothing after it is the empty `oldHeader`/`newHeader` — GNU puts a timestamp there.

## Suggested fix

In preference order:

1. **Default to normal format and make `-u` mean something.** Emit `NcN` / `<` / `---` / `>` when no format flag is given, and route `-u` to the unified path. This is the spec-correct, least-surprise behavior, and it's what every script and every LLM writing shell expects. `Diff.structuredPatch` already gives the hunk data needed to render normal format — no new dependency.
2. If unified-by-default is a deliberate choice, then at minimum: make `--normal` and `-c` work so the other formats are reachable, and **drop the `===` banner** regardless — it is not valid output in any GNU format and breaks patch parsers. Post-process `createTwoFilesPatch` output, or build the header from `structuredPatch` directly.

Either way, `--version` would be a cheap addition.

I'm happy to send a PR if you'd like to point me at a preferred direction on (1) vs (2).

## Environment

- `just-bash@3.4.2` (latest published), Node 26.7.0, macOS arm64
- Reference: GNU diffutils 3.12 (`brew install diffutils`)
- Originally surfaced in [SLICC](https://github.com/ai-ecoverse/slicc), which uses just-bash as its shell runtime — but reproduced above against a bare just-bash install with no host involvement.

trieloff:
Fixes #412.

## The bug

`diff` with no flags emitted **unified** format. Three defects fell out of that:

1. **Normal format was unreachable.** `-u` was parsed and then discarded (`void parsed.result.flags.unified`), and `createTwoFilesPatch` was the only non-brief output path — so no flag produced `<` / `>`.
2. **`-c` and `--version` did not exist.**
3. **A 67-character `===…===` banner** led every invocation — a `createTwoFilesPatch` artifact that is not valid output in any GNU format.

The practical impact was a *silent* false negative: `diff a b | grep '^<'`, the canonical "lines only in A" idiom, matched nothing and exited cleanly on files that differ.

This takes option 1 from the issue.

## What changed

- **Normal format is the default** (`NcN` / `NaN` / `NdN`, `<`, `---`, `>`), matching GNU and POSIX. `--normal` selects it explicitly.
- **`-u` / `--unified` now selects unified format** instead of being thrown away.
- **New `-c` / `--context`** for context format (`*** ` / `--- ` / `***************` / `! `), including GNU's rule that a hunk's old or new block body is omitted when that side has no changes.
- **New `--version`**, exit 0.
- **The `===` banner is gone** from every format; `-u` / `-c` headers are built directly rather than post-processed.
- **Filenames in `-u` / `-c` headers are quoted and escaped as GNU does** — see the security note below.
- **Two output styles at once is an error**, as in GNU: `diff: conflicting output style options`, exit 2.
- `\ No newline at end of file` is reported in all three formats, on the correct side, in the right position.
- `flagsForFuzzing` and the `--help` options block are updated; help no longer claims unified is the default.

Unchanged and covered by existing tests: `-q`, `-s`, `-i`, `-` for stdin, missing operand → 2, unreadable file → 2, identical files → empty stdout and exit 0.

`package.json` and the lockfile are untouched.

## How the formats were derived

The three formatters were written by running GNU diffutils 3.12 and matching the bytes it emits — no GNU source was used for them. The new tests freeze that output verbatim for: a single changed line, insertions at the middle/start/end, deletions at the middle/start, a multi-line deletion, two hunks far enough apart to stay separate, two changes close enough to merge into one, an insertion and deletion sharing a hunk, empty vs non-empty in both directions, all three trailing-newline combinations, a blank context line, and a line changed to a blank line.

## Two deliberate deviations from GNU

**1. No timestamp in the `-u` / `-c` header.** GNU emits `--- d1\t2026-09-07 16:45:17.720325084 +0200`; this emits `--- d1`. Inside a virtual filesystem the mtime is often synthetic or absent (stdin has none), and a timestamp would make output non-reproducible for callers and for this repo's fixture-based comparison tests. Patch consumers ignore the field. Happy to emit mtimes instead if you'd rather have exact header parity.

**2. Hunk grouping can differ on ambiguous input.** Several equally minimal edit scripts usually exist for the same pair of files, and which one you get depends on the implementation. Edit-script selection still comes from the `diff` package, so on some inputs the `NcN` line numbers group differently than GNU's. Measured over 199 mutated source-file pairs:

- **149 / 199** unified diffs are byte-identical to GNU
- **199 / 199** apply cleanly under GNU `patch` and reproduce the target file exactly
- **0** differ from GNU in edit count — every diff is equally minimal

So the output is always a correct, minimal patch; only the grouping of ambiguous changes can differ. This is unchanged from the current behaviour on `main`, and it is called out in the changeset.

## On the automated review

Thanks — the GPL finding was correct and serious, and I've dealt with it.

**GPL (🔴 blocker) — fixed by rewriting the branch.** The first push contained `myers.ts`, a line-by-line port of diffutils' `compareseq`, `diag`, `shift_boundaries` and `discard_confusing_lines`, plus a port of the `find_identical_ends` trim in `format.ts`. It preserved structure, identifiers and comments, so it was plainly a derivative of GPLv3 code and could not ship in an Apache-2.0 package. That was my mistake. The branch has been **force-pushed as a single clean commit** so the GPL-derived code is not in this PR's history at all, not merely removed by a follow-up. Edit-script selection is back on the `diff` package (BSD-3-Clause), which stays a dependency. The formatters and their tests are unaffected — they came from observed output, and GNU is explicit that a program's output is not covered by the GPL. The cost is deviation 2 above.

**`--version` staleness (🔴) — fixed.** It now prints `diff (just-bash)` with no number. A hard-coded `3.4.2` would have been wrong the moment this minor changeset published 3.5.0, and this command has no release cadence of its own. Build-time injection is possible (everything shipped as runnable JS is esbuild-bundled), but it means threading a `--define` through five build scripts plus a from-source fallback, which felt disproportionate for one string — happy to do it if you'd prefer a real version.

**Filename injection (🟡) — fixed.** Good catch. `-u` / `-c` headers now quote and escape filenames, so a name containing a newline can no longer forge header or `@@` lines in the patch stream. The rule was taken from GNU's observed behaviour in a UTF-8 locale: quote when the name contains whitespace, a `"` or a control character; escape `\`, `"`, `\t`, `\n`, `\r` and `\NNN`; leave non-ASCII bare (GNU only octal-escapes it in the C locale). 21/21 differential cases match GNU, including tab, CR, `\x01`, quotes, backslashes and multi-byte names. Normal format prints no filenames at all, so it has no such surface. Note that `-q` / `-s` still print filenames unquoted; GNU shell-quotes them there (`Files 'with space' and plain differ`), but that is pre-existing behaviour the issue explicitly asked me not to change, so I left it — happy to fix separately.

**`-q` conflict (🟡) — not applied, the finding looks incorrect.** GNU does not treat `-q` as a conflicting output style. Against diffutils 3.12:

```
$ diff -q -u a b   → Files a and b differ    (exit 1)
$ diff -u -q a b   → Files a and b differ    (exit 1)
$ diff -q -c a b   → Files a and b differ    (exit 1)
$ diff -q --normal a b → Files a and b differ (exit 1)
```

Only two *format* styles together conflict (`diff -u -c` → exit 2), which is what the code implements. Happy to be shown otherwise.

## Testing

`pnpm typecheck`, `pnpm lint` (incl. banned-patterns), `pnpm knip` and `pnpm build` are clean. `pnpm test:run --exclude src/spec-tests` passes 11,689 tests; the 6 failures are CPython-WASM sandbox tests that fail identically on a clean `origin/main` worktree here, so they are pre-existing and unrelated.

### DIFF
--- a/.changeset/diff-normal-format-default.md
+++ b/.changeset/diff-normal-format-default.md
@@ -1,0 +1,19 @@
+---
+"just-bash": minor
+---
+
+`diff` now defaults to POSIX normal format (`2c2` / `<` / `---` / `>`) instead of unified, matching GNU diffutils and POSIX. **This changes the default output and will break anything parsing the previous unified-by-default output** — pass `-u` to keep unified.
+
+Previously `-u` was parsed and discarded, `createTwoFilesPatch` was the only output path, and every invocation was prefixed with jsdiff's 67-character `===...===` banner, which is not valid output in any GNU format. Normal format was unreachable by any flag, so `diff a b | grep '^<'` — the canonical "lines only in A" idiom — silently matched nothing and exited cleanly on files that differ.
+
+- `-u` / `--unified` now selects unified format, and `--normal` selects the default explicitly.
+- New `-c` / `--context` for context format (`*** ` / `--- ` / `***************` / `! `).
+- New `--version`.
+- The `===` banner is gone from every format. The `-u` / `-c` file headers carry no timestamp, so output is reproducible; GNU puts each file's mtime there.
+- Filenames in `-u` / `-c` headers are quoted and escaped as GNU does, so a filename containing a newline can no longer forge header or `@@` lines in the emitted patch.
+- Two output styles at once (for example `diff -u -c`) is now an error, as in GNU: `diff: conflicting output style options`, exit 2.
+- `\ No newline at end of file` is reported in all three formats.
+
+`-q`, `-s`, `-i`, `-` for stdin, and the exit codes are unchanged.
+
+Hunk selection is unchanged: where several equally minimal edit scripts exist, just-bash may group ambiguous changes differently than GNU. The output is always a correct, equally minimal patch, but the `NcN` line numbers can differ on such inputs.
--- a/packages/just-bash/src/commands/diff/diff.gnu-context.test.ts
+++ b/packages/just-bash/src/commands/diff/diff.gnu-context.test.ts
@@ -1,0 +1,223 @@
+import { describe, expect, it } from "vitest";
+import { Bash } from "../../Bash.js";
+
+/**
+ * Context format, selected with -c.
+ *
+ * Every expected string in this file was produced by running GNU diffutils
+ * 3.12 on the same two inputs and freezing its output verbatim,
+ * with only the two header lines replaced (GNU stamps them with the files'
+ * mtimes; just-bash omits the timestamp so its output is reproducible).
+ */
+describe("diff context format (-c) vs GNU diffutils 3.12", () => {
+  describe("context format (-c)", () => {
+    it("matches GNU for a single changed line", async () => {
+      const env = new Bash({
+        files: { "/a.txt": "a\nb\nc\n", "/b.txt": "a\nX\nc\n" },
+      });
+      const result = await env.exec("diff -c /a.txt /b.txt");
+      expect(result.stdout).toBe(
+        "*** /a.txt\n--- /b.txt\n***************\n*** 1,3 ****\n  a\n! b\n  c\n--- 1,3 ----\n  a\n! X\n  c\n",
+      );
+      expect(result.stderr).toBe("");
+      expect(result.exitCode).toBe(1);
+    });
+
+    it("matches GNU for a pure insertion in the middle", async () => {
+      const env = new Bash({
+        files: { "/a.txt": "l1\nl2\nl3\n", "/b.txt": "l1\nNEW\nl2\nl3\n" },
+      });
+      const result = await env.exec("diff -c /a.txt /b.txt");
+      expect(result.stdout).toBe(
+        "*** /a.txt\n--- /b.txt\n***************\n*** 1,3 ****\n--- 1,4 ----\n  l1\n+ NEW\n  l2\n  l3\n",
+      );
+      expect(result.stderr).toBe("");
+      expect(result.exitCode).toBe(1);
+    });
+
+    it("matches GNU for an insertion at the end", async () => {
+      const env = new Bash({
+        files: { "/a.txt": "a\nb\n", "/b.txt": "a\nb\nc\nd\n" },
+      });
+      const result = await env.exec("diff -c /a.txt /b.txt");
+      expect(result.stdout).toBe(
+        "*** /a.txt\n--- /b.txt\n***************\n*** 1,2 ****\n--- 1,4 ----\n  a\n  b\n+ c\n+ d\n",
+      );
+      expect(result.stderr).toBe("");
+      expect(result.exitCode).toBe(1);
+    });
+
+    it("matches GNU for an insertion at the start", async () => {
+      const env = new Bash({
+        files: { "/a.txt": "b\nc\n", "/b.txt": "a\nb\nc\n" },
+      });
+      const result = await env.exec("diff -c /a.txt /b.txt");
+      expect(result.stdout).toBe(
+        "*** /a.txt\n--- /b.txt\n***************\n*** 1,2 ****\n--- 1,3 ----\n+ a\n  b\n  c\n",
+      );
+      expect(result.stderr).toBe("");
+      expect(result.exitCode).toBe(1);
+    });
+
+    it("matches GNU for a pure deletion in the middle", async () => {
+      const env = new Bash({
+        files: { "/a.txt": "l1\nl2\nl3\n", "/b.txt": "l1\nl3\n" },
+      });
+      const result = await env.exec("diff -c /a.txt /b.txt");
+      expect(result.stdout).toBe(
+        "*** /a.txt\n--- /b.txt\n***************\n*** 1,3 ****\n  l1\n- l2\n  l3\n--- 1,2 ----\n",
+      );
+      expect(result.stderr).toBe("");
+      expect(result.exitCode).toBe(1);
+    });
+
+    it("matches GNU for a deletion at the start", async () => {
+      const env = new Bash({
+        files: { "/a.txt": "a\nb\nc\n", "/b.txt": "b\nc\n" },
+      });
+      const result = await env.exec("diff -c /a.txt /b.txt");
+      expect(result.stdout).toBe(
+        "*** /a.txt\n--- /b.txt\n***************\n*** 1,3 ****\n- a\n  b\n  c\n--- 1,2 ----\n",
+      );
+      expect(result.stderr).toBe("");
+      expect(result.exitCode).toBe(1);
+    });
+
+    it("matches GNU for a multi-line deletion", async () => {
+      const env = new Bash({
+        files: { "/a.txt": "a\nb\nc\nd\n", "/b.txt": "a\nd\n" },
+      });
+      const result = await env.exec("diff -c /a.txt /b.txt");
+      expect(result.stdout).toBe(
+        "*** /a.txt\n--- /b.txt\n***************\n*** 1,4 ****\n  a\n- b\n- c\n  d\n--- 1,2 ----\n",
+      );
+      expect(result.stderr).toBe("");
+      expect(result.exitCode).toBe(1);
+    });
+
+    it("matches GNU for two hunks far enough apart to stay separate", async () => {
+      const env = new Bash({
+        files: {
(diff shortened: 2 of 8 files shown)


trieloff:
The responsible agent has received a stern talking to and will ensure license compliance.

trieloff:
Force-pushed as a single clean commit ([`3185241`](https://github.com/vercel-labs/just-bash/commit/3185241860ec85ac326a2d3d97005572a4b00dac)) to address the review.

**The GPL finding was right, and it was a serious mistake on my part.** The first push contained `myers.ts`, a line-by-line port of GNU diffutils' `compareseq`, `diag`, `shift_boundaries` and `discard_confusing_lines`, plus a port of `find_identical_ends` inside `format.ts`. It kept the original structure, identifiers and comments — unambiguously a derivative of GPLv3 code, which cannot ship in an Apache-2.0 package. I rewrote the branch rather than adding a removal commit, so the GPL-derived code is not in this PR's history at all. `package.json` and the lockfile are back to matching `main`; `diff` (BSD-3-Clause) stays the dependency for edit-script selection.

What survives is the format work, which was written from the bytes GNU emits rather than from its source — the frozen GNU-output expectations all still pass unchanged. The cost is that hunk grouping is no longer GNU-identical on ambiguous input. Measured over 199 mutated source-file pairs: 149/199 unified diffs byte-identical to GNU, **199/199 apply cleanly under `patch` and reproduce the target file**, and **0** differ from GNU in edit count. Correct and equally minimal every time; only the grouping of ambiguous changes can differ. That is called out in the changeset and the PR body.

Also addressed:

- **`--version`** now prints `diff (just-bash)` with no number, since a hard-coded `3.4.2` would have been stale the moment this minor changeset published 3.5.0. Build-time injection is doable if you'd rather have a real number — say the word.
- **Filename injection** fixed: `-u`/`-c` headers quote and escape filenames, so a newline in a name can no longer forge header or `@@` lines. 21/21 differential cases match GNU (tab, CR, control chars, quotes, backslashes, non-ASCII). `-q`/`-s` still print names unquoted — GNU shell-quotes there, but that's pre-existing output the issue asked me not to change, so I left it for a separate PR.

One finding I did **not** apply, because it appears to be incorrect: **`-q` is not a conflicting output style in GNU.** Against diffutils 3.12, `diff -q -u`, `diff -u -q`, `diff -q -c` and `diff -q --normal` all print `Files a and b differ` and exit 1 — no conflict error. Only two *format* styles together conflict (`diff -u -c` → exit 2), which is what's implemented. Happy to be corrected if you're seeing different behaviour.

cramforce (approved):

PR #414, opened 2026-09-08
fix(find): report a directory it cannot read and keep going

mutewinter:
## Problem

`find` over a tree that holds one directory the process cannot read returns nothing at all.

```js
// /home/user/.Trash exists but readdir on it throws EACCES
await bash.exec("find /home/user -name '*.md'");
// stdout: ""   stderr: "find: EACCES: permission denied, scandir '<path>'\n"   exitCode: 1

await bash.exec("find /home/user -name '*.md' 2>/dev/null | head -1");
// stdout: ""   stderr: ""   exitCode: 0
```

Every macOS home directory holds at least one such directory (`.Trash`, and several under `Library`), so a recursive `find` over a mounted home directory never returns a result. In the piped form, which is the shape an agent writes almost every time, the failure is byte-identical to a genuine no-match: exit 0, no output, no stderr. In a transcript I have from an agent searching a home directory for an Obsidian vault, nine of fourteen shell calls were `find` over the mount and every one that crossed an unreadable directory came back empty; the one aimed at `Library/Application Support` would have landed on `obsidian/obsidian.json`, which names the vault. `rg` walks past the same directory, so the two commands disagree about whether the tree has anything in it, which is what makes this hard to attribute from the inside.

GNU find prints `find: '/home/user/.Trash': Permission denied`, continues with everything else, and exits 1 at the end.

## Cause

`processNode` in `find.ts` wraps `lstat` in a `try`/`catch` and skips the node on failure, but the two `readdir` calls beneath it are bare:

```ts
entriesWithTypes = await ctx.fs.readdirWithFileTypes(currentPath);
...
entries = await ctx.fs.readdir(currentPath);
```

A rejection there propagates out of the `Promise.all` over the batch and out of the traversal, so one unreadable directory ends the whole search. The existing find tests build their trees on `InMemoryFs`, which never refuses a `readdir`, so nothing exercised the path.

## Fix

Both `readdir` calls sit in one `try`/`catch`. On failure the directory is reported on stderr as `find: <path>: <reason>`, `exitCode` becomes 1, and traversal continues; the directory itself is still evaluated and printed, as GNU does. The reason is derived from the errno alone (`EACCES`/`EPERM` read as `Permission denied`; `ENOENT`, `ENOTDIR`, `ELOOP` get their usual phrases; anything else prints its code), so a host path in the underlying message never reaches the output. An error with no errno shape is rethrown, so an abort still aborts.

## Scope

Unchanged: the `lstat` failure path, the traversal budget checkpoints, `-empty` (an unreadable directory is not called empty; `entries` stays null), and the maxdepth optimization, under which a directory at the depth limit is never read and so never reported, as before.

Not addressed, deliberately: `ls -R` over the same tree. I have not checked whether it has the same shape, and it is a separate command.

## Tests

`find.unreadable.test.ts` builds an in-memory home directory behind a `Proxy` whose `readdir` and `readdirWithFileTypes` refuse one path, and asserts: results past the unreadable directory are returned with the message on stderr and exit 1; the directory itself is still listed; the results survive `2>/dev/null | head -1`; `-empty` does not claim it; and `-maxdepth 1` neither reads nor reports it. Four of the five fail without the fix, verified by stashing this change and re-running; the maxdepth case passes either way and is there to pin that behavior. The find suite is 13 files, 226 tests, all passing. The full package suite has hundreds of pre-existing failures in the `js-exec` and `python3` test files in my environment, present on `main` without this change (345 failing there against 325 with it, none in `find`), verified by running `main` separately, not inferred. What the test cannot prove: behavior against a real `ReadWriteFs` on a `chmod 000` directory, which I reproduced by hand but did not automate, since it depends on not running the suite as root.

---

Authored with Claude Fable 5.1

### DIFF
--- a/.changeset/find-unreadable-directory.md
+++ b/.changeset/find-unreadable-directory.md
@@ -1,0 +1,11 @@
+---
+"just-bash": patch
+---
+
+find: report a directory it cannot read and keep going
+
+A `readdir` failure inside the traversal threw out of the whole search, so one unreadable directory ended `find` with no results: exit 1 and `find: EACCES: permission denied, scandir '<path>'` on its own, and exit 0 with nothing at all once piped through `head` with stderr silenced. Every macOS home directory holds such a directory (`.Trash`, several under `Library`), so a recursive `find` over a mounted home directory never returned anything.
+
+GNU find names the directory on stderr, continues with everything else, and exits 1 at the end. It now does the same here. The message is `find: <path>: Permission denied`, with the phrase taken from the errno alone, so nothing from the underlying error's text reaches the output. A failure that is not one of the errnos a directory read can produce (a cancellation, an execution limit, a filesystem policy refusal) still ends the search as before.
+
+Messages are emitted in traversal order beside the node's own output, whatever order the parallel batch settled in, and a failed read still counts toward the trace's `readdirCalls` and `readdirTime`.
--- a/packages/just-bash/src/commands/find/find.ts
+++ b/packages/just-bash/src/commands/find/find.ts
@@ -1186,15 +1186,15 @@ function formatCtimeDate(date: Date): string {
  * carries a code of its own and must end the search, not become a line of
  * stderr, so the phrase comes from this table and never from the error.
  */
-const UNREADABLE_DIRECTORY_REASONS: Record<string, string> = {
-  EACCES: "Permission denied",
-  EIO: "Input/output error",
-  ELOOP: "Too many levels of symbolic links",
-  ENAMETOOLONG: "File name too long",
-  ENOENT: "No such file or directory",
-  ENOTDIR: "Not a directory",
-  EPERM: "Permission denied",
-};
+const UNREADABLE_DIRECTORY_REASONS = new Map<string, string>([
+  ["EACCES", "Permission denied"],
+  ["EIO", "Input/output error"],
+  ["ELOOP", "Too many levels of symbolic links"],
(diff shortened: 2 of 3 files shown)


mutewinter:
Addressed the automated review in 5caeef20, and the lint failure in the same push.

- **Ordering.** The message now rides with its node as an effect, so several unreadable directories report in traversal order rather than in the order the parallel batch settled. Under `-depth` it comes before the directory's own line, where GNU meets the failure on the way down and prints the directory on the way back up; otherwise after it. `find.unreadable.test.ts` pins both orders with two unreadable directories in different subtrees.
- **Trace counters.** A failed read counts toward `readdirCalls` and `readdirTime`, and `resultsFound` no longer counts the diagnostics.
- **Recoverable errors.** Only the errnos a directory read can fail with are recoverable (`EACCES`, `EPERM`, `ENOENT`, `ENOTDIR`, `ELOOP`, `EIO`, `ENAMETOOLONG`), with the phrase taken from a fixed table rather than from the error, so nothing from the error's text reaches stderr. Anything else, including an `ABORT_ERR`, an execution limit, or a filesystem policy refusal, ends the search as before; a test covers that path.
- **Changeset** added.

The lint failure was formatting on the test file. The three `unit-tests` failures are in `timeout.test.ts`, which does not run `find`; they are the cancellation timing cases, and `main` passed them on its last five runs, so I read those as flaky on the runner rather than caused here, but I have not proven it.

PR #420, opened 2026-09-09
fix(stat): expand -c directives instead of printing them literally

mutewinter:
## Problem

```bash
echo hi > f.txt
stat -c '%Y' f.txt       # prints: %Y
stat -c 'mtime=%y' f.txt # prints: mtime=%y
echo $?                  # 0
```

`stat -c '%Y'` is the idiom for reading a modification time, and it returns the format string. Nothing says a value is missing: exit 0, empty stderr, and output that looks like output.

This bites agents rather than humans. A human sees `%Y` and knows; an agent has no reason to distrust a zero exit and a non-empty stdout, so it reports the literal or churns through alternatives. In a production transcript that prompted this, one task spent five of fifteen shell calls on `stat -c '%Y'`, then `date -d @%Y`, then `ls --full-time`, then BSD `stat -f`, before plain `stat` gave it the `Modify:` line. Every directive outside the implemented ten behaves this way, so the failure is not specific to timestamps; timestamps are just the family that gets asked for.

## Cause

`stat.ts` expanded the format with a regex that matched only what it implemented, and returned everything else as itself:

```ts
const output = format.replace(/%[nNsFaAuUgG]/g, (directive) => {
  return replacements.get(directive) ?? directive;
});
```

The `?? directive` fallback is unreachable, since the regex only matches keys already in the map. The real fallthrough is the regex: `%Y` never matches, so it is copied to the output as ordinary text.

The existing tests could not see it because they only assert directives that are implemented. There is no test that asks for one that is not, which is the case where the silent path lives.

Separately, `%a` was `stat.mode.toString(8)`, which carries the file type bits: a 0644 file reported `100644` where GNU reports `644`.

## Fix

FORMAT is scanned rather than regex-replaced, so a directive is parsed as `%`, optional `-` and `0` flags, an optional width, then the directive character. The timestamp family expands, `%%` is a literal percent, widths pad, and a directive with no value prints `?` as GNU does.

`FsStat` carries one timestamp, so `%y` and `%Y` are the only times that can be answered. `%w` / `%W` report `-` and `0`, which is what GNU prints where a filesystem does not record birth time, and the access and change times print `?` with everything else that cannot be answered: reporting mtime for them would be a plausible wrong answer on the read-write backend, where the three genuinely differ and a `chmod` moves only one. `%y` renders through the existing `formatStrftime`, in `$TZ` when Intl accepts it and UTC otherwise, matching the contract `date` already uses, and is built only when a FORMAT names it, so a format without one cannot fail a limit its own output fits.

The broader alternative is extending `FsStat` with the access and change times, allocated blocks, link count, inode and device so `%x`, `%X`, `%z`, `%Z`, `%b`, `%B`, `%h`, `%i`, `%d`, `%D`, `%o` and `%m` can answer truthfully. That is not here: it touches four filesystem implementations, and inventing a plausible value for each is the exact failure this is fixing. Those directives print `?`. Say the word and the `FsStat` version is a small follow-up.

## Scope

Unchanged: `%n`, `%N`, `%s`, `%F`, `%A`, `%u`, `%U`, `%g`, `%G`, the default (no `-c`) output, the `stat: missing operand` and `cannot stat` paths, and the output-size limit behavior. Filenames still interpolate literally, so a name containing `$` or a backtick cannot forge output.

Not addressed, deliberately:

- `--printf`, which interprets backslash escapes and drops the trailing newline. `-c` correctly does not interpret `\n` today, so the family is self-consistent without it; it is maybe ten lines if you want it in the same PR.
- `--format=` as a long spelling of `-c`.
- `%F` for a symlink, which reports `regular file`. `ctx.fs.stat` follows links, so this is only reachable through a path that does not exist in the current code.

## Tests

`commands/stat/stat.format.test.ts`, 15 cases: the `%y` rendering with nanoseconds and offset, `%Y` as whole seconds, access and change left unanswered, birth reported as unrecorded, `$TZ` honored and an unparseable `$TZ` falling back to UTC, `%a` as permission bits for a file and a directory, `%f` in hexadecimal, `?` for an unanswerable directive, `%%`, a trailing `%`, width padding on both sides, a width large enough to trip the output limit, and a timestamp-free FORMAT succeeding under a limit smaller than a timestamp.

`comparison-tests/stat-format.comparison.test.ts`, 6 cases, recorded from GNU coreutils 9.2 and locked, since `stat -c` does not exist on the BSD `stat` a macOS recording would capture. It covers `%s`, `%a %A %f %F` for a file and a directory, `%%` with an unknown directive, width padding, and a trailing `%`. Timestamps are deliberately not in it: they cannot be compared against a recording, so they are unit-tested instead. Re-recording on a Linux box with `RECORD_FIXTURES=force` should reproduce the locked values byte for byte, which is the check I could not run myself.

What the tests cannot prove: that `%y`'s offset is right in a zone whose rules Intl and GNU disagree about. The rendering path is `formatStrftime`, which `date` already relies on for the same directives, so this adds no new timezone logic.

`pnpm test:unit`: 14809 passed, 48 failed across 7 files, all pre-existing and unrelated (three `*.bundle.test.ts` needing a build, `readme.test.ts`, three `js-exec` security tests that fail on `Cannot find module 'run'`). Verified by stashing this change and re-running, not inferred. `pnpm typecheck` reports errors only in `js-exec/run-runtime.ts`, from the same missing module. `pnpm lint:fix` and `pnpm knip` are clean.

---

Authored with Claude Opus 5

### DIFF
--- a/.changeset/stat-format-directives.md
+++ b/.changeset/stat-format-directives.md
@@ -1,0 +1,18 @@
+---
+"just-bash": minor
+---
+
+stat: expand every `-c` directive, and print `?` for one that cannot be answered
+
+`stat -c '%Y' file` printed the two characters `%Y` and exited 0. The format expansion matched `/%[nNsFaAuUgG]/g` and returned any other directive as itself, so the whole timestamp family passed through as its own source text. `stat -c 'mtime=%Y'` is the idiom for reading a modification time, and its output was `mtime=%Y`, which looks like output rather than like a gap: nothing in the exit code, stderr, or the string says the value is missing.
+
+- `%y` / `%Y` (modify), `%x` / `%X` (access), `%z` / `%Z` (change) are expanded. The filesystem keeps one timestamp per file, so access and change report the modification time.
+- `%w` / `%W` (birth) report `-` and `0`, which is what GNU prints where a filesystem does not record birth time.
+- `%y`, `%x` and `%z` render as GNU does, `2024-01-15 09:17:14.764000000 +0000`, in `$TZ` when it names a zone Intl accepts and UTC otherwise, the same contract `date` uses. The filesystem stores milliseconds, so the last six digits of the nanoseconds are always zero.
+- `%%` is a literal percent, a `%` at the end of FORMAT prints as itself, and `-`, `0` and a width are honored, so `%5s` and `%-5s` pad.
+- A directive with no value, whether unknown or naming something this filesystem does not record, prints `?` as it does in GNU.
+- `%a` is the permission bits alone. It was `stat.mode.toString(8)`, which carries the file type: a 0644 file reported `100644` where GNU reports `644`.
+
+`%n`, `%N`, `%s`, `%F`, `%A`, `%u`, `%U`, `%g` and `%G` are unchanged, as is the default output.
+
+`%b`, `%B`, `%h`, `%i`, `%d`, `%D`, `%f`, `%o`, `%t`, `%T` and `%m` print `?`. Each names something `FsStat` does not carry (allocated blocks, link count, inode, device, mount point), and inventing a plausible constant for them is the failure this change is fixing. They are one small `FsStat` addition away if that is wanted.
--- a/.changeset/stat-format-directives.md
+++ b/.changeset/stat-format-directives.md
@@ -6,13 +6,13 @@ stat: expand every `-c` directive, and print `?` for one that cannot be answered
 
 `stat -c '%Y' file` printed the two characters `%Y` and exited 0. The format expansion matched `/%[nNsFaAuUgG]/g` and returned any other directive as itself, so the whole timestamp family passed through as its own source text. `stat -c 'mtime=%Y'` is the idiom for reading a modification time, and its output was `mtime=%Y`, which looks like output rather than like a gap: nothing in the exit code, stderr, or the string says the value is missing.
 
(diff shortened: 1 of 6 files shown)


mutewinter:
Took all three review notes in e22e252.

**Access and change times now print `?`.** The note is right and it caught an inconsistency in the PR body: arguing that a fabricated `%b` is the failure being fixed, then fabricating `%x` from mtime, is the same mistake in a different coat. The single-timestamp reasoning holds for the in-memory filesystem and not for the read-write one, where the three are real and a `chmod` moves only `ctime`.

**The wall clock is built on demand.** Confirmed as a regression: `formatStrftime` was called for every `-c` format and handed `maxOutputBytes` as its own limit, so `stat -c '%s'` under a `maxOutputSize` below 19 threw while constructing a string it never used. `resolve` is now a function `expandFormat` calls only for the directives a FORMAT names, memoized per file. Covered by a case that runs `stat -c '%s'` at `maxOutputSize: 12`.

**`%f` is answered.** With one correction to the suggestion: `stat.mode.toString(16)` gives `1a4` on the in-memory filesystem, where `mode` holds no type bits, against GNU's `81a4`. The type is composed from `isDirectory` instead, the same reason `formatMode` is passed it separately, and both a file and a directory are in the locked GNU fixture now (`81a4`, `41ed`).

`%b`, `%B`, `%h`, `%i`, `%d`, `%D`, `%o` and `%m` stay `?` for the reason the first note gives. `dev` and `ino` are on `FsStat` but optional, so answering them would mean output that varies by backend; extending `FsStat` properly is the follow-up offered in the body.

PR #422, opened 2026-09-10
fix(fs): finish a cross-mount copy around an entry it cannot copy, then report them

mutewinter:
## Problem

```ts
import { Bash, MountableFs, OverlayFs, ReadWriteFs } from "just-bash";

// docs/ holds a.txt, link.txt -> a.txt, and z.txt
const fs = new MountableFs();
fs.mount("/mnt/docs", new OverlayFs({ root: "/path/to/docs", readOnly: true }));
fs.mount("/work", new ReadWriteFs({ root: "/path/to/work" }));
const bash = new Bash({ fs, cwd: "/work" });

const { stderr, exitCode } = await bash.exec("cp -R /mnt/docs copy");
// stderr:   cp: cannot copy '/mnt/docs': EPERM: operation not permitted, symlink '/work/copy/link.txt'
// exitCode: 1
// copy/ holds a.txt and nothing else; z.txt was never reached
```

A recursive copy from one mount into another ends at the first entry the destination refuses, and everything after it is left uncopied. The ordinary case is a symlink: `ReadWriteFs` and `OverlayFs` refuse to create one unless `allowSymlinks` is set, so any tree with a link in it (a checked-out repository, a `node_modules`) fails this way. For an agent this reads as the folder being off limits rather than one entry being skipped, and it is easy to misattribute to the host: the errno is the one macOS raises when it denies an app a folder.

## Cause

`crossMountCopy` recurses per child with no handling between entries:

```ts
for (const child of children) {
  const srcChild = joinPath(src, child);
  const destChild = joinPath(dest, child);
  await this.crossMountCopy(srcChild, destChild, options);
}
```

The first rejected promise unwinds the whole walk. The existing cross-mount tests copy between two `InMemoryFs`, which create symlinks, so no entry ever fails in them.

## Fix

The walk continues past an entry it cannot copy and, once the tree is done, throws one error naming the entries it could not copy (capped at ten, then `and N more`), the way GNU `cp` reports each failed entry and exits 1 at the end. The `cp` command needs no change: it prints the error as `cp: cannot copy '<src>': copied all but 2 entries: ...` and exits 1.

## Scope

Unchanged: a copy within one mount is delegated to that filesystem's own `cp` as before. A failure on the source operand itself (unreadable, missing) still throws directly. A cross-mount `mv` that could not copy everything leaves the source in place, as it did when the copy aborted.

Not addressed, deliberately: a symlink policy for `cp` (`-L`, `--no-dereference`), which is a command-level feature rather than this bug; a structured error type carrying the failures, which nothing consumes yet; per-entry lines on stderr, since the filesystem layer has no stderr and the command prints one line per source operand.

## Tests

Two cases in `mountable-fs.test.ts`, using an `InMemoryFs` subclass whose `symlink()` throws `EPERM` (the policy `ReadWriteFs` and `OverlayFs` default to). The first asserts the files before, beside, and after two links (one nested) all arrive, the links do not, and the message names both in traversal order. The second asserts twelve refused links report ten and `and 2 more`. They do not exercise a real `ReadWriteFs` destination or the `cp` command's rendering; both are covered in the consumer suite this was found in, not here.

`vitest run src/fs/ src/commands/cp/ src/commands/mv/`: 33 files, 1197 passed, 1 skipped (pre-existing). `tsc --noEmit` in my checkout reports errors in `src/commands/js-exec/run-runtime.ts` only, a `run` module not installed here; pre-existing and untouched by this change.

---

Authored with Claude Opus 5

### DIFF
--- a/.changeset/cross-mount-copy-reports-failed-entries.md
+++ b/.changeset/cross-mount-copy-reports-failed-entries.md
@@ -1,0 +1,9 @@
+---
+"just-bash": patch
+---
+
+MountableFs: copy every other entry across mounts when one cannot be copied, then report the failures together
+
+A recursive `cp` from one mount into another walks the tree entry by entry, and one entry the destination refuses ended the whole walk: everything after it was left uncopied, and the error named only that entry. The ordinary case is a symlink, since `ReadWriteFs` and `OverlayFs` refuse to create one unless `allowSymlinks` is set, so a tree holding a single link (a checked-out repository, a `node_modules`) failed on it with `EPERM: operation not permitted, symlink ...` and nothing to show for the rest.
+
+The walk now continues past an entry it cannot copy, and once the tree is done throws one error naming the entries it could not copy, capped at ten, the way GNU `cp` reports each failed entry and exits 1 at the end. `cp -R` through `MountableFs` therefore leaves a usable copy with the links missing, and the error says which ones. A copy within one mount is that mount's own and is unchanged.
--- a/packages/just-bash/src/fs/mountable-fs/mountable-fs.test.ts
+++ b/packages/just-bash/src/fs/mountable-fs/mountable-fs.test.ts
@@ -373,7 +373,7 @@ describe("MountableFs", () => {
       await expect(
         fs.cp("/mnt/a/dir", "/mnt/b/dir", { recursive: true }),
       ).rejects.toThrow(
-        "copied all but 2 entries: EPERM: operation not permitted, symlink '/dir/link'; EPERM: operation not permitted, symlink '/dir/sub/up'",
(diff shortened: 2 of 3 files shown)


mutewinter:
Addressed the three points from the automated review in 39b0903b:

- Only the first ten failure messages are retained during the walk; the rest are counted, so memory no longer grows with the number of refused entries.
- A directory that fails before its children are reached is named as `<path> and everything in it`, so the count no longer understates a skipped subtree.
- Each named failure leads with the source entry (`/mnt/a/dir/link: EPERM: ...`) rather than whatever path the destination's error embedded.

New test for the skipped-directory case; the two existing ones updated for the message shape.

PR #424, opened 2026-09-11
fix(python): report an oversize file as EFBIG and a read-only mount as EROFS

mutewinter:
## Problem

```js
const bash = new Bash({ python: true, executionLimits: { maxStringLength: 128 } });
await bash.fs.writeFile("/tmp/big.bin", "x".repeat(256));
await bash.exec(`python3 -c "open('/tmp/big.bin').read()"`);
// OSError: [Errno 33] No file descriptors available: '/host/tmp/big.bin'
// OSError: [Errno 33] No file descriptors available: '/lib/python313.zip'
```

Any file over `maxStringLength` (8 MB by default through the bridge buffer) does this, after a noticeable pause, and the second line is the stdlib zip: the interpreter cannot import `traceback` to print the first error. Two smaller ones: a file the bridge refuses as too large for its buffer is reported as `FileNotFoundError` for a file that exists, and a write into a read-only mount is `OSError: [Errno 29] I/O error` rather than the `EROFS` the mount actually raised.

## Cause

The errno table in `createHOSTFS` has

```ts
EFBIG: 27,
```

Under Emscripten's numbering 27 is `EINTR`; `EFBIG` is 22 (`import errno; errno.EFBIG` in the vendored build says so, and `os.strerror(22)` is `File too large`). CPython retries an `open()` that fails with `EINTR` (PEP 475), and Emscripten's `FS.open` has already allocated the stream when `stream_ops.open` throws, so every retry leaks a descriptor until the table is exhausted and the retry fails with the real `EMFILE`, 33. `ENODATA: 42` is wrong the same way (116).

The two mappings: `stream_ops.open` catches every `backend.readFile` failure as `ENOENT`, so the bridge's `Result too large: N > 8388608` reads as a missing file; and `tryFSOperation` has no branch for `read-only`/`erofs`, so the mount's `EROFS: read-only file system, ...` falls through to `EIO`.

`python3.security.test.ts` asserted `[Errno 27]` for the truncate case, which is how the wrong number stayed in place: the test was checking the message the bug produced.

## Fix

`EFBIG` is 22 and `ENODATA` 116; `tryFSOperation` maps `read-only`/`erofs` to `EROFS` and `too large` to `EFBIG`; `stream_ops.open` throws `EFBIG` rather than `ENOENT` when the read failed as too large.

## Scope

Unchanged: `maxFileSize` and where it is enforced; a missing file is still `ENOENT`.

Not addressed, deliberately: the 8 MB per-operation bridge buffer (`Size.DATA_BUFFER`), which is what makes a 9 MB file unreadable at any `maxStringLength`. Chunked reads across the bridge would lift it; that is a protocol change and a separate PR if you want it.

## Tests

`python3.files.test.ts`: a 256-byte file under `maxStringLength: 128` raises `OSError: [Errno 22] File too large` once, with no `No file descriptors available`; a write into an `OverlayFs({ readOnly: true })` mount raises `[Errno 69] Read-only file system` and creates nothing, while a read of the same mount works. `python3.security.test.ts`: the truncate assertion now names `[Errno 22] File too large`. Both new tests fail before the change with the output quoted above.

Suite: 240 passed, 2 skipped across the 16 files.

---

Authored with Claude Opus 5

### DIFF
--- a/.changeset/python3-hostfs-errnos.md
+++ b/.changeset/python3-hostfs-errnos.md
@@ -1,0 +1,7 @@
+---
+"just-bash": patch
+---
+
+python3: report a file over the size limit as EFBIG, and a read-only mount as EROFS
+
+The worker's errno table had `EFBIG` as 27, which is `EINTR` under Emscripten's numbering (`EFBIG` is 22). CPython retries an `open()` that fails with `EINTR`, and each retry leaked the stream Emscripten had already allocated for the failed open, so reading a file over `maxStringLength` looped until the descriptor table was exhausted and the script died of `OSError: [Errno 33] No file descriptors available`, taking any later stdlib import with it. A read the bridge refused as too large for its buffer surfaced as `ENOENT`, and a write into a read-only mount as `EIO`; both now carry their own errno, so Python reports `File too large` and `Read-only file system`. `ENODATA` is corrected to 116 alongside.
--- a/packages/just-bash/src/commands/python3/python3.files.test.ts
+++ b/packages/just-bash/src/commands/python3/python3.files.test.ts
@@ -1,5 +1,11 @@
+import { mkdtempSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
 import { describe, expect, it } from "vitest";
 import { Bash } from "../../Bash.js";
+import { InMemoryFs } from "../../fs/in-memory-fs/in-memory-fs.js";
+import { MountableFs } from "../../fs/mountable-fs/mountable-fs.js";
+import { OverlayFs } from "../../fs/overlay-fs/overlay-fs.js";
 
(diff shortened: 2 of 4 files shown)


mutewinter:
Good catch on the append path, and it was real: with the first commit, `open('/tmp/big.log', 'a')` on a 9 MB file wrote `tail` back over the whole file with no error, because the bridge's `Result too large` fell into the `isCreate && isWrite` branch as an empty file. The size failure is now classified before the create fallback (`72b6d8bf` → `HEAD`), and `python3.files.test.ts` gained a test that appends to a 9 MB file and asserts `[Errno 22] File too large` with the file left intact; it fails against the previous commit with an empty stderr. Note the earlier 256-byte case never reached this branch, since a read under the bridge buffer succeeds and is caught by the `maxFileSize` check after it; only a file the bridge itself refuses did.

PR #425, opened 2026-09-11
fix(python): name the program in tracebacks and run it in its own namespace

mutewinter:
## Problem

```js
await bash.exec(`python3 -c "import json
x = 1
raise KeyError(2)"`);
// Traceback (most recent call last):
//   File "/tmp/_jb_script.py", line 413, in <module>
//     raise KeyError(2)
// KeyError: 2
```

A three-line program fails at line 413 of a file nobody wrote. A script file gets the same treatment: `python3 report.py` reports `/tmp/_jb_script.py` too, so the path and line number are wrong in every traceback the runtime produces, which is the one output an agent debugging its script reads most carefully. Related: `sys.exit("message")` exits 1 silently where CPython prints the message, `__file__` is unset for a script, `sys.path[0]` is the wrapper's `/tmp`, and the wrapper's own imports leak into the program:

```js
await bash.exec(`python3 -c "print(sorted(k for k in globals() if not k.startswith('__')))"`);
// ['Path', '_JbHttp', '_JbHttpResponse', '_base64', '_glob_module', ... 'json', 'os', 'sys', 'types']
```

## Cause

`runPython` pastes `input.pythonCode` into the wrapper's `try:` block after roughly four hundred lines of path shims and the HTTP bridge, and runs the whole file:

```ts
${input.pythonCode
  .split("\n")
  .map((line) => `    ${line}`)
  .join("\n")}
except SystemExit as e:
```

So the program is part of `/tmp/_jb_script.py`, its line numbers are offset by the wrapper's length (which varies with the environment variable count), and it shares the wrapper's module globals. `traceback.print_exc()` then prints the wrapper's own frame first. The existing tests assert on stdout and on `ValueError` appearing in stderr, never on the file or line a traceback names.

## Fix

The program is compiled with `compile(code, name, 'exec')` under its own name (`<string>` for `-c` and `-m`, `<stdin>` for a program read from stdin, the script path as typed for a file) and run with `exec` in a fresh `__main__` namespace. `python3.ts` passes the name as a new `fileName` on `WorkerInput`. For a script file, `__file__` is the absolute path and `sys.path[0]` is `/host` plus the script's directory, so a sibling module imports without the `/host` prefix the current test comment documents; for inline code, `sys.path[0]` is `''` as CPython sets it. The `except` clause prints from `e.__traceback__.tb_next`, which drops the wrapper's frame, and `sys.exit` with a non-integer code prints it to stderr.

`JSON.stringify` produces the Python literal: JSON's escapes are a subset of Python's, and the wrapper is written to MEMFS as UTF-8.

## Scope

Unchanged: the path shims, the HTTP bridge and `jb_http`, `sys.argv`, exit codes, and `runpy` behavior for `-m`, whose tracebacks already named real files because `runpy` compiles them itself.

Not addressed, deliberately: frames inside the shims themselves (`_redir_open` and friends) still appear below the program's frame when an `OSError` comes through one, and `BaseException`s other than `Exception` still propagate through the wrapper. Both are small; say the word and either is a few lines.

## Tests

`python3.files.test.ts`: `-c` code fails at `File "<string>", line 3`; a script fails at `File "report.py", line 4` and `line 2, in fail` with no wrapper frame; `__file__` is `/tmp/app/main.py` and `import helper` finds a sibling without touching `sys.path`; the program's non-dunder globals are `[]`. `python3.env.test.ts`: `sys.exit('error message')` writes `error message\n` to stderr. All five fail before the change. The existing `runpy` and `/host/tmp` `sys.path` tests still pass unchanged.

Suite: 242 passed, 2 skipped across the 16 files.

---

Authored with Claude Opus 5

### DIFF
--- a/.changeset/python3-tracebacks-name-the-script.md
+++ b/.changeset/python3-tracebacks-name-the-script.md
@@ -1,0 +1,7 @@
+---
+"just-bash": patch
+---
+
+python3: name the program in tracebacks and run it in its own namespace
+
+The program was pasted into the worker's wrapper script, so every traceback named `/tmp/_jb_script.py` at a line four hundred past the program's own, for `-c` code and script files alike, and the wrapper's helpers (`json`, `Path`, `_orig_open`, ...) sat in the program's globals. The program is now compiled under its own name (`<string>`, `<stdin>`, or the script path as typed) and run in a fresh `__main__` namespace, the wrapper's own frame is dropped from what is printed, `__file__` and `sys.path[0]` are set for a script file as CPython sets them (so a sibling module imports without a `/host` prefix), and `sys.exit("message")` prints the message before exiting 1.
--- a/packages/just-bash/src/commands/python3/python3.env.test.ts
+++ b/packages/just-bash/src/commands/python3/python3.env.test.ts
@@ -171,6 +171,16 @@ python3 -c "import os; print(os.getcwd())"
       expect(result.exitCode).toBe(1);
     });
 
+    it("exits 1 and prints the value for any other non-integer code", async () => {
+      const env = new Bash({ python: true });
+      const empty = await env.exec(`python3 -c "import sys; sys.exit('')"`);
+      expect(empty.stderr).toBe("\n");
+      expect(empty.exitCode).toBe(1);
+      const float = await env.exec(`python3 -c "import sys; sys.exit(0.0)"`);
+      expect(float.stderr).toBe("0.0\n");
+      expect(float.exitCode).toBe(1);
+    });
+
     it("should return exit code 1 from sys.exit(None)", async () => {
       const env = new Bash({ python: true });
       const result = await env.exec(`python3 -c "import sys; sys.exit(None)"`);
--- a/packages/just-bash/src/commands/python3/python3.env.test.ts
+++ b/packages/just-bash/src/commands/python3/python3.env.test.ts
@@ -167,6 +167,7 @@ python3 -c "import os; print(os.getcwd())"
         `python3 -c "import sys; sys.exit('error message')"`,
       );
       // sys.exit with string message prints to stderr and exits with 1
+      expect(result.stderr).toBe("error message\n");
       expect(result.exitCode).toBe(1);
     });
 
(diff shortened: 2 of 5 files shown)


mutewinter:
All four of the maintainer/general findings were right and are fixed in the second commit, with a test for each: the program now runs in a `types.ModuleType('__main__')` registered in `sys.modules`, so `import __main__` is the program and `pickle.dumps` of a class it defines round-trips (`unittest.main()` and `doctest` look it up the same way); a compile-time error has no program frame and now prints from the exception alone, which is CPython's own SyntaxError shape, `File "<string>", line 1` and no wrapper; `sys.exit('')` and `sys.exit(0.0)` print the value and exit 1; and the worker is told where the program came from (`source: "file" | "stdin" | "inline"`) instead of reading it off the file name, so a script named `<stdin>` is still a file.

The PEP 263 finding does not hold on the vendored 3.13: `compile()` of a `str` whose first line is `# -*- coding: utf-8 -*-` (or `latin-1`) compiles without error; I checked both in this build. Compiling `.encode('utf-8')` instead would be worse for this pipeline, since the filesystem has already decoded the file as UTF-8 and a `latin-1` cookie would then re-decode every non-ASCII literal. The source stays a `str`, and a test with a `latin-1` cookie over UTF-8 text asserts the literal survives. Suite: 246 passed, 2 skipped across the 16 files; the three new tests for the findings fail against the first commit.

mutewinter:
Pushed 5bf12fe3, two things.

The traceback cases now live in `python3.tracebacks.test.ts`, registered in the fork pool like the other real-worker suites, rather than appended to `python3.files.test.ts`. #424 appends to that same file, so whichever of the two merged second was going to conflict on it; now neither touches it.

While looking at #431, which reports the wrapper's four-space indentation rewriting the contents of triple-quoted strings, I checked whether this change already covers it: it does, since the program reaches `compile()` as its own string literal and never passes through the indented paste. Five cases pin that (a triple-quoted string from a script, stdin, and `-c`; code built in a string and run with `exec()`; a module-level `from __future__` import), and all five fail against a worker built from `main`'s `worker.ts`. The body now names #431 and the finding.

PR #439, opened 2026-09-19
fix(bash): pass per-exec env to nested sh/bash and stop leaking parent vars

### LINKED ISSUE #438: Bash.exec env is not exported to nested sh/bash (replaceEnv leaves exportedVars from parent)
## Summary

`Bash.exec(…, { env, replaceEnv: true })` correctly puts keys into the execution environment map, so a directly executed command such as `env` / `printenv` sees them. A nested `sh` or `bash` (`sh -c`, `bash -c`, `/bin/sh`, script file) does **not**: the nested shell never receives those variables, and with `replaceEnv: true` it still inherits the **parent** shell's exported environment.

This is silent (`exitCode: 0` with an empty expansion), which makes it easy to misread as a caller bug.

Measured on **just-bash@3.4.2** and confirmed still present on current `main` (`packages/just-bash/src/Bash.ts` exec-state construction; `packages/just-bash/src/commands/bash/bash.ts` `executeScript`).

## Reproduction

```js
import { Bash } from "just-bash";

const bash = new Bash({ env: { SECRET: "leak" } });
const env = { MARKER: "YES", PATH: "/usr/bin:/bin" };
const opts = { env, replaceEnv: true };

const direct = await bash.exec("env", opts);
console.log("direct MARKER", direct.stdout.includes("MARKER=YES")); // true
console.log("direct SECRET", direct.stdout.includes("SECRET"));     // false

const nested = await bash.exec("sh", {
  args: ["-c", "echo [$MARKER]; printenv MARKER; printenv SECRET"],
  ...opts,
});
console.log("nested stdout:", JSON.stringify(nested.stdout));
// actual:   "[]\n\nleak\n"   (or similar: MARKER missing, SECRET present)
// expected: "[YES]\nYES\n" with printenv SECRET exiting 1 / absent

console.log("nested exitCode", nested.exitCode); // 1 from missing SECRET printenv, not from MARKER
```

Also fails the same way for:

| invocation | observed |
|---|---|
| `bash.exec("bash", { args: ["-c", "echo [$MARKER]"], env, replaceEnv: true })` | `[]` |
| `bash.exec("/bin/sh", { args: ["-c", "printenv MARKER"], env, replaceEnv: true })` | empty, exit 1 |
| `bash.exec("/bin/bash", { args: ["-c", "env"], env, replaceEnv: true })` | `MARKER` absent; parent secrets still present |
| script-file form of `sh`/`bash` with the same `env`/`replaceEnv` | same |

Contrast that works today:

```js
await bash.exec("echo [$MARKER]", { env: { MARKER: "YES" }, replaceEnv: true });
// expands MARKER — the *outer* exec path applies env to the interpreter that runs the string
```

So the distinguishing factor is not `env` vs no `env`, but **target**: a real command/builtin that reads `ctx.env` vs the `sh`/`bash` builtin that re-execs from `exportedEnv`.

## Export / leak side checks (same root cause)

```js
const bash = new Bash();

// Exported vars *do* reach sh -c (exportedVars path works when the key is exported)
(await bash.exec("export FOO=bar\nsh -c 'echo [$FOO]'")).stdout.trim(); // "[bar]"

// Unexported shell vars incorrectly leak into sh -c today (merge without replaceEnv;
// nested exec does not start from exported-only + replaceEnv)
(await bash.exec("FOO=noleak\nsh -c 'echo [$FOO]'")).stdout.trim();
// actual:   "[noleak]"
// expected: "[]"  (real bash: unexported vars are not in the child's environment)

// Prefix assignment still works
(await bash.exec("FOO=pref sh -c 'echo [$FOO]'")).stdout.trim(); // "[pref]"
```

With `replaceEnv: true`, parent secrets that were never in `options.env` should not appear in the nested shell either; today `printenv SECRET` after `{ env: { MARKER: "YES" }, replaceEnv: true }` still prints the constructor `SECRET`.

## Where it breaks (labelled diagnosis)

1. **`Bash.exec`** (`packages/just-bash/src/Bash.ts`): builds `execEnv` from `replaceEnv` / `options.env`, then:

   ```ts
   const execState: InterpreterState = {
     ...this.state,   // ← copies parent exportedVars unchanged
     env: execEnv,
     arrays: effectiveOptions.replaceEnv ? new Map() : cloneArrays(this.state.arrays),
     // …
   };
   ```

   Per-exec keys land in `env` but are **not** added to `exportedVars`. On `replaceEnv: true`, `exportedVars` still points at the persistent shell's set.

2. **`sh` / `bash`** (`packages/just-bash/src/commands/bash/bash.ts` `executeScript`): builds `positionalEnv` from `ctx.exportedEnv` only and calls `ctx.exec` / `execWithInheritedStdin` **without** `replaceEnv: true`:

   ```ts
   await ctx.exec(scriptToRun, {
     env: positionalEnv,
     cwd: ctx.cwd,
     // no replaceEnv
   });
   ```

   Nested exec therefore merges onto the persistent parent `this.state.env` and only overlays exported keys. Result: `MARKER` missing; `SECRET` still present.

Both halves are required for a correct fix:

- When applying per-exec `env`, update `exportedVars` (on `replaceEnv`, set it to the keys of the replacement map, including `PWD` if set).
- When `sh`/`bash` starts the nested interpreter, pass `replaceEnv: true` so the child starts from the forwarded exported environment rather than the persistent shell.

Either change alone is incomplete: exporting keys without nested `replaceEnv` still leaks parent secrets; nested `replaceEnv` alone still omits `MARKER` because it is not in `exportedEnv`.

## Why this matters

Callers that implement Node-like `child_process` semantics (or any API that passes `{ env, replaceEnv: true }` and then runs `sh -c '…'`) see a working direct binary and a broken nested shell. `sh -c` is a common composition pattern when the host has no real process model / `chdir`. The failure mode is success-looking empty expansions.

The README documents `replaceEnv` as “Start with empty env instead of merging”, which matches what direct `env` does today and what nested `sh`/`bash` does not.

## Suggested fix (source-level)

In `Bash.exec`, after building `execEnv` / `newPwd`:

```ts
const exportedVars = effectiveOptions.replaceEnv
  ? new Set(execEnv.keys())
  : new Set(this.state.exportedVars);
if (!effectiveOptions.replaceEnv && effectiveOptions.env) {
  for (const key of Object.keys(effectiveOptions.env)) {
    exportedVars.add(key);
  }
  if (newPwd !== undefined) exportedVars.add("PWD");
}

const execState: InterpreterState = {
  ...this.state,
  env: execEnv,
  exportedVars,
  arrays: effectiveOptions.replaceEnv ? new Map() : cloneArrays(this.state.arrays),
  // …
};
```

In `executeScript`, add `replaceEnv: true` to both nested option objects.

## Environment

- just-bash **3.4.2** (published) and current `main` source as of filing
- Node 22+, macOS

Happy to send a PR against source if that would help; we currently carry an equivalent dist patch downstream.

trieloff:
Fixes #438.

## Problem

`Bash.exec(…, { env, replaceEnv: true })` sets the per-exec variables for directly run commands, but a nested `sh`/`bash` never sees them, while the constructor's env (e.g. `SECRET`) still reaches it. There were two halves, as the issue describes:

1. `Bash.exec` added per-exec keys to `env` but not to `exportedVars`. It also kept the parent's export set under `replaceEnv`, and shared that `Set` by reference, so an `export` in one `exec()` leaked into every later one.
2. `sh`/`bash` (`executeScript`) called the nested exec without `replaceEnv`, so the child merged onto the persistent shell's full env. That made unexported variables and parent secrets visible.

## Fix

- **Exports** (`Bash.ts`): the export set is copied per exec. Only the host `Bash.exec` API and new shells treat `env` as *the environment* and export its keys, and only valid names count, so positional parameters (`0`, `#`, `1`) stay local. Internal callers such as `env` and `time` pass the full variable map, so they keep the persistent export set as before.
- **New shells** (`newShell` on the internal `CommandExecOptions`): nested `sh`/`bash` passes only the parent's exported env plus positional parameters, and asks for a new shell. `Bash.exec` then initializes what bash sets up at startup, all **unexported**:
  - `PATH` and the host variables (`OSTYPE`, `MACHTYPE`, `HOSTTYPE`, `HOSTNAME`) get fixed defaults when the environment lacks them
  - `IFS`, `OPTIND`, `SHELLOPTS` and `BASHOPTS` are always reset
- **`cd -`**: under `replaceEnv` the previous directory comes only from the provided `OLDPWD`, not the parent's state. A child that didn't inherit `OLDPWD` reports `cd: OLDPWD not set`.

Top-level `replaceEnv` still starts from exactly the given env. `printenv` with `{ env: { A: "1" }, replaceEnv: true }` prints only `A=1`.

## Behaviour (checked against `/bin/bash` 3.2.57 and bash 5.3)

| script | before | after / bash |
|---|---|---|
| `sh -c 'echo [$MARKER]; printenv SECRET'` with `{ env: {MARKER:"YES"}, replaceEnv: true }` | `[]` / `leak` | `[YES]` / exit 1 |
| `export FOO=bar` in one exec, then `FOO=noleak; sh -c 'echo [$FOO]'` | `[noleak]` | `[]` |
| `FOO=secret; env sh -c 'echo [$FOO]'` | `[]` | `[]` |
| `export -n HOME; sh -c 'echo [$HOME]'` | `[/home/user]` | `[]` |
| `HOSTNAME=secret; export -n HOSTNAME; sh -c 'echo $HOSTNAME'` | `secret` | fresh default |
| `unset PATH; sh -c 'IFS=:; export -p'` | — | lists none of `PATH`/`IFS`/`OPTIND`/… |
| `export IFS=: OPTIND=5; sh -c '…'` | `IFS=:`… | `IFS=' \t\n'`, `OPTIND=1` |
| `cd /tmp; export -n OLDPWD; sh -c 'cd -'` | `/home/user` | `OLDPWD not set`, exit 1 |

(`OSTYPE`/`HOSTNAME` defaults are the sandbox's `linux-gnu`/`localhost` rather than the host's values.)

## Out of scope, unchanged from `main`

- `FOO=x env sh -c 'echo $FOO'` prints `[]` (bash: `[x]`), because the `env` command doesn't mark its assignments exported.
- `env -i printenv` prints `PWD=…`.
- `env`/`printenv` list all shell variables, not just exported ones.

## Testing

- `src/commands/bash/bash.env.test.ts` (24 tests, one per scenario above). 6 of them cover the review findings and fail on the first revision of this PR.
- `pnpm typecheck`, `pnpm lint:fix`, `pnpm lint:banned`, `pnpm knip` are clean. There is a changeset.
- `pnpm test:run`: everything passes except 6 Python/WASM tests that time out loading CPython locally (they pass in CI).


trieloff:
Thanks. All findings are addressed in 2780bf9b:

- **Seeded variables exported / copied from the parent map** (bash.ts:159/165/169, both reviewers): nested `sh`/`bash` no longer seeds anything through `env`. It passes only the exported env plus positional parameters and sets an internal `newShell` flag. `Bash.exec` then initializes `PATH` and the host variables from fixed defaults (only when not exported), and always resets `IFS`/`OPTIND`/`SHELLOPTS`/`BASHOPTS`, all unexported. `HOSTNAME=secret; export -n HOSTNAME` no longer reaches the child, and `export -p` there doesn't list any of these.
- **`env`/`time` re-export the whole variable map** (Bash.ts:749): only the host `Bash.exec` API and new shells export their `env` keys. Internal callers keep the persistent export set, so `FOO=secret; env sh -c 'echo [$FOO]'` prints `[]` again.
- **SHELLOPTS/BASHOPTS injected into `env -i`** (Bash.ts:761): these are now set only for new shells.
- **`previousDir` leaks through `replaceEnv`** (Bash.ts:777): `replaceEnv` takes the previous directory only from the provided `OLDPWD`, and `cd -` without one reports `OLDPWD not set`.
- **Changeset**: added `.changeset/nested-shell-environment.md`.

Each finding has a regression test in `bash.env.test.ts`.

packages/just-bash/src/Bash.ts
```diff
@@ -737,14 +749,59 @@ export class Bash {
...
+      // callers such as `env` and `time` pass the full variable map and keep
+      // the persistent export set. The set is always copied so an `export` in
+      // one exec does not leak into later ones. Only valid names can be
+      // exported: nested sh/bash passes positional parameters ("0", "#", "1")
+      // through env.
+      const exportsEnv = execDepth === 0 || effectiveOptions.newShell;
+      const exportedVars =
+        replaceEnv && exportsEnv
+          ? new Set<string>()
+          : new Set(this.state.exportedVars);
```
privatenumber: Could the recursive exec path inherit the calling execution's export set rather than the instance's original set?

With `export FOO=bar; env sh -c 'echo [$FOO]'`, `env` forwards the value of `FOO`, but this branch copies export flags from `this.state.exportedVars`, which no longer includes the export made in the running script, so the nested shell drops `FOO`. Copying the set correctly stops leakage between separate exec calls, but the recursive path needs to preserve exports within the current call. Could we add this case alongside the existing test that unexported variables stay private?

packages/just-bash/src/Bash.ts
```diff
@@ -723,7 +733,9 @@ export class Bash {
       }
 
       // Create environment for this execution
-      const execEnv = effectiveOptions.replaceEnv
+      const replaceEnv =
+        effectiveOptions.replaceEnv || effectiveOptions.newShell;
+      const execEnv = replaceEnv
```
privatenumber: Could we track this remaining wrapper path explicitly? I think it is pre-existing, so I'm not treating it as a regression introduced by this PR.

`timeout` calls `ctx.exec` without forwarding an environment, which reaches this branch and restores the constructor's variables. For example, `new Bash({ env: { SECRET: "example-only" } }).exec("timeout 5 printenv SECRET", { env: {}, replaceEnv: true })` would still expose `SECRET` despite the outer replacement. The direct `sh`/`bash` fix is useful; recursive execution would also need to inherit the active caller's environment to keep that isolation through wrappers.

privatenumber (changes requested):
Thanks for the fix. The direct `sh`/`bash` path now behaves correctly and this is worth landing. I'd like to request changes for one regression, with one remaining gap flagged separately.

**Blocking:** recursive execution (used by `env`, `time`, `timeout`, ...) copies the instance's export set instead of the running execution's, so an `export` made in the script is lost when a nested `sh`/`bash` is launched through a wrapper. Details inline.

**Follow-up:** a `replaceEnv` isolation gap through wrappers looks pre-existing rather than introduced here. I've flagged it inline so we can decide whether to fix it in this PR or track it.

